PR #2, opened 2017-01-28
Documentation

bartlannoeye:
Setting up initial documentation sample so we can go configure readthedocs.io


Templates/Docs_Effect_Template.md
```diff
@@ -0,0 +1,54 @@
+# CapitalizeKeyboardEffect
+
+The **CapitalizeKeyboardEffect** enforces caps on the keyboard for an Entry control.
+
+## Syntax
+
+You can either use the blur behavior from your XAML code:
```
Depechie: Not the blur effect ;)

PR #27, opened 2017-03-09
Adding media

bartlannoeye:
- Putting Media in master for nuspecs
(commit will be merged to dev after PR)


bartlannoeye:
CI fails because master doesn't have code yet to build, so can be ignored.

Depechie:
Do we still want to use this media? Was there from the first version of the lib but since then changed name... so for me this can be omitted.

PR #39, opened 2017-06-25
FormsCommunityToolkit: Added Animations Sample

jsuarezruiz:
Added Animations Sample that allows to test all the possibilities of the animations library.


Samples/Samples/Samples/Views/AnimationsPage.xaml.cs
```diff
@@ -0,0 +1,19 @@
...
+using Xamarin.Forms;
+
+namespace FormsCommunityToolkit.Samples.Views
+{
+    public partial class AnimationsPage : TabbedPage
+    {
+        public AnimationsPage()
+        {
+            InitializeComponent();
+            var test = new BeginAnimationBehavior();
```
Depechie: Why is this test variable being declared here?

src/Animations/Animations/Triggers/BeginAnimation.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace FormsCommunityToolkit.Animations
+{
+    public class BeginAnimation : TriggerAction<VisualElement>
+    {
+        public AnimationBase Animation { get; set; }
+
+        protected override async void Invoke(VisualElement sender)
+        {
+            if (Animation != null)
```
Depechie: I know, it's hard ;) But please only positive checks. if( == null ) return;

Samples/Samples/Samples/Views/AnimationsPage.xaml.cs
jsuarezruiz: Ooops. It was a test that has totally escaped!

src/Animations/Animations/Triggers/BeginAnimation.cs
jsuarezruiz: Yes, no problem ;)

PR #45, opened 2017-07-31
Added 2 entry sample pages in the samples repo

Depechie:
Nothing fancy, added disable auto correct for entry and italic placeholder for entry sample pages


Samples/Samples/Samples/Views/EffectsEntryDisableAutoCorrectPage.xaml.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+using Xamarin.Forms;
+using Xamarin.Forms.Xaml;
+
+namespace FormsCommunityToolkit.Samples.Views
+{
+	[XamlCompilation(XamlCompilationOptions.Compile)]
```
jimbobbennett: Any reason why we can't move this to the assemblyinfo.cs an turn it on for the whole project?

Depechie: No reason at all! I keep forgetting that VS2017 auto adds this to the pages when you select Add new content page...
Will adjust this :)

Depechie:
@jimbobbennett done :)

PR #49, opened 2017-09-18
Added Size Font To Fit label effect sample

### LINKED ISSUE #48: Sample of LableSizeFontToFit
## Feature Request:
I am trying to get LableSizeFontToFit working but without any luck. It would be great to have a working sample of how to use it since in this case it's probably my code that's the issue.

Depechie:
Fixes #48  .

### DIFF
--- a/Samples/Samples/Samples/Views/EffectsLabelPage.xaml.cs
+++ b/Samples/Samples/Samples/Views/EffectsLabelPage.xaml.cs
@@ -1,0 +1,19 @@
+﻿using Xamarin.Forms;
+using Xamarin.Forms.Xaml;
+
+namespace FormsCommunityToolkit.Samples.Views
+{
+    [XamlCompilation(XamlCompilationOptions.Compile)]
+    public partial class EffectsLabelPage : ContentPage
+    {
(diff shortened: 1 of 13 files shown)


Samples/Samples/Samples/Views/EffectsLabelPage.xaml.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Xamarin.Forms;
+using Xamarin.Forms.Xaml;
+
+namespace FormsCommunityToolkit.Samples.Views
+{
+    [XamlCompilation(XamlCompilationOptions.Compile)]
```
jsuarezruiz: Do we put the XAML compilation at the library level?

Depechie: Oh for F S.
Indeed we do, but thanks to the add new template in VS2017 it's being added each time on file level and I keep forgetting to delete it...
Good catch

Depechie: Did a new code push, that fixes this!

PR #53, opened 2017-10-14
Added an effect to remove the underline of an editor in android EditT…

mohaider:

Added an effect to hide the underline of an editor.


Depechie:
@mohaider is this one different from the one we already have? https://github.com/FormsCommunityToolkit/FormsCommunityToolkit/blob/dev/src/Effects/Effects.Android/Effects/Entry/EntryRemoveLine.cs
Have not yet checked your code though... any idea?

mohaider:
Hmm, it appears the they both do the same thing, although the one commited sets the drawable to null whereas the EntryRemoveLine sets a new drawable with a transparent color. I'll test to see if this functions similarly.

Issue #56, opened 2017-10-17
Multiple converters


wcoder:
## Feature Request:

Multiple converters.

Conversion values using the sequential conversion of each of the converters from the first to the last.

```xaml
<converters:ValueConverterGroup x:Key="IsEmptyInverse">
    <converters:IsEmptyConverter />
    <converters:InvertedBooleanConverter />
</converters:ValueConverterGroup>
```

PR #57, opened 2017-10-17
Multiple converters

wcoder:
Multiple converters #56 

Migrated from my repository https://github.com/XForms/XForms.Converters

### DIFF
--- a/src/Converters/Converters.Tests/Mocks/MockConverter.cs
+++ b/src/Converters/Converters.Tests/Mocks/MockConverter.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Globalization;
+using Xamarin.Forms;
+
+namespace Converters.Tests.Mocks
+{
+    public class MockConverter : IValueConverter
+    {
+        public const string TagFormat = "c{0}<-";
+
+        public int Id { get; }
+
+        public MockConverter()
+        {
+            Id = 0;
+        }
+
+        public MockConverter(int id)
+        {
+            Id = id;
+        }
+
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            return string.Format(TagFormat, Id) + value;
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture) => null;
+    }
+}
--- a/src/Converters/Converters.Tests/ValueConverterGroupTests.cs
+++ b/src/Converters/Converters.Tests/ValueConverterGroupTests.cs
@@ -1,0 +1,57 @@
+﻿using System;
+using System.Globalization;
+using FormsCommunityToolkit.Converters;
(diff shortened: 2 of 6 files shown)


src/Converters/Converters/ValueConverterGroup.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    [ValueConversion(typeof(IEnumerator<IValueConverter>), typeof(IEnumerator<IValueConverter>))]
+    public class ValueConverterGroup : List<IValueConverter>, IValueConverter
+    {
+        public static ValueConverterGroup Instance { get; } = new ValueConverterGroup();
+
+        /// <summary>
+        /// Convert the value using a sequence of converters, from first to last.
+        /// </summary>
+        /// <param name="value"></param>
```
bartlannoeye: Either provide useful explanation of each property, or you can leave these away. Right now it takes screen estate without added value.

src/Converters/Converters.Tests/ValueConverterGroupTests.cs
```diff
@@ -0,0 +1,57 @@
...
+using FormsCommunityToolkit.Converters;
+using NUnit.Framework;
+using Xamarin.Forms;
+
+namespace Converters.Tests
+{
+    [TestFixture]
+    public class ValueConverterGroupTests
+    {
+        class MockConverter : IValueConverter
```
bartlannoeye: I would prefer to see this type in a separate file (maybe even folder/namespace) so it can be reused for other tests.

wcoder: done

src/Converters/Converters/ValueConverterGroup.cs
wcoder: done

src/Converters/Converters.Tests/Mocks/MockConverter.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Globalization;
+using Xamarin.Forms;
+
+namespace Converters.Tests.Mocks
+{
+    public class MockConverter : IValueConverter
```
Depechie: Please move the converter tests to the sample application, instead inline in the general library ( to keep the actual nuget result small )

wcoder: Ok, but why? tests are not contained in the resulting Nuget package:
https://github.com/FormsCommunityToolkit/FormsCommunityToolkit/blob/dev/nuget/Converters.nuspec#L27
only general assembly of converters

Depechie: Are they stripped? Ok never mind then, did not double check the nuget, was something James has set up.

PR #61, opened 2017-11-12
Updated Animations library, updated sample and added Animations nuspec

jsuarezruiz:
Added ColorAnimation to the Animations Library
Updated Animations Sample
Added Animations nuspec

### DIFF
--- a/nuget/Animations.nuspec
+++ b/nuget/Animations.nuspec
@@ -12,12 +12,12 @@
        <requireLicenseAcceptance>false</requireLicenseAcceptance>
        <description>
          Xamarin.Forms Community Toolkit Animations Library
-         Built against: 2.3.2.127
+         Built against: 2.3.4.247
        </description>
        <summary>Xamarin.Forms Community Toolkit Animations Library</summary>
        <tags>community, toolkit, xamarin, pcl, android, xamarin.forms, ios, uwp, animations</tags>
        <dependencies>
-         <dependency id="Xamarin.Forms" version="2.3.2.127" />
+         <dependency id="Xamarin.Forms" version="2.3.4.247" />
        </dependencies>
      <releaseNotes>
      </releaseNotes>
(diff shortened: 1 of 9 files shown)


nuget/Animations.nuspec
```diff
@@ -0,0 +1,56 @@
...
+       <title>Xamarin.Forms Community Toolkit: Animations</title>
+       <authors>FormsCommunityToolkit</authors>
+       <owners>FormsCommunityToolkit</owners>
+       <projectUrl>https://github.com/FormsCommunityToolkit/FormsCommunityToolkit</projectUrl>
+       <licenseUrl>https://github.com/FormsCommunityToolkit/FormsCommunityToolkit/blob/master/LICENSE</licenseUrl>
+       <iconUrl>https://github.com/FormsCommunityToolkit/FormsCommunityToolkit/blob/master/Media/OrganonLogo.64x64.png</iconUrl>
+       <requireLicenseAcceptance>false</requireLicenseAcceptance>
+       <description>
+         Xamarin.Forms Community Toolkit Animations Library
+         Built against: 2.3.2.127
```
bartlannoeye: I see another version in the samples?

src/Animations/Animations/Animations/ColorAnimation.cs
```diff
@@ -0,0 +1,38 @@
+﻿using FormsCommunityToolkit.Animations.Extensions;
+using System;
```
bartlannoeye: system usings first

nuget/Animations.nuspec
jsuarezruiz: Yes, you are right, the 2.3.4.247 version.

src/Animations/Animations/Animations/ColorAnimation.cs
jsuarezruiz: Ok!

PR #71, opened 2018-02-15
Merge into master

jamesmontemagno:
We should get rid of dev. Master should always be shipping IMHO


bartlannoeye:
If we only have master, we can't do quick bugfixes on master (since future changes will already be on it). It's also more easily to mark big releases (1.0 - 1.1 - 2.0) when using dev+master. Up to you all decide if that is an issue or not.

listepo-alterpost:
👍

PR #74, opened 2018-03-15
Add Stylecop ruleset used by UWPCommunityToolkit.

jgiacomini:
Create coding guidelines : #1

- Add static code analysis tool that checks C# code conformance to recommended coding styles and a subset of Microsoft's .NET Framework Design Guidelines
- Fix warnings generated by stylecop


src/Animations/Animations/Shared/Animations/Base/AnimationBase.cs
```diff
@@ -6,8 +6,13 @@ namespace Xamarin.Toolkit.Animations
     public abstract class AnimationBase : BindableObject
     {
         public static readonly BindableProperty TargetProperty =
-            BindableProperty.Create(nameof(Target), typeof(VisualElement), typeof(AnimationBase), null,
-                BindingMode.TwoWay, null);
+            BindableProperty.Create(
```
bartlannoeye: not a fan of this, takes too much lines => less useful code on my screen

same remark can be made on quite a few files below

src/Animations/Animations/Shared/Animations/StoryBoard.cs
```diff
@@ -27,7 +27,9 @@ public List<AnimationBase> Animations
             foreach (var animation in Animations)
             {
                 if (animation.Target == null)
+                {
```
bartlannoeye: Is this a general rule or can it be specified for specific cases? I don't mind dropping braces in case of exception / return. Adding in this particular case is ok for me.

src/Animations/Animations/Shared/Triggers/BeginAnimation.cs
```diff
@@ -9,7 +9,9 @@ public class BeginAnimation : TriggerAction<VisualElement>
         protected override async void Invoke(VisualElement sender)
         {
             if (Animation == null)
+            {
```
bartlannoeye: see above, here I would keep without braces

src/Behaviors/Behaviors/Shared/Entry/EntryEmailValidation.cs
```diff
@@ -9,24 +9,24 @@ namespace Xamarin.Toolkit.Behaviors
...
+              Color.Default);
 
         /// <summary>
         /// The is valid property.
         /// </summary>
         public static readonly BindableProperty TextColorInvalidProperty =
             TextColorInvalidKey.BindableProperty;
-        
+
+        private bool colorSet;
```
bartlannoeye: why are these placed between 2 private statics?

src/Behaviors/Behaviors/Shared/Entry/EntryEmptyValidation.cs
```diff
@@ -8,20 +8,21 @@ namespace Xamarin.Toolkit.Behaviors
...
+              typeof(EntryEmptyValidation),
+              Color.Default);
 
         /// <summary>
         /// The is valid property.
         /// </summary>
         public static BindableProperty TextColorInvalidProperty =
             TextColorInvalidKey.BindableProperty;
 
+        private bool colorSet;
```
bartlannoeye: idem

src/Behaviors/Behaviors/Shared/Entry/EntryEmptyValidation.cs
```diff
@@ -74,24 +78,25 @@ protected override void OnDetachingFrom(Entry bindable)
...
-        void HandleTextChanged(object sender, TextChangedEventArgs e)
+        private void HandleTextChanged(object sender, TextChangedEventArgs e)
         {
             var text = e?.NewTextValue ?? string.Empty;
             IsValid = !string.IsNullOrWhiteSpace(text);
 
             var entry = sender as Entry;
 
             if (entry == null)
+            {
```
bartlannoeye: braces

src/Behaviors/Behaviors/Shared/Entry/NumericValidationBehavior.cs
```diff
@@ -6,19 +6,21 @@ namespace Xamarin.Toolkit.Behaviors
...
+                typeof(EntryEmailValidation),
+                Color.Default);
 
         /// <summary>
         /// The is valid property.
         /// </summary>
         public static readonly BindableProperty TextColorInvalidProperty =
             TextColorInvalidKey.BindableProperty;
 
+        private bool colorSet;
```
bartlannoeye: idem

src/Animations/Animations/Shared/Animations/Base/AnimationBase.cs
jgiacomini: The stylecop rule say : 
 - all parameters must be in one line
- or one parameter by line

Here => we have a mix 4 parameters in first ligne and 1 parameter in last line

src/Animations/Animations/Shared/Animations/StoryBoard.cs
jgiacomini: Yes => in the file Toolkit.ruleset

src/Behaviors/Behaviors/Shared/Entry/EntryEmptyValidation.cs
jgiacomini: A rule specify you have to put braces. :)
You can disable it if you want. It's a guideline.

src/Animations/Animations/Shared/Animations/Base/AnimationBase.cs
jgiacomini: I think we have to think if we take the same rules of UWPCommunityToolkit or if we use custom rules for this project.

The UWPCommunityToolkit rules =>
https://github.com/Microsoft/UWPCommunityToolkit/blob/master/Toolkit.ruleset

jgiacomini:
I think I didn't well proceed.
I will revert all modification on code to fix the warnings. (commit 2)

We have to talk about which rules we take on it.

I propose to debate for each already selected "warning". 

All code rule selected are here 
https://github.com/Microsoft/UWPCommunityToolkit/blob/master/Toolkit.ruleset

If you open this file with VS you can have description of each rule.

For example  =>
CA1001: Types that own disposable fields should be disposable 
https://msdn.microsoft.com/en-US/library/ms182172.aspx

bartlannoeye:
It is indeed easier to discuss before putting the effort into it (and most of us are already very busy on other projects, so discussion goes slow). Sorry for that. Nevertheless, we still appreciate your engagement!

jamesmontemagno:
I have created an issue for this: https://github.com/xamarin/XamarinCommunityToolkit/issues/77

We should follow and integrate this stylecop: https://github.com/xamarin/Caboodle

jamesmontemagno:
I tied this all into #82 and put in our xamarin style cop and fixed up all of the files :)

PR #99, opened 2018-06-19
Update SwitchChangeColor.android.cs

### LINKED ISSUE #98: Switch effect starts always with switch off color
## Bug 

Version Number of Plugin: 0.0.1.34-preview
Device Tested On: Huawei Nexus 6p
Simulator Tested On: Nexus 6p AVD

### Expected Behavior
When you use the SwitchChangeColor effect, you expect the switch to have the true-color when the switch is initialized to checked.

### Actual Behavior
The switch is always initialized to the false-color.

### Steps to reproduce the Behavior
Add the SwitchChangeColor effect to a switch in xaml and initialize that switch to checked

squibel:
Fixes #98  .

- Update SwitchChangeColor.android.cs to set the starting color to the correct color depending the isChecked state instead of the default falseColor


bartlannoeye:
@jamesmontemagno build got stuck? Or not reporting back correctly?

pauldipietro:
Yeah, something's going on with the VSTS build that I have to look into. @jamesmontemagno might also have some ideas.

PR #112, opened 2019-09-02
Support VS Mac 8.3

mrward:
Fix some things which were preventing multi-target code completion from working with VS Mac 8.3.

- Remove duplicate None items preventing code completion from working properly.
- Lower case target frameworks in conditions which was preventing files from being included in code completion.
- Update to MSBuild.Sdk.Extras 2.0.41
- Use .NET Core 2.1.701 sdk. Some problems building when .NET Core 3.0 is installed (A numeric comparison was attempted on "$(_TargetFrameworkVersionWithoutV)" error).

These changes allow me to build the solution in VS Mac, code completion working, and the tests pass. Not tried the sample solutions.

Not sure if you want to tie this to .NET Core 2.1.701. That change is in a separate commit.

### DIFF
--- a/Toolkit/Animations/Animations.csproj
+++ b/Toolkit/Animations/Animations.csproj
@@ -9,7 +9,6 @@
     <Description>Xamarin Community Toolkit Animations Library</Description>
     <EnableDefaultCompileItems>true</EnableDefaultCompileItems>
   </PropertyGroup>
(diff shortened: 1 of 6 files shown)


jfversluis:
Rebooting this project I think this is a bit outdated. Thanks for any effort!

PR #120, opened 2020-05-31
Basic project setup & added some basic converters

### LINKED ISSUE #119: [Enhancement] Enable Source Link
Let's make sure we have Source Link enabled from day 1!

Instructions on how to do that are here: https://github.com/dotnet/sourcelink/blob/master/README.md

sthewissen:
### Description of Change ###

Added a multi-targeting `.csproj` file targeting some of the more popular platforms. I also added a few basic converters that are useful as a starting point. Has no tests or sample just yet. Unit tests should probably be added and we can already think of a nice sample app.

### Bugs Fixed ###

- Fixed #119 

### API Changes ###

Added: 
- `InvertedBooleanConverter`
- `IsNullOrEmptyConverter`
- `IsNotNullOrEmptyConverter`

### PR Checklist ###

- [x] Rebased on top of master at time of PR
- [ ] Updated documentation


sthewissen:
- How do we want to handle converters that are basically the opposite of another converter? Just create a copy and do the inversion or could we come up with something more elegant through ConverterParameter or an additional property?
- Do we want to support things like Apple Watch, Tizen etc.
- Are we going for one big NuGet package or do we separate by type? (converters/effects etc.)

AndreiMisiukevich:
@sthewissen 

*How do we want to handle converters that are basically the opposite of another converter? Just create a copy and do the inversion or could we come up with something more elegant through ConverterParameter or an additional property?* **The first approach looks good to me because every converter will be independent of each other and we will avoid one additional step with the "unboxing" of converter's parameter**.

*Do we want to support things like Apple Watch, Tizen etc.* **Can we do it partially? I mean all cross-platform stuff can be used on those platforms.**
*Are we going for one big NuGet package or do we separate by type? (converters/effects etc.)* **Good question _________**
*Who will be the "Author" for this package on NuGet as it is a community effort with potentially a lot of authors?* **I guess Microsoft ?**

XamarinCommunityToolkit/Converters/StringIsNotNullOrEmptyConverter.shared.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Globalization;
+using Xamarin.Forms;
+
+namespace XamarinCommunityToolkit.Converters
+{
+    public class StringIsNotNullOrEmptyConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture) => !string.IsNullOrEmpty(value?.ToString());
```
AndreiMisiukevich, line 9: Perhaps, it makes sense to cast the value to "string" explicitly, because it another case that's possible to use this converted with any type and you won't get an exception.

XamarinCommunityToolkit/Converters/StringIsNotNullOrEmptyConverter.shared.cs
```diff
@@ -0,0 +1,16 @@
...
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
```
AndreiMisiukevich, line 11: How about to create a base internal abstract class "BaseConverter" with non-implemented "ConvertBack". It will allow us to not copy-paste "ConvertBack" everywhere

bartlannoeye: I wouldn't increase the object dependency tree (which brings a minimal performance hit) just to prevent a NotImplementedException, which you don't even have to copy as it's generated by default by implementing the interface.

bartlannoeye:
> Are we going for one big NuGet package or do we separate by type? (converters/effects etc.)

Multiple packages is somewhat cleaner as you only pull in what you want, on the other hand it brings a lot of overhead as well (both producer and consumer side). In the end @jamesmontemagno did merge all his tools into Xamarin.Essentials as well.

XamarinCommunityToolkit/Converters/StringIsNotNullOrEmptyConverter.shared.cs
AndreiMisiukevich: > I wouldn't increase the object dependency tree (which brings a minimal performance hit) just to prevent a NotImplementedException, which you don't even have to copy as it's generated by default by implementing the interface.

I believe this performance hit is insignificant.

roubachof: I tend to agree with @bartlannoeye here.

sthewissen:
I'm inclined to agree with @bartlannoeye regarding single library. The simple single include in XAML pages makes things cleaner and the additional overhead from having packaged a few additional `.cs` files in the binary is insignificant to me.

AndreiMisiukevich:
@sthewissen what do you think about making a couple of converters more common?
- StringIsNullOrEmptyConverter -> **IsNullOrEmptyConverter**
```=> value == null || (value is string str && string.IsNullOrWhiteSpace(str));```
- StringIsNotNullOrEmptyConverter -> **IsNotNullOrEmptyConverter**

File: XamarinCommunityToolkit/Converters/StringIsNotNullOrEmptyConverter.shared.cs (hunk 2 above)
Comment (sthewissen) [reply XamarinCommunityToolkit/Converters/StringIsNotNullOrEmptyConverter.shared.cs]: I also agree that it brings more clarity to have it say right there where it matters that that conversion back is not possible instead of hiding it away in a base class for the sake of abstraction.

File: XamarinCommunityToolkit/Converters/IsNotNullOrEmptyConverter.shared.cs [hunk 3]
```diff
@@ -0,0 +1,28 @@
...
+        /// <summary>
+        /// Converts the incoming value to a boolean indicating whether or not the value is not null and not empty.
+        /// </summary>
+        /// <param name="value">The value to convert.</param>
+        /// <param name="targetType">The type of the binding target property.</param>
+        /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>A boolean indicating if the incoming value is not null and not empty.</returns>
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture) =>
+            value != null && ((value is string str && !string.IsNullOrWhiteSpace(str)) || !(value is string));
```
AndreiMisiukevich, line 21: :)
```suggestion
            value is string str ? !string.IsNullOrWhiteSpace(str) : value != null;
```

XamarinCommunityToolkit/Converters/IsNotNullOrEmptyConverter.shared.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            throw new NotImplementedException();
+        }
```
AndreiMisiukevich, line 26:
```suggestion
        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture) =>
            throw new NotImplementedException();
```

XamarinCommunityToolkit/Converters/IsNullOrEmptyConverter.shared.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>A boolean indicating if the incoming value is null or empty.</returns>
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture) =>
+            value == null || (value is string str && string.IsNullOrWhiteSpace(str));
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            throw new NotImplementedException();
+        }
```
AndreiMisiukevich, line 26:
```suggestion
        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture) =>
            throw new NotImplementedException();
```

XamarinCommunityToolkit/XamarinCommunityToolkit.csproj
```diff
@@ -0,0 +1,107 @@
...
+        -->
+    </ItemGroup>
+    <ItemGroup Condition=" $(TargetFramework.StartsWith('MonoAndroid')) ">
+        <Compile Include="**\*.android.cs" />
+        <Compile Include="**\*.android.*.cs" />
+    </ItemGroup>
+    <!-- NOTE: When Android 11.x comes out we neeed to make this check more robust -->
+    <ItemGroup Condition="$(TargetFramework.StartsWith('MonoAndroid')) And '$(TargetFramework)' != 'MonoAndroid10.0'">
+        <PackageReference Include="Xamarin.Android.Support.Compat" Version="28.0.0.3" />
+        <PackageReference Include="Xamarin.Android.Support.CustomTabs" Version="28.0.0.3" />
```
jfversluis, line 82: Do we actually need these references? Or others listed below. I'd like as less dependencies as possible

XamarinCommunityToolkit/XamarinCommunityToolkit.csproj
```diff
@@ -0,0 +1,107 @@
...
+        <DebugSymbols>true</DebugSymbols>
+    </PropertyGroup>
+    <PropertyGroup Condition=" '$(Configuration)'=='Release' And '$(OS)' == 'Windows_NT' ">
+        <!-- Publish the repository URL in the built .nupkg (in the NuSpec <Repository> element) -->
+        <PublishRepositoryUrl>true</PublishRepositoryUrl>
+        <!-- Embed source files that are not tracked by the source control manager in the PDB -->
+        <EmbedUntrackedSources>true</EmbedUntrackedSources>
+        <!-- Build symbol package (.snupkg) to distribute the PDB containing Source Link -->
+        <IncludeSymbols>true</IncludeSymbols>
+        <SymbolPackageFormat>snupkg</SymbolPackageFormat>
```
jfversluis, line 48: I think snupkg's aren't supported for Xamarin at this point. I don't think it will do any harm though. Just leaving this here as a reminder when things don't work we might need to look here.

XamarinCommunityToolkit/XamarinCommunityToolkit.csproj
```diff
@@ -0,0 +1,107 @@
...
+    </PropertyGroup>
+    <ItemGroup Condition=" '$(Configuration)'=='Release' And '$(OS)' == 'Windows_NT' ">
```
jfversluis, line 50: Why only on Windows?

XamarinCommunityToolkit/XamarinCommunityToolkit.csproj
```diff
@@ -0,0 +1,107 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="MSBuild.Sdk.Extras/2.0.54">
+    <PropertyGroup>
+        <TargetFrameworks>netstandard1.0;netstandard2.0;Xamarin.iOS10;MonoAndroid80;MonoAndroid81;MonoAndroid90;MonoAndroid10.0;Xamarin.TVOS10;Xamarin.WatchOS10;tizen40</TargetFrameworks>
+        <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);uap10.0.16299</TargetFrameworks>
```
AndreiMisiukevich, line 5: Maybe let's enable C# 8?

```xml
<LangVersion>8.0</LangVersion>
```

AndreiMisiukevich:
@sthewissen @jfversluis congratulations with the 1st one :)

PR #124, opened 2020-06-04
Set up CI with Azure Pipelines

jfversluis:


azure-pipelines.yml
```diff
@@ -0,0 +1,96 @@
+
+variables:
+  CurrentSemanticVersionBase: '1.5.3'
+  PreviewNumber: $[counter(variables['CurrentSemanticVersionBase'], 1001)]
+  CurrentSemanticVersion: '$(CurrentSemanticVersionBase)-preview$(PreviewNumber)'
+  NugetPackageVersion: '$(CurrentSemanticVersion)'
+  MONO_VERSION: 6_4_0
+  XCODE_VERSION: 11.1
```
AndreiMisiukevich, line 8: 11.5 is the current stable version, but I don't know if it's available on Azure

jfversluis: Yep! Don't think so, but I'm still working on this one :)

jfversluis: Updated to macOS 10.15 and Xcode 11.4 which seems to be the default on the Hosted Mac agents for now

jfversluis:
OK, for I first step I think this is good now. Just CI for now and publishing the NuGets to the Azure DevOps server.

I guess it would be nice to get a nightly build up on GitHub packages here and then whenever we tag something let that go to actual NuGet.

azure-pipelines.yml
```diff
@@ -0,0 +1,141 @@
...
+          inputs:
+            scanType: LogOnly
+        - task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
+          condition: startsWith(variables['Build.SourceBranch'], 'refs/tags/')
+          displayName: Component Detection - Report
+        - task: securedevelopmentteam.vss-secure-development-tools.build-task-policheck.PoliCheck@1
+          condition: eq(variables['RunPoliCheck'], 'true')
+          displayName: 'PoliCheck'
+          inputs:
+            targetType: F
```
jfversluis, line 99: @jamesmontemagno this looked important, is this OK or do I need to do anything else for this?

PR #125, opened 2020-06-04
Lower Forms dependency to 3.5

jfversluis:
### Description of Change ###

Like it says in the title! Let's start out with a low dependency and see if we need to move up whenever we need something specific in newer Forms versions.

This should also fix the build error we're seeing right now


jfversluis:
I'm going to go ahead and use my super powers to just merge this at this stage in the project...

AndreiMisiukevich:
> I'm going to go ahead and use my super powers to just merge this at this stage in the project...

👍

Issue #116, opened 2020-05-29
[Proposal] Converters


jfversluis:
## Feature Request:

From the thread [here](https://twitter.com/devnl/status/1266109538580795402) it seems like people have some converters that they use all the time. Let's come up with some of the most basic/most wanted and get them in.

From the thread:
* InvertBoolConverter
* [EmptyNullToBoolConverter](https://github.com/pingzing/trippit/blob/master/Trippit/Converters/EmptyNullToBoolConverter.cs)
* DateTime Converters
* CollectionToBoolConverter
* UnitToBool

(More) examples: https://github.com/DIPSAS/DIPS.Xamarin.UI/tree/master/src/DIPS.Xamarin.UI/Converters/ValueConverters

Depechie:
I like this proposal... and if I can be so bold, I think Pedro Lamas could be a prime candidate to do a big PR for a lot of them.
His Cimbalino toolkit ( https://github.com/Cimbalino/Cimbalino-Toolkit/tree/master/src/Cimbalino.Toolkit%20(WP8)/Converters ) contains a huge list that were very relevant for Windows Phone and UWP.
I know he was looking at Xamarin when it came into Microsoft, but never got around it... but much of the converters are platform agnostic.

haavamoa:
I think that converters should implement `IMarkupExtension` to make them more accessible from XAML and the consumer would not have to add it to their resources for every page. It also enables you to pass properties, like a bool `Inverted` to invert it. 

Here is a couple of converters we are using in our apps :
https://github.com/DIPSAS/DIPS.Xamarin.UI/wiki/Converters

jfversluis:
OK, I think we got a lot of these in now. From here on out let's see what people come up with before we stuff more in.

PR #128, opened 2020-06-08
Added converters | Equal | NotEqual | ListIsNullOrEmpty | ListIsNotNullOrEmpty

sthewissen:
### Description of Change ###

Added a few more common converters.

### Bugs Fixed ###

- Related to issue #116

### API Changes ###

Added: 
 
- `EqualConverter`
- `NotEqualConverter`
- `ListIsNullOrEmptyConverter`
- `ListIsNotNullOrEmptyConverter`

### PR Checklist ###

- [x] Rebased on top of master at time of PR
- [ ] Updated documentation

### DIFF
--- a/XamarinCommunityToolkit/Converters/EqualsConverter.shared.cs
+++ b/XamarinCommunityToolkit/Converters/EqualsConverter.shared.cs
@@ -17,8 +17,10 @@ namespace XamarinCommunityToolkit.Converters
         /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
         /// <param name="culture">The culture to use in the converter.</param>
         /// <returns>An inverted boolean from the one coming in.</returns>
-        public object Convert(object value, Type targetType, object parameter, CultureInfo culture) => value.Equals(parameter);
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+            => value != null && value.Equals(parameter);
 
-        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture) => throw new NotImplementedException();
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+            => throw new NotImplementedException();
     }
 }
--- a/XamarinCommunityToolkit/Converters/EqualsConverter.shared.cs
+++ b/XamarinCommunityToolkit/Converters/EqualsConverter.shared.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.Globalization;
+using Xamarin.Forms;
+
+namespace XamarinCommunityToolkit.Converters
+{
+    /// <summary>
+    /// Checks whether the incoming value equals the provided parameter.
+    /// </summary>
+    public class EqualsConverter : IValueConverter
+    {
+        /// <summary>
+        /// Checks whether the incoming value equals the provided parameter.
+        /// </summary>
+        /// <param name="value">The value to convert.</param>
+        /// <param name="targetType">The type of the binding target property.</param>
+        /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>An inverted boolean from the one coming in.</returns>
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture) => value.Equals(parameter);
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture) => throw new NotImplementedException();
+    }
+}
--- a/XamarinCommunityToolkit/Converters/InvertedBoolConverter.shared.cs
+++ b/XamarinCommunityToolkit/Converters/InvertedBoolConverter.shared.cs
@@ -22,7 +22,7 @@ namespace XamarinCommunityToolkit.Converters
             if (value is bool result)
                 return !result;
 
-            throw new ArgumentException($"Value is not a valid boolean", nameof(value));
+            throw new ArgumentException("Value is not a valid boolean", nameof(value));
         }
 
         /// <summary>
@@ -38,7 +38,7 @@ namespace XamarinCommunityToolkit.Converters
             if (value is bool result)
                 return !result;
 
-            throw new ArgumentException($"Value is not a valid boolean", nameof(value));
+            throw new ArgumentException("Value is not a valid boolean", nameof(value));
         }
     }
 }
--- a/XamarinCommunityToolkit/Converters/InvertedBoolConverter.shared.cs
+++ b/XamarinCommunityToolkit/Converters/InvertedBoolConverter.shared.cs
@@ -18,12 +18,7 @@ namespace XamarinCommunityToolkit.Converters
         /// <param name="culture">The culture to use in the converter.</param>
         /// <returns>An inverted boolean from the one coming in.</returns>
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
-        {
-            if (value is bool result)
-                return !result;
-
-            throw new ArgumentException("Value is not a valid boolean", nameof(value));
-        }
+            => InverseBool(value);
 
         /// <summary>
         /// Converts a boolean to its inverse value.
@@ -34,6 +29,14 @@ namespace XamarinCommunityToolkit.Converters
         /// <param name="culture">The culture to use in the converter.</param>
         /// <returns>An inverted boolean from the one coming in.</returns>
         public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+            => InverseBool(value);
+
+        /// <summary>
+        /// Inverses an incoming boolean.
+        /// </summary>
+        /// <param name="value">The value to inverse.</param>
+        /// <returns>The inverted value of the incoming boolean.</returns>
+        bool InverseBool(object value)
         {
             if (value is bool result)
                 return !result;
--- a/XamarinCommunityToolkit/Converters/InvertedBoolConverter.shared.cs
+++ b/XamarinCommunityToolkit/Converters/InvertedBoolConverter.shared.cs
@@ -18,7 +18,7 @@ namespace XamarinCommunityToolkit.Converters
         /// <param name="culture">The culture to use in the converter.</param>
         /// <returns>An inverted boolean from the one coming in.</returns>
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
-            => !(bool)value;
+            => !(value as bool?);
 
         /// <summary>
         /// Converts a boolean to its inverse value.
@@ -29,6 +29,6 @@ namespace XamarinCommunityToolkit.Converters
         /// <param name="culture">The culture to use in the converter.</param>
         /// <returns>An inverted boolean from the one coming in.</returns>
         public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
-            => !(bool)value;
+            => !(value as bool?);
     }
 }
--- a/XamarinCommunityToolkit/Converters/InvertedBoolConverter.shared.cs
+++ b/XamarinCommunityToolkit/Converters/InvertedBoolConverter.shared.cs
@@ -18,7 +18,12 @@ namespace XamarinCommunityToolkit.Converters
         /// <param name="culture">The culture to use in the converter.</param>
         /// <returns>An inverted boolean from the one coming in.</returns>
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
-            => !(value as bool?);
+        {
+            if (value is bool result)
+                return !result;
+
+            throw new ArgumentException($"Value is not a valid boolean", nameof(value));
+        }
 
         /// <summary>
         /// Converts a boolean to its inverse value.
@@ -29,6 +34,11 @@ namespace XamarinCommunityToolkit.Converters
         /// <param name="culture">The culture to use in the converter.</param>
         /// <returns>An inverted boolean from the one coming in.</returns>
         public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
-            => !(value as bool?);
+        {
+            if (value is bool result)
+                return !result;
+
+            throw new ArgumentException($"Value is not a valid boolean", nameof(value));
+        }
     }
 }
--- a/XamarinCommunityToolkit/Converters/InvertedBoolConverter.shared.cs
+++ b/XamarinCommunityToolkit/Converters/InvertedBoolConverter.shared.cs
@@ -7,7 +7,7 @@ namespace XamarinCommunityToolkit.Converters
     /// <summary>
     /// Converts true to false and false to true.
     /// </summary>
-    public class InvertedBooleanConverter : IValueConverter
+    public class InvertedBoolConverter : IValueConverter
     {
         /// <summary>
         /// Converts a boolean to its inverse value.
--- a/XamarinCommunityToolkit/Converters/InvertedBooleanConverter.shared.cs
+++ b/XamarinCommunityToolkit/Converters/InvertedBooleanConverter.shared.cs
@@ -17,7 +17,8 @@ namespace XamarinCommunityToolkit.Converters
         /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
         /// <param name="culture">The culture to use in the converter.</param>
         /// <returns>An inverted boolean from the one coming in.</returns>
-        public object Convert(object value, Type targetType, object parameter, CultureInfo culture) => !(bool)value;
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+            => !(bool)value;
 
         /// <summary>
         /// Converts a boolean to its inverse value.
@@ -27,6 +28,7 @@ namespace XamarinCommunityToolkit.Converters
         /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
         /// <param name="culture">The culture to use in the converter.</param>
         /// <returns>An inverted boolean from the one coming in.</returns>
-        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture) => !(bool)value;
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+            => !(bool)value;
     }
 }
--- a/XamarinCommunityToolkit/Converters/ListIsNotNullOrEmptyConverter.shared.cs
+++ b/XamarinCommunityToolkit/Converters/ListIsNotNullOrEmptyConverter.shared.cs
@@ -1,0 +1,26 @@
+﻿using System;
+using System.Collections;
+using System.Globalization;
+using Xamarin.Forms;
+using System.Linq;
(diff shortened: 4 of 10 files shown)


XamarinCommunityToolkit/Converters/ListIsNullOrEmptyConverter.shared.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary>
+        /// Converts the incoming value to a boolean indicating whether or not the value is null or empty.
+        /// </summary>
+        /// <param name="value">The value to convert.</param>
+        /// <param name="targetType">The type of the binding target property.</param>
+        /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>A boolean indicating if the incoming value is null or empty.</returns>
+		public object Convert(object value, Type targetType, object parameter, CultureInfo culture) => value == null || (value is IEnumerable list && !list.GetEnumerator().MoveNext());
```
AndreiMisiukevich, line 21: Please fix the indention.

XamarinCommunityToolkit/Converters/ListIsNullOrEmptyConverter.shared.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture) => null;
```
AndreiMisiukevich, line 23:
```suggestion
        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture) => throw new NotImplementedException();;
```

?

XamarinCommunityToolkit/Converters/ListIsNotNullOrEmptyConverter.shared.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        /// <summary>
+        /// Converts the incoming value to a boolean indicating whether or not the value is not null and not empty.
+        /// </summary>
+        /// <param name="value">The value to convert.</param>
+        /// <param name="targetType">The type of the binding target property.</param>
+        /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>A boolean indicating if the incoming value is not null and not empty.</returns>
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture) => value != null && value is IEnumerable list && list.GetEnumerator().MoveNext();
```
AndreiMisiukevich, line 22: can we change ```list.GetEnumerator().MoveNext()``` to ```list.Any()``` ?

sthewissen: `IEnumerable` does not have that, only `IEnumerable<T>` does.

jfversluis:
@sthewissen can you rebase this please, then the CI should also pick this up. Thanks :)

XamarinCommunityToolkit/Converters/EqualsConverter.shared.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        /// <summary>
+        /// Checks whether the incoming value equals the provided parameter.
+        /// </summary>
+        /// <param name="value">The value to convert.</param>
+        /// <param name="targetType">The type of the binding target property.</param>
+        /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>An inverted boolean from the one coming in.</returns>
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture) => value.Equals(parameter);
```
AndreiMisiukevich, line 20: I wonder, should we use ```culture``` parameter for strings comparison? Not it's ignored.

XamarinCommunityToolkit/Converters/ListIsNotNullOrEmptyConverter.shared.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture) => throw new NotImplementedException();
```
AndreiMisiukevich, line 24: We should reach an agreement regarding the code style. I mean perhaps it makes sense to write **=>** from the new line.

```csharp
public object Method(object parameter)
    => new object();
```

XamarinCommunityToolkit/Converters/EqualsConverter.shared.cs
AndreiMisiukevich, line 20: possible NullReferenceException if value == null

XamarinCommunityToolkit/Converters/ListIsNotNullOrEmptyConverter.shared.cs
sthewissen: Agreed. Will put that in for each.

XamarinCommunityToolkit/Converters/EqualsConverter.shared.cs
sthewissen: I guess the question also is; do we start a completely new means of checking for string equality within this converter or do we start a different one for it? And are there other types that need "special treatment" that we might need to take into consideration?

XamarinCommunityToolkit/Converters/EqualsConverter.shared.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// <summary>
+        /// Checks whether the incoming value equals the provided parameter.
+        /// </summary>
+        /// <param name="value">The value to convert.</param>
+        /// <param name="targetType">The type of the binding target property.</param>
+        /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>An inverted boolean from the one coming in.</returns>
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+            => value != null && value.Equals(parameter);
```
AndreiMisiukevich, line 21: I believe we should return **True** if both value and parameter equal **null**.

XamarinCommunityToolkit/Converters/ListIsNullOrEmptyConverter.shared.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary>
+        /// Converts the incoming value to a boolean indicating whether or not the value is null or empty.
+        /// </summary>
+        /// <param name="value">The value to convert.</param>
+        /// <param name="targetType">The type of the binding target property.</param>
+        /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>A boolean indicating if the incoming value is null or empty.</returns>
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture) => value == null || (value is IEnumerable list && !list.GetEnumerator().MoveNext());
```
AndreiMisiukevich, line 21: Should **=>** be translated to the new line, I believe)

XamarinCommunityToolkit/Converters/NotEqualsConverter.shared.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// <summary>
+        /// Checks whether the incoming value doesn't equal the provided parameter.
+        /// </summary>
+        /// <param name="value">The value to convert.</param>
+        /// <param name="targetType">The type of the binding target property.</param>
+        /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>An inverted boolean from the one coming in.</returns>
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+            => value == null || !value.Equals(parameter);
```
AndreiMisiukevich, line 21: I believe we should return **False** if both value and parameter are equal **null**.

XamarinCommunityToolkit/Converters/NotEqualsConverter.shared.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Globalization;
+using Xamarin.Forms;
+
+namespace XamarinCommunityToolkit.Converters
+{
+    /// <summary>
+    /// Checks whether the incoming value doesn't equal the provided parameter.
+    /// </summary>
+    public class NotEqualsConverter : IValueConverter
```
AndreiMisiukevich, line 10: **NotEqualConverter** ? Seems *s* is redundant here.

XamarinCommunityToolkit/Converters/ListIsNotNullOrEmptyConverter.shared.cs
```diff
@@ -19,8 +19,10 @@ public class ListIsNotNullOrEmptyConverter : IValueConverter
         /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
         /// <param name="culture">The culture to use in the converter.</param>
         /// <returns>A boolean indicating if the incoming value is not null and not empty.</returns>
-        public object Convert(object value, Type targetType, object parameter, CultureInfo culture) => value != null && value is IEnumerable list && list.GetEnumerator().MoveNext();
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+            => value != null && value is IEnumerable list && list.GetEnumerator().MoveNext();
```
Depechie, line 23: Use of value != is redundant because of the is IEnumerable check.

XamarinCommunityToolkit/Converters/EqualsConverter.shared.cs
sthewissen: I was always taught that `null` is equal to nothing and that even `null`  is not equal to `null` 🤔

XamarinCommunityToolkit/Converters/NotEqualsConverter.shared.cs
sthewissen: I mainly named it that to indicate that it's the exact opposite of the `EqualsConverter`. I've also considered naming it `UnequalsConverter` but I suppose this could also be a greater discussion regarding the naming of inverse converters?

- `ListIsNullOrEmptyConverter` and:
   - `NotListIsNullOrEmptyConverter` vs.
   - `ListIsNotNullOrEmptyConverter`

AndreiMisiukevich: @sthewissen I mean Equal**s** is a verb, but Equal is an adjective.

EqualConverter and NotEqualConverter.

XamarinCommunityToolkit/Converters/ListIsNotNullOrEmptyConverter.shared.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// <summary>
+        /// Converts the incoming value to a boolean indicating whether or not the value is not null and not empty.
+        /// </summary>
+        /// <param name="value">The value to convert.</param>
+        /// <param name="targetType">The type of the binding target property.</param>
+        /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>A boolean indicating if the incoming value is not null and not empty.</returns>
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+            => value is IEnumerable list && list.GetEnumerator().MoveNext();
```
AndreiMisiukevich, line 23: Should the converter raise "ArgumentException" if the provided value couldn't be cast to IEnumerable?

XamarinCommunityToolkit/Converters/EqualsConverter.shared.cs
AndreiMisiukevich: > I was always taught that `null` is equal to nothing and that even `null` is not equal to `null` 🤔

object x = null;
object y = null;
var result = x == y; // True;

XamarinCommunityToolkit/Converters/ListIsNotNullOrEmptyConverter.shared.cs
sthewissen: I'm not sure, to be honest. It feels like something as simple as a converter shouldn't crash someone's code if we can't quite manage to convert it. Not sure what other people think about this one? @Depechie @jfversluis @roubachof @bartlannoeye

roubachof: ahah! the classic silently fails vs crash in prod dilemma!

roubachof: In a ideal world we should write something in the error output to not just silently fail.

Depechie: If we don’t want to crash we also need to change the inverse bool converter because it does an implicit type casting instead of using the AS keyword. So the current version will also crash there if no bool was supplied... with AS you get null.

But sure what I like best to be honest :/ if it does not crash you can have a hard time searching for why some XAML is not working as it should...

XamarinCommunityToolkit/Converters/InvertedBooleanConverter.shared.cs
```diff
@@ -17,7 +17,8 @@ public class InvertedBooleanConverter : IValueConverter
         /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
```
AndreiMisiukevich, line 17: Inverted**Bool**Converter
Are you OK with this renaming?

XamarinCommunityToolkit/Converters/EqualsConverter.shared.cs
sthewissen: I'm not sure either, but... we don't need to make a decision right here and now. If the need arises in the community people can request this kind of stuff and/or add to it themselves right? If the need arises we can act on it then.

XamarinCommunityToolkit/Converters/ListIsNotNullOrEmptyConverter.shared.cs
sthewissen: Since this is probably not a discussion we'll manage to figure out on our own I'd say choose one and let the community decide. As per Glenn's last comment, I will change the `InverseBoolConverter` to have everything in line with one another so we can move forward.

XamarinCommunityToolkit/Converters/InvertedBoolConverter.shared.cs
```diff
@@ -17,7 +17,8 @@ public class InvertedBooleanConverter : IValueConverter
         /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
         /// <param name="culture">The culture to use in the converter.</param>
         /// <returns>An inverted boolean from the one coming in.</returns>
-        public object Convert(object value, Type targetType, object parameter, CultureInfo culture) => !(bool)value;
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+            => !(value as bool?);
```
AndreiMisiukevich, line 21: if the value is **null* than "null" will be returned as the result from the method.
Do we actually want to have this behavior?

sthewissen: If a user provides something that's not a boolean we cannot invert it. In my opinion, we shouldn't return `false` as that suggests some inversion actually took place but it hasn't. Returning null at least indicates that it couldn't be inverted and people would be triggered earlier that something is going wrong without immediately crashing their app.

AndreiMisiukevich: Got your point.
So we decided to go ahead with "no crash" approach.
I don't want to stop the works and if everyone is fine with this approach, then I am fine too.

@jfversluis @roubachof @Depechie

roubachof: I'm still pretty convinced that we should log any error to output windows. It could lead to very painful debugging when some code silently fails.

jfversluis: Yeah that would certainly be nice. There was something in place for that, but not sure if that still/has ever worked perfectly.

This was added, but I think I also saw an issue of it not working: https://blog.verslu.is/xamarin/xamarin-forms-xamarin/logwarningstoapplicationoutput-xamarin-forms/

I think this should still work: https://blog.verslu.is/xamarin/xamarin-forms-xamarin/binding-errors-output-window/

sthewissen: My point mainly is that we need to make a choice together _or_ we pick something and let the community decide later if that has their majority vote. Else we'll be sitting here with open PRs until eternity 😂

jfversluis: Then I would either say raise exception or figure out how to do the logging to the output window. The first option is probably fastest to implement at this point.

Users will be aware that something is up with their code if it fails

AndreiMisiukevich: @sthewissen :) agreed.

My main concern about this PR is that **null** equals **null** :) That's only one thing that stops me to approve this PR

XamarinCommunityToolkit/Converters/InvertedBoolConverter.shared.cs
```diff
@@ -27,6 +33,12 @@ public class InvertedBooleanConverter : IValueConverter
         /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
         /// <param name="culture">The culture to use in the converter.</param>
         /// <returns>An inverted boolean from the one coming in.</returns>
-        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture) => !(bool)value;
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            if (value is bool result)
```
AndreiMisiukevich, line 38: Since **Convert** and **ConvertBack** do the same, we probably could extract the logic to the separate private method to satisfy DRY principle, but uh

XamarinCommunityToolkit/Converters/NotEqualsConverter.shared.cs
AndreiMisiukevich, line 10: nitpick: NotEqualConverter (Equal**S**), since we have EqualConverter

AndreiMisiukevich: @sthewissen just this is quite important, I guess (Make names consistent)

EqualConverted & NotEqualConverter

Alright?

sthewissen: Hmm, thought I already fixed this one. Probably in my dreams then 🤔

PR #130, opened 2020-06-08
Added converters | IntToBool | DoubleToInt | BoolToObject | IndexToArrayItem | LowerCaseText | UpperCaseText

AndreiMisiukevich:
### Description of Change ###

Added a few more common converters.

**Note**: I haven't added comments yet, because I don't know which converters will be accepted.

### Bugs Fixed ###

- Related to issue #116

### API Changes ###

Added: 
 
- `IntToBoolConverter`
- `DoubleToIntConverter`
- `BoolToObjectConverter`
- `IndexToArrayItemConverter`
- `TextCaseConverter`

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in the description)
- [ ] Has samples (if omitted, state reason in the description)
- [X] Rebased on top of the master at the time of PR
- [X] Changes adhere to the coding standard
- [ ] Updated documentation


sthewissen:
I'm trying to think about what would be a use case for the `ToArray` converters. You already mentioned that they're probably the most confusing but I'm trying to find a way in my day to day that I would use these :D Do you happen to have a sample from the real world on how I could use these (apart from the conceptual one above)?

AndreiMisiukevich:
> I am afraid that if the converter's name is not self explainable, it will be never used.
> That seems the case for `BoolToArray`, `BoolToObject`, `IntToArray`

Yeah, so I am open for conversation :)

AndreiMisiukevich:
> I'm trying to think about what would be a use case for the `ToArray` converters. You already mentioned that they're probably the most confusing but I'm trying to find a way in my day to day that I would use these :D Do you happen to have a sample from the real world on how I could use these (apart from the conceptual one above)?

Seems **BoolToArray** doesn't make sense since BoolToObject exists.
As for **IntToArray**: for example, I have five color themes for my page. And by setting values from 0 to 4 I can pick a Color from the array.

XamarinCommunityToolkit/Converters/IntToBoolConverter.shared.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Globalization;
+using Xamarin.Forms;
+
+namespace XamarinCommunityToolkit.Converters
+{
+    public class IntToBoolConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+            => (int)value != 0;
```
Depechie, line 10: @sthewissen added throw new ArgumentException when the value being passed is not of the proposed type. So maybe keep that same pattern and check if the value passed is an int in this case.

XamarinCommunityToolkit/Converters/IntToBoolConverter.shared.cs
```diff
@@ -0,0 +1,15 @@
...
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+            => (bool)value ? 1 : 0;
```
Depechie, line 13: Same remark about adding throw new ArgumentException

XamarinCommunityToolkit/Converters/IntToArrayValueConverter.shared.cs
```diff
@@ -0,0 +1,29 @@
...
+using Xamarin.Forms;
+
+namespace XamarinCommunityToolkit.Converters
+{
+    public class IntToArrayValueConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            var array = (Array)parameter;
+            return array.GetValue((int)value);
```
Depechie, line 12: Same remark about adding throw new ArgumentException

XamarinCommunityToolkit/Converters/DoubleToIntConverter.shared.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Globalization;
+using Xamarin.Forms;
+
+namespace XamarinCommunityToolkit.Converters
+{
+    public class DoubleToIntConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+            => (int)Math.Round((double)value * GetParameter(parameter));
```
Depechie, line 10: Same remark about adding throw new ArgumentException

XamarinCommunityToolkit/Converters/BoolToObjectConverter.shared.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace XamarinCommunityToolkit.Converters
+{
+    public class BoolToObjectConverter<T> : IValueConverter
+    {
+        public T TrueObject { set; get; }
+
+        public T FalseObject { set; get; }
+
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
```
Depechie, line 13: Same remark about adding throw new ArgumentException

XamarinCommunityToolkit/Converters/BoolToArrayValueConverter.shared.cs
```diff
@@ -0,0 +1,35 @@
...
+using Xamarin.Forms;
+
+namespace XamarinCommunityToolkit.Converters
+{
+    public class BoolToArrayValueConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            var array = (Array)parameter;
+            return (bool)value ? array.GetValue(1) : array.GetValue(0);
```
Depechie, line 12: Same remark about adding throw new ArgumentException

XamarinCommunityToolkit/Converters/BoolToObjectConverter.shared.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Globalization;
+using Xamarin.Forms;
+
+namespace XamarinCommunityToolkit.Converters
+{
+    public class BoolToObjectConverter<T> : IValueConverter
```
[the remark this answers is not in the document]
AndreiMisiukevich:
```xaml
<Label>
    <Label.Text>
        <Binding Source="{x:Reference switch1}"
                 Path="IsToggled">
            <Binding.Converter>
                <local:BoolToObjectConverter x:TypeArguments="x:String"
                                             TrueObject="Of course!"
                                             FalseObject="No way!" />
            </Binding.Converter>
        </Binding>
    </Label.Text>
</Label>
```

AndreiMisiukevich:
All requests are resolved, please re-review.

XamarinCommunityToolkit/Converters/IndexToArrayItemConverter.shared.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System;
+using System.Globalization;
+using Xamarin.Forms;
+
+namespace XamarinCommunityToolkit.Converters
+{
+    /// <summary>
+    /// Converts an integer index to corresponding array item and vice versa.
+    /// </summary>
+    public class IndexToArrayItemConverter : IValueConverter
```
sthewissen, line 10: I like the rename 👍

roubachof: aaaah that's more like it :)

sthewissen:
> Added two new converters (Upper/Lower case)

Perhaps we could do a single TextCaseConverter that has an enum property defining the casing? That way we can expand in the future easier e.g. for CamelCase, TitleCase etc.

XamarinCommunityToolkit/Converters/DoubleToIntConverter.shared.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// </summary>
+        /// <param name="value">Double value.</param>
+        /// <param name="targetType">The type of the binding target property.</param>
+        /// <param name="parameter">Multiplier (Equals 1 by default).</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>Integer value.</returns>
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+            => value is double result
+            ? (int)Math.Round(result * GetParameter(parameter))
+            : throw new ArgumentException("Value is not a valid double", nameof(value));
```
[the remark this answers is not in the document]
sthewissen: I would add an additional tab in front of the ? and : 

Personal preference.

AndreiMisiukevich:
> > Added two new converters (Upper/Lower case)
> 
> Perhaps we could do a single TextCaseConverter that has an enum property defining the casing? That way we can expand in the future easier e.g. for CamelCase, TitleCase etc.

Very good point. I like this idea

AndreiMisiukevich:
> > Added two new converters (Upper/Lower case)
> 
> Perhaps we could do a single TextCaseConverter that has an enum property defining the casing? That way we can expand in the future easier e.g. for CamelCase, TitleCase etc.

Should we use converter parameter for this?

sthewissen:
> Should we use converter parameter for this?

We could do that or we could just introduce a typed property for it:

```
public class TextCaseConverter
{
    ...
    
    public TextCase TextCase { get; set; }

    ...
}
```

And use it as:

```
 <Label Text="{Binding MyString, Converter={xct:TextCaseConverter TextCase=UpperCase}}" />
```

AndreiMisiukevich:
@sthewissen Don't you think we should use converter parameter for this?

AndreiMisiukevich:
@sthewissen How about such rule
1) We try to parse parameter from the "ConverterParameter"
IF NO LUCK THEN
2) We take value from the property

AndreiMisiukevich:
Oh @sthewissen for doing as you described our converters should conform IMarkupExtension interface.

Perhaps it will work for us as implemented right now?

sthewissen:
> Oh @sthewissen for doing as you described our converters should conform IMarkupExtension interface.
> Perhaps it will work for us as implemented right now?

I don't see a problem in also implementing `IMarkupExtension` if that brings some more type-safety into the mix (and thus also Intellisense). I'm personally not a fan of the string matching / integer matching on a type `object` parameter as it doesn't really provide any feedback to the user from the IDE on what to put in.

To move forward with this PR and get some stuff done/in we could merge it, but I would consider the above.

AndreiMisiukevich:
> > Oh @sthewissen for doing as you described our converters should conform IMarkupExtension interface.
> > Perhaps it will work for us as implemented right now?
> 
> I don't see a problem in also implementing `IMarkupExtension` if that brings some more type-safety into the mix (and thus also Intellisense). I'm personally not a fan of the string matching / integer matching on a type `object` parameter as it doesn't really provide any feedback to the user from the IDE on what to put in.
> 
> To move forward with this PR and get some stuff done/in we could merge it, but I would consider the above.

I propose to merge it and create a separate ticket to make the necessary converters conforming IMarkupExtension.

jfversluis:
@AndreiMisiukevich Looks like this PR is ready to go, but you mention

> I haven't added comments yet, because I don't know which converters will be accepted.

Do you still want to add anything?

AndreiMisiukevich:
> @AndreiMisiukevich Looks like this PR is ready to go, but you mention
> 
> > I haven't added comments yet, because I don't know which converters will be accepted.
> 
> Do you still want to add anything?

No, I don't think so. I've already added comments

sthewissen:
> > > Oh @sthewissen for doing as you described our converters should conform IMarkupExtension interface.
> > > Perhaps it will work for us as implemented right now?
> > 
> > 
> > I don't see a problem in also implementing `IMarkupExtension` if that brings some more type-safety into the mix (and thus also Intellisense). I'm personally not a fan of the string matching / integer matching on a type `object` parameter as it doesn't really provide any feedback to the user from the IDE on what to put in.
> > To move forward with this PR and get some stuff done/in we could merge it, but I would consider the above.
> 
> I propose to merge it and create a separate ticket to make the necessary converters conforming IMarkupExtension.

Then LGTM.

PR #126, opened 2020-06-05
Expansion Pack: Sample App, Behaviors, & More

SunnyMukherjee:
### Description of Change ###

Change includes a list of the top common behavior classes used in Xamarin Forms apps.  Other Behaviors may exist out there that are needed in this toolkit.  I have personally used these Behaviors in my apps.

### Issues Fixed ###

- Related to issue #122 
- Related to issue #8 
- Related to issue #129 

### API Changes ###

Added sample app to test Behaviors

Added the following Behavior classes: 
 
- BaseBehavior
- DataChangedBehavior
- EffectBehavior
- EventToCommandBehavior
- LengthValidationBehavior
- RegularExpressionValidationBehavior
- AttachedNumericValidationBehavior
- MinMaxNumericValidationBehavior
- NumericValidationBehavior

PR #132, opened 2020-06-11
[POC] Validation behaviors

AndreiMisiukevich:
### Description of Change ###

Hey folks, I'd like to introduce the validator's proof of concept.

### Bugs Fixed ###
- Related to issue #8 

### API Changes ###
Added: 
 
**Abstractions**
- `BaseBehavior`
- `ValidationBehavior`
**Implementations**
- `TextValidationBehavior`
- `EmailValidationBehavior`
**Enums**
- `ValidationBehaviorFlags`

### Example ###
Jun-12-2020 04-25-42

```xaml
       <Entry>
           <Entry.Behaviors>
               <local:EmailValidationBehavior>
                   <local:EmailValidationBehavior.InvalidStyle>
                       <Style TargetType="Entry">
                           <Setter Property="TextColor" Value="Red" />
                       </Style>
                   </local:EmailValidationBehavior.InvalidStyle>
                </local:EmailValidationBehavior>
           </Entry.Behaviors>
       </Entry>
```

### PR Checklist ###

- [ ] Updated documentation


SunnyMukherjee:
Some of these behaviors exist in the other PR I started.  I'm already implementing the same logic in multiple behaviors.  Perhaps we should have a single PR.

AndreiMisiukevich:
> Some of these behaviors exist in the other PR I started.  I'm already implementing the same logic in multiple behaviors.  Perhaps we should have a single PR.

This PR is more about concept rather than exact implementation)

XamarinCommunityToolkit/Behaviors/LengthValidationBehavior.shared.cs
```diff
@@ -0,0 +1,30 @@
+﻿using Xamarin.Forms;
+
+namespace XamarinCommunityToolkit.Behaviors
+{
+    public class LengthValidationBehavior : TextValidationBehavior
+    {
+        public static readonly BindableProperty MinimumLengthProperty = BindableProperty.Create(nameof(MinimumLength), typeof(int), typeof(LengthValidationBehavior), 0, propertyChanged: OnValidationPropertyChanged);
+
+        public static readonly BindableProperty MaximumLengthProperty = BindableProperty.Create(nameof(MaximumLength), typeof(int), typeof(LengthValidationBehavior), int.MaxValue, propertyChanged: OnValidationPropertyChanged);
```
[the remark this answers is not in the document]
AndreiMisiukevich: I think, no. By default minimum length is 0 (it means that even empty string is valid) and maximum length is int.MaxValue (it means that even 2147483647 characters string is valid).

If you would like to change boundaries, just put desired values.

XamarinCommunityToolkit/Behaviors/EmailValidationBehavior.shared.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Text.RegularExpressions;
+
+namespace XamarinCommunityToolkit.Behaviors
+{
+    public class EmailValidationBehavior : TextValidationBehavior
+    {
+        static readonly Regex emailRegex = new Regex(@"\A(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*@(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?)\Z", RegexOptions.IgnoreCase);
```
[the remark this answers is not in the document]
AndreiMisiukevich: I believe, for archiving that the users can use **RegexValidationBehavior**

XamarinCommunityToolkit/Behaviors/LengthValidationBehavior.shared.cs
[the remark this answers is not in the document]
sthewissen: Hmm, not quite sure what I was thinking here, I was thinking of `integer` lengths for some reason but we're obviously talking `string` here. Dumb me.

XamarinCommunityToolkit/Behaviors/RegexValidationBehavior.shared.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        Regex Regex { get; set; }
+
+        public string Pattern
+        {
+            get => (string)GetValue(PatternProperty);
+            set => SetValue(PatternProperty, value);
+        }
+
+        public RegexOptions Options
```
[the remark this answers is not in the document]
AndreiMisiukevich: Since RegexOptions is an Enum, that's super convenient to set it through the xaml

```xaml
<local:RegexValidationBehavior Options="ExplicitCapture">
</local:RegexValidationBehavior>
```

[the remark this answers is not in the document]
sthewissen: I thought it was a class of some sorts with all sorts of properties in it. My bad.

AndreiMisiukevich:
> One additional thing I'm wondering is why we couldn't use the existing PR #126 that already made a start on validation and base off of that or steer it into the direction we wanted validation to go. Now we're essentially saying that the majority of that PR is not going to go in because we want to have it in a different way. We probably could've gone for the more cooperative way of bringing that PR to the point we wanted it together with the community contributor as we want to encourage community contributions. Don't get me wrong, there's not necessarily a right or wrong way here. This is just my opinion on how I see one of our main roles in this is to encourage contributions and make sure it fits with our direction.

Yes, I agree with you.
But I wanted to develop a flexible and scalable mechanism of validators.
This PR is about validators only (e.g. I didn't implement **EventToCommandBehavior** that is great but actually it's not a validator).

I wanted to make a groundwork only and added some validators for proofing the concept.

SunnyMukherjee:
> Just a few comments from my end (see below).
> 
> One additional thing I'm wondering is why we couldn't use the existing PR #126 that already made a start on validation and base off of that or steer it into the direction we wanted validation to go. Now we're essentially saying that the majority of that PR is not going to go in because we want to have it in a different way. We probably could've gone for the more cooperative way of bringing that PR to the point we wanted it together with the community contributor as we want to encourage community contributions. Don't get me wrong, there's not necessarily a right or wrong way here. This is just my opinion on how I see one of our main roles in this is to encourage contributions and make sure it fits with our direction.

Please let me know about the direction of #126 because I put a lot of work in it over this past week. I was hoping for it to get reviewed and approved at some point.  Now we have 2 overlapping PR's because both have similar validators.  That is why I suggested making any revisions or suggestions to that PR so the combined changes can be absorbed into master.  That PR is 99% complete as far as I'm concerned.  It is just left for approvals.

AndreiMisiukevich:
> Please let me know about the direction of #126 because I put a lot of work in it over this past week. I was hoping for it to get reviewed and approved at some point. Now we have 2 overlapping PR's because both have similar validators. That is why I suggested making any revisions or suggestions to that PR so the combined changes can be absorbed into master. That PR is 99% complete as far as I'm concerned. It is just left for approvals.

My opinion is that we should merge this PR as a ground.
Then apply your changes over this PR (EventToCommandBehavior, EffectBehavior and NumericValidationBehavior)

SunnyMukherjee:
> > Please let me know about the direction of #126 because I put a lot of work in it over this past week. I was hoping for it to get reviewed and approved at some point. Now we have 2 overlapping PR's because both have similar validators. That is why I suggested making any revisions or suggestions to that PR so the combined changes can be absorbed into master. That PR is 99% complete as far as I'm concerned. It is just left for approvals.
> 
> My opinion is that we should merge this PR as a ground.
> Then apply your changes over this PR (EventToCommandBehavior, EffectBehavior and NumericValidationBehavior)

I disagree.  I think your changes should be merged into #126 because either merging #126 before or after #132 would cause too many conflicts.  It would be easier to merge your changes as the base into #126 so there are fewer conflicts during the merge.  Besides, #126 has been going on for almost 2 weeks now.

AndreiMisiukevich:
> I disagree. I think your changes should be merged into #126 because either merging #126 before or after #132 is merged into master would cause too many conflicts since #132 refactors too much of the code. It would be easier to merge your changes as the base into #126 so there are fewer conflicts during the merge into master. All you would need to do is replace the BaseBehavior class in #126 and import your other Behaviors since they already inherit from your BaseBehavior.

I believe it's doesn't matter what the order of PR merging.
They both introduce new classes. And anyway some changes are required (removing BaseBehavior class for example).

No worries, I will help to resolve conflicts and clean code :)

AndreiMisiukevich:
@sthewissen @Depechie @roubachof please review it 👍

XamarinCommunityToolkit/XamarinCommunityToolkit.csproj
```diff
@@ -72,9 +66,7 @@
     <Compile Include="**/*.shared.*.cs" />
     <None Include="..\LICENSE" PackagePath="" Pack="true" />
     <Folder Include="Behaviors\" />
-    <Folder Include="Controls\" />
```
sthewissen, line 75: Why are these folders removed? I would expect Controls to still be around as `AvatarView` should be in there.

AndreiMisiukevich: There is another one "Views"

XamarinCommunityToolkit/Behaviors/BaseBehavior.shared.cs
```diff
@@ -1,74 +1,63 @@
...
-    /// </summary>
-    /// <typeparam name="T"></typeparam>
-    [Preserve(AllMembers = true)]
-    public class BaseBehavior<T> : Behavior<T> where T : BindableObject
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public abstract class BaseBehavior : Behavior<View>
     {
-        public static readonly BindableProperty ActionsProperty =
-            BindableProperty.Create(nameof(Actions), typeof(ActionCollection), typeof(BaseBehavior<BindableObject>), null);
+        static readonly MethodInfo getContextMethod = typeof(BindableObject).GetRuntimeMethods()?.FirstOrDefault(m => m.Name == "GetContext");
```
jfversluis, line 12: I'm assuming there is no other way than doing it like this? If we can make a case for why we want this, we could open a dialog with the Forms team to see if they can provide us with a way to access this more easily?

XamarinCommunityToolkit/Behaviors/EventToCommandBehavior.shared.cs
```diff
@@ -42,85 +44,56 @@ public object CommandParameter
...
 
-            var eventInfo = AssociatedObject.GetType().GetRuntimeEvent(name);
+            var eventName = EventName;
+            if (View == null || string.IsNullOrWhiteSpace(eventName))
+                return;
 
-            if (eventInfo == null)
-                throw new ArgumentException($"{nameof(EventToCommandBehavior)}: Can't de-register the {EventName} event.");
+            eventInfo = View.GetType().GetRuntimeEvent(eventName) ??
+                throw new ArgumentException($"{nameof(EventToCommandBehavior)}: Could't resolve the event.", nameof(EventName));
```
jfversluis, line 77:
```suggestion
                throw new ArgumentException($"{nameof(EventToCommandBehavior)}: Couldn't resolve the event.", nameof(EventName));
```

XamarinCommunityToolkit/Behaviors/EventToCommandBehavior.shared.cs
```diff
@@ -42,85 +44,56 @@ public object CommandParameter
...
-                bindable.BindingContext = BindingContext;
-                var action = (IAction)bindable;
-                await action.Execute(sender, eventArgs);
-            }
+            eventInfo?.RemoveEventHandler(View, eventHandler);
+            eventInfo = null;
         }
 
-        static void OnEventNameChanged(BindableObject bindable, object oldValue, object newValue)
+        void OnEventOccurred(object sender = null, object eventArgs = null)
```
jfversluis, line 88: I think in Forms the "Trigger<event>" is used mostly. Would it make sense to be more consistent and follow that? This is used to trigger the actual Command, so I think that makes a bit more sense

XamarinCommunityToolkit/Behaviors/EventToCommandBehavior.shared.cs
```diff
@@ -42,85 +44,56 @@ public object CommandParameter
...
-            var oldEventName = (string)oldValue;
-            var newEventName = (string)newValue;
+            var parameter = CommandParameter
+                ?? EventArgsConverter?.Convert(eventArgs, typeof(object), null, null)
+                ?? eventArgs;
 
-            behavior.DeregisterEvent(oldEventName);
-            behavior.RegisterEvent(newEventName);
+            var command = Command;
+            if (command?.CanExecute(parameter) ?? false)
```
jfversluis, line 95: What happens is `parameter == null` and the user did not specify any parameters on their command. Will this still work? Not sure how they solved that internally, maybe it will just ignore it because the value is null anyway. Is that something we can (unit) test?

XamarinCommunityToolkit/Behaviors/NumericValidationBehavior.shared.cs
```diff
@@ -1,66 +1,28 @@
...
 namespace XamarinCommunityToolkit.Behaviors
 {
-    /// <summary>
-    /// Validates if the text value of the Entry is numeric
-    /// </summary>
-    public class NumericValidationBehavior : BaseBehavior<Entry>
+    public class NumericValidationBehavior : ValidationBehavior
     {
-        Color currentTextColor;
+        public static readonly BindableProperty MinimumValueProperty = BindableProperty.Create(nameof(MinimumValue), typeof(double), typeof(NumericValidationBehavior), double.NegativeInfinity, propertyChanged: OnValidationPropertyChanged);
```
jfversluis, line 7: Can we add a break somewhere in this line for readability

XamarinCommunityToolkit/Behaviors/NumericValidationBehavior.shared.cs
```diff
@@ -1,66 +1,28 @@
...
 
-        /// <summary>
-        /// Bindable text color to apply when validation fails
-        /// </summary>
-        public static readonly BindableProperty TextColorProperty =
-            BindableProperty.Create(nameof(TextColor), typeof(Color), typeof(NumericValidationBehavior));
+        public static readonly BindableProperty MaximumValueProperty = BindableProperty.Create(nameof(MaximumValue), typeof(double), typeof(NumericValidationBehavior), double.PositiveInfinity, propertyChanged: OnValidationPropertyChanged);
```
jfversluis, line 9: Again; break line somewhere

XamarinCommunityToolkit/Behaviors/TextValidationBehavior.shared.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        public static readonly BindableProperty ShouldTrimValueProperty = BindableProperty.Create(nameof(ShouldTrimValue), typeof(bool), typeof(TextValidationBehavior), false, propertyChanged: OnValidationPropertyChanged);
+
+        public static readonly BindableProperty ShouldConvertNullToEmptyValueProperty = BindableProperty.Create(nameof(ShouldConvertNullToEmptyValue), typeof(bool), typeof(TextValidationBehavior), true, propertyChanged: OnValidationPropertyChanged);
+
+        public static readonly BindableProperty RegexPatternProperty = BindableProperty.Create(nameof(RegexPattern), typeof(string), typeof(TextValidationBehavior), defaultValueCreator: GetDefaultRegexPattern, propertyChanged: OnRegexPropertyChanged);
+
+        public static readonly BindableProperty RegexOptionsProperty = BindableProperty.Create(nameof(RegexOptions), typeof(RegexOptions), typeof(TextValidationBehavior), defaultValueCreator: GetDefaultRegexOptions, propertyChanged: OnRegexPropertyChanged);
```
jfversluis, line 18: Some breaks on these lines might improve readability

XamarinCommunityToolkit/Behaviors/TextValidationBehavior.shared.cs
```diff
@@ -0,0 +1,110 @@
...
+{
+    public class TextValidationBehavior : ValidationBehavior
+    {
+        public static readonly BindableProperty MinimumLengthProperty = BindableProperty.Create(nameof(MinimumLength), typeof(int), typeof(TextValidationBehavior), 0, propertyChanged: OnValidationPropertyChanged);
+
+        public static readonly BindableProperty MaximumLengthProperty = BindableProperty.Create(nameof(MaximumLength), typeof(int), typeof(TextValidationBehavior), int.MaxValue, propertyChanged: OnValidationPropertyChanged);
+
+        public static readonly BindableProperty ShouldTrimValueProperty = BindableProperty.Create(nameof(ShouldTrimValue), typeof(bool), typeof(TextValidationBehavior), false, propertyChanged: OnValidationPropertyChanged);
+
+        public static readonly BindableProperty ShouldConvertNullToEmptyValueProperty = BindableProperty.Create(nameof(ShouldConvertNullToEmptyValue), typeof(bool), typeof(TextValidationBehavior), true, propertyChanged: OnValidationPropertyChanged);
```
jfversluis, line 14: Not sure if we should set the default value to true. This will, by default, mutate the data of the user which can cause effects they aren't aware off. I guess we can leave it for now, just something to think about

XamarinCommunityToolkit/Behaviors/TextValidationBehavior.shared.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            OnRegexPropertyChanged();
+            base.OnAttachedTo(bindable);
+        }
+
+        protected override object DecorateValue()
+        {
+            var value = base.DecorateValue()?.ToString();
+
+            if (ShouldTrimValue)
```
jfversluis, line 72: Should we go for maximum flexibility and add the ability to only trim left/right or both? Can be in a follow up change

XamarinCommunityToolkit/Behaviors/ValidationBehavior.shared.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        public static readonly BindableProperty FlagsProperty = BindableProperty.Create(nameof(Flags), typeof(ValidationFlags), typeof(ValidationBehavior), ValidationFlags.ValidateOnUnfocusing | ValidationFlags.ForceMakeValidWhenFocused, propertyChanged: OnValidationPropertyChanged);
+
+        public static readonly BindableProperty ValueProperty = BindableProperty.Create(nameof(Value), typeof(object), typeof(ValidationBehavior), propertyChanged: OnValuePropertyChanged);
+
+        public static readonly BindableProperty ValuePropertyNameProperty = BindableProperty.Create(nameof(ValuePropertyName), typeof(string), typeof(ValidationBehavior), defaultValueCreator: GetDefaultValuePropertyName, propertyChanged: OnValuePropertyNamePropertyChanged);
+
+        public static readonly BindableProperty ForceValidateCommandProperty = BindableProperty.Create(nameof(ForceValidateCommand), typeof(ICommand), typeof(ValidationBehavior), defaultValueCreator: GetDefaultForceValidateCommand, defaultBindingMode: BindingMode.OneWayToSource);
```
jfversluis, line 22: I'll stop mentioning it from here, there might be more, but again some breaks :)

XamarinCommunityToolkitSample/Pages/Behaviors/EmailValidationBehaviorPage.xaml
```diff
@@ -0,0 +1,18 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
+<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             xmlns:behaviors="clr-namespace:XamarinCommunityToolkit.Behaviors;assembly=XamarinCommunityToolkit"
+             x:Class="XamarinCommunityToolkitSample.Pages.Behaviors.EmailValidationBehaviorPage">
+
+    <StackLayout Padding="{StaticResource ContentPadding}"
+                 Spacing="50"
+                 VerticalOptions="CenterAndExpand">
+        <Label Text="The text color will be changed to red in case of the wrong email is entered" />
```
jfversluis, line 10:
```suggestion
        <Label Text="Text color will change accordingly to the style that is configured when a invalid value (email address) is entered" />
```

XamarinCommunityToolkitSample/Pages/Behaviors/EmailValidationBehaviorPage.xaml
```diff
@@ -0,0 +1,18 @@
...
+        <Entry Placeholder="Email">
+            <Entry.Behaviors>
+                <behaviors:EmailValidationBehavior InvalidStyle="{StaticResource InvalidEntryStyle}"/>
```
jfversluis, line 13: I feel that it might be more clear for people, when they start looking at samples on how to use this, when the style is directly on this page? Instead of them also having to go look for where this style is defined.

IMO the samples should be as self-contained as possible

XamarinCommunityToolkitSample/Pages/Behaviors/EventToCommandBehaviorPage.xaml
```diff
@@ -0,0 +1,28 @@
...
+             xmlns:vm="clr-namespace:XamarinCommunityToolkitSample.ViewModels.Behaviors"
+             xmlns:behaviors="clr-namespace:XamarinCommunityToolkit.Behaviors;assembly=XamarinCommunityToolkit"
+             x:Class="XamarinCommunityToolkitSample.Pages.Behaviors.EventToCommandBehaviorPage">
+
+    <ContentPage.BindingContext>
+        <vm:EventToCommandBehaviorViewModel />
+    </ContentPage.BindingContext>
+
+    <ContentView Padding="{StaticResource ContentPadding}">
+        <StackLayout VerticalOptions="CenterAndExpand"
```
jfversluis, line 13: I'm missing the descriptive label on this page. What is the user looking at here?

XamarinCommunityToolkitSample/Pages/Behaviors/EventToCommandBehaviorPage.xaml
```diff
@@ -0,0 +1,28 @@
...
+                     Spacing="20">
+            <Label Text="{Binding ClickCount, StringFormat='{0} clicks'}" />
+            <Button Text="CLICK ME"
+                    TextColor="White"
+                    BackgroundColor="{StaticResource PrimaryColor}">
+                <Button.Behaviors>
+                    <behaviors:EventToCommandBehavior
+                        EventName="Clicked"
+                        Command="{Binding Command}" />
```
jfversluis, line 22: Maybe rename the actual command to something like `IncrementCommand` so it's clear that we are not referring to the `Command` property in this case

XamarinCommunityToolkitSample/Pages/Behaviors/NumericValidationBehaviorPage.xaml
```diff
@@ -0,0 +1,20 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
+<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             xmlns:behaviors="clr-namespace:XamarinCommunityToolkit.Behaviors;assembly=XamarinCommunityToolkit"
+             x:Class="XamarinCommunityToolkitSample.Pages.Behaviors.NumericValidationBehaviorPage">
+
+    <StackLayout Padding="{StaticResource ContentPadding}"
+                 Spacing="50"
+                 VerticalOptions="CenterAndExpand">
+        <Label Text="The text color will be changed to red in case of the wrong numeric format or if numeric is outside of [1.0; 100.0] boundaries." />
```
jfversluis, line 10:
```suggestion
        <Label Text="Text color will be changed according to the configured style in case of an invalid numeric format or when the numeric value is outside of the configured boundaries (min: 1.0, max: 100.0)." />
```

XamarinCommunityToolkitSample/ViewModels/AboutViewModel.cs
```diff
@@ -46,30 +49,29 @@ public string EmptyViewText
 
         public async Task OnAppearing()
         {
-            if (Contributors != null)
+            if (Contributors.Any())
```
jfversluis, line 52: I think this check is wrong now. Before we would return when `Contributors != null` now we return whenever the `Contributors` has `Any()`

XamarinCommunityToolkitSample/ViewModels/AboutViewModel.cs
```diff
@@ -46,30 +49,29 @@ public string EmptyViewText
...
             {
-                if(Contributors is null || !Contributors.Any())
-                    EmptyViewText = "No data loaded...";
+                // Suppress
             }
+
+            if (Contributors.Any())
+                return;
+
+            EmptyViewText = "No data loaded...";
```
jfversluis, line 74: Why don't we always set the `EmptyViewText`? The empty view is only shown whenever the collection is empty, right? We should not have to think about that logic. If we do, then we should consider if it's a bug in Forms

XamarinCommunityToolkitSample/ViewModels/WelcomeViewModel.cs
```diff
@@ -1,16 +1,15 @@
...
-            new WelcomeSectionModel(WelcomeSectionId.Behaviors),
-            new WelcomeSectionModel(WelcomeSectionId.Converters),
-            new WelcomeSectionModel(WelcomeSectionId.Extensions),
-            new WelcomeSectionModel(WelcomeSectionId.Views),
-            new WelcomeSectionModel(WelcomeSectionId.TestCases)
-        };
+        public IEnumerable<WelcomeSectionModel> Items { get; } =
+            ((WelcomeSectionId[])Enum
+            .GetValues(typeof(WelcomeSectionId)))
+            .Select(id => new WelcomeSectionModel(id));
```
jfversluis, line 13: I see this code mostly duplicated three times now. Can we fix that somehow? Also, it seems a bit complicated for what it does, but that might just be me

XamarinCommunityToolkit/Behaviors/BaseBehavior.shared.cs
AndreiMisiukevich: Yes, we cannot do it right now by normal approach cuz existing method **GetIsBound** is **internal** and even isn't what we need. That method checks property binding (if it exists), but we also need to check if this binding is ours or not.

So, I use this method for checking whether we need to set the default binding to some properties or not (BindingContext for every BaseBehavior & Value for ValidationBehavior)

XamarinCommunityToolkit/Behaviors/EventToCommandBehavior.shared.cs
AndreiMisiukevich: renamed

XamarinCommunityToolkit/Behaviors/EventToCommandBehavior.shared.cs
AndreiMisiukevich: Yes, it will work fine. CanExecute should always take some parameter even though it's null

We can unit test it for sure.

AndreiMisiukevich: > Yes, it will work fine. CanExecute should always take some parameter even though it's null
> 
> We can unit test it for sure, but it won't fail (can execute management is users responsibility. We can't manage it)

XamarinCommunityToolkitSample/ViewModels/AboutViewModel.cs
AndreiMisiukevich: 1) That's impossible to have 0 contributors on the GitHub project (only if not you've just created it without readme)
2) From my perspective, if for some reason we pulled 0 contributors on OnAppearing, we should try to pull it again on OnAppearing.

Also now Contributors always has value.

XamarinCommunityToolkitSample/ViewModels/AboutViewModel.cs
AndreiMisiukevich: ```string emptyViewText = "Loading data...";```

Initially, EmptyViewText is used for showing a different message. So here we replace it :)

XamarinCommunityToolkitSample/ViewModels/WelcomeViewModel.cs
AndreiMisiukevich: Unfortunately, that's impossible.

1) Generic constraints don't allow to specify constructors with parameters. We won't be able to pass id there if we'd like to extract this logic into separate class.

XamarinCommunityToolkit/Behaviors/TextValidationBehavior.shared.cs
AndreiMisiukevich: > Not sure if we should set the default value to true. This will, by default, mutate the data of the user which can cause effects they aren't aware off. I guess we can leave it for now, just something to think about

I believe that's expected behavior -- consider null string as empty. Just for me. Perpahs I am wrong :)

XamarinCommunityToolkit/Behaviors/TextValidationBehavior.shared.cs
AndreiMisiukevich: I've come up with an idea:
think we can introduce additional enum

```
[Flags]
public enum TextValidationFlags {
    None = 0,
    TrimStart = 1,
    TrimEnd = 2,
    Trim = 3,
    NullToEmpty = 4
}
```

Thoughts?

XamarinCommunityToolkitSample/Pages/Behaviors/EventToCommandBehaviorPage.xaml
AndreiMisiukevich: added

AndreiMisiukevich:
@jfversluis

jfversluis:
Don't worry. I didn't forget you yet. Please have a little patience

AndreiMisiukevich:
> Don't worry. I didn't forget you yet. Please have a little patience

It was a kind reminder. Sorry for disturb)

Issue #121, opened 2020-05-31
[Enhancement] Docs


AlexHedley:
## Summary

Is there a preference on how to document the toolkit?

Would it be ok to use [DocFX](https://dotnet.github.io/docfx/) and host on GitHub Pages?

We can automate the API documentation then have separate Articles written to explain how to use the various components. 

The following Actions can be used in a workflow:

- [GitHub Actions for GitHub Pages](https://github.com/peaceiris/actions-gh-pages) from [peaceiris](https://github.com/peaceiris) - Shohei Ueda

- [docfx-action](https://github.com/marketplace/actions/docfx-action) from [nikeee](https://github.com/nikeee) - Niklas Mollenhauer

Happy to add a PR with this functionality once there’s an initial code commit to work with.

SunnyMukherjee:
I have not used GitHub Pages, but once the authors settle on a path for documentation, I'll be happy to assist in the documentation effort with any approach.

jfversluis:
I think the Windows toolkit has their documentation included in the official Microsoft Docs. If that is a possibility for us as well I think that is the way to go. 

This will have the most discoverability and will have everything in one place.

sthewissen:
Agreed! Taking Xamarin.Essentials as an example (which does something similar I think using docfx) would also fit the project.

https://github.com/xamarin/Essentials/tree/develop/docs/en

jfversluis:
Sorry for the radio silence @AlexHedley! We have a separate repo and actual Docs URL for our documentation and are working on putting something on there and more importantly: have some kind of guide on how to contribute to that.

To be continued...

AlexHedley:
No problem at all, I'll keep an eye out for this and look forward to contributing as soon as possible :)

Shall we just close the PR and use this Issue to track things?

jfversluis:
Keep it open for the time being. You will still have the content as a reference of course, but it's a good indicator for us to remember we need to get an outline like that on the Docs. When we port everything over there we can close it.

Thanks for your support and patience!

AlexHedley:
What do I see here :)

https://github.com/MicrosoftDocs/xamarin-communitytoolkit

jfversluis:
There it is! I think it's safe to start making some changes there, so feel free to port some things over

jfversluis:
OK I think the Docs are well underway and will be an ongoing thing so we don't need this issue to nag us :D

PR #135, opened 2020-06-14
Initial Docs Setup

AlexHedley:
### Description of Change ###

This PR brings in a GitHub Pages hosted documentation site using [DocFX](https://dotnet.github.io/docfx/).

#121 has a discussion about possible solutions, this is just a temporary measure until the docs can be added to the official MS site like [Essentials](https://github.com/xamarin/Essentials/tree/develop/docs/en) but being that this is using the same technology it could mean minimal effort to transfer over but allows for proper documentation to be written whilst there is on-going development.

The workflow `yml` shows the Actions being used.

One manual step to make, if merged, is you need to toggle: 
Settings | **GitHub Pages** | Source to *master branch* then to *gh-pages branch* for the first build to trigger then subsequent pushes work fine.

You can see an example on my fork [here](https://alexhedley.github.io/XamarinCommunityToolkit/).


jfversluis:
Awesome work @AlexHedley ! Thank you so much :D 

I'm not entirely familiar with GH Pages, and I think you mention it, but the branch switch is just once right?

AlexHedley:
@jfversluis yes, just the once or it doesn’t trigger the initial release of the website unfortunately, but then new pushes will update fine

SunnyMukherjee:
Hi @AlexHedley, I want to generate the docs for the Behaviors PR #126 so it is available whenever it is merged.  Can I follow the instructions on the DocFX site to generate the docs?  Or is there a better tutorial somewhere else?  Thanks.

AlexHedley:
Hi @SunnyMukherjee 

The DocFX site is what I used to produce this,

I guess it depends, if this PR is completed first you could then pull that into yours.

If you create a new docs site in your branch you’d need to copy over this one to stop MCs and get all my config

You could add just a single markdown file with some Doco in it then when both are merged a maintainer could move it over, 

I’m not sure the best way to handle multiple forks atm

jfversluis:
I would hold of on doing too much work here right now. As much as I appreciate the effort, we probably get more news about hosting our documentation on the Docs pages this week. Let’s wait for that a little to decide the right course of action and see if we can still use something from this PR

SunnyMukherjee:
> Hi @SunnyMukherjee
> 
> The DocFX site is what I used to produce this,
> 
> I guess it depends, if this PR is completed first you could then pull that into yours.
> 
> If you create a new docs site in your branch you’d need to copy over this one to stop MCs and get all my config
> 
> You could add just a single markdown file with some Doco in it then when both are merged a maintainer could move it over,
> 
> I’m not sure the best way to handle multiple forks atm

I read through their site.  I thought it was a matter of adding their NuGet package and it would generate the documentation in a docs folder using the comments in our classes.  It sounds like what you have done in your PR and you are describing is different than what I understand.  I may have used DocFX or something similar a long time ago, but I guess I need to understand your approach to be on the same page.

AlexHedley:
@SunnyMukherjee there’s two parts to the documentation.

One is the API side which is auto generated, just add good xml docs to your classes etc.

The other side it the hand written Articles which you write in Markdown i.e. example of how to use your behaviours with images, notes etc

jfversluis:
I think all this moved to the Docs repo now, so we can close this here

Issue #129, opened 2020-06-08
[Enhancement] Add sample app


jfversluis:
## Summary

We need to add a (good looking!) sample app that shows all the stuff that is in the library. Additionally, add in the PR template and CONTRIBUTING.md that people are required to add a sample page for their new functionality as well as tests.

AndreiMisiukevich:
@jfversluis can be closed, right?

jfversluis:
We have a sample app which should just be evolving together with our codebase, so I think this can be closed for now

PR #143, opened 2020-06-18
Sample app redesign [Part 1]

AndreiMisiukevich:
### Description of Change ###

- Redesigned sample app;
- Cleaned and refactored code;
- Made XF version consistent across all solution projects;
- Added a new "folder" - **Extensions** with single MarkupExtension that helps to work with local images.

**Note**: this is the first step of the sample app redesign.
Once this PR is merged, I will submit a follow-up PR with updates within #132 

### Bugs Fixed ###
- Related to issue #129 

### API Changes ###

Added: 
- `ImageResourceExtension`

### Behavioral Changes ###
Jun-18-2020 03-49-01

### PR Checklist ###

- [ ] Updated documentation


XamarinCommunityToolkitSample/App.xaml
```diff
@@ -4,8 +4,47 @@
...
+        </Style>
+
+        <Style TargetType="ContentPage" ApplyToDerivedTypes="True">
+            <Setter Property="BackgroundColor" Value="{StaticResource AppBackgroundColor}" />
+            <Setter Property="ios:Page.UseSafeArea" Value="True" />
+        </Style>
+
+        <Style TargetType="Button">
+            <Setter Property="FontSize" Value="Medium" />
+            <Setter Property="BackgroundColor" Value="#1976D2" />
```
Depechie, line 33: Maybe also add this as static resource Color.

XamarinCommunityToolkitSample/Views/WelcomePage.xaml.cs
```diff
@@ -0,0 +1,48 @@
...
+            => id switch
+            {
+                Section.SectionId.Behaviors => new BehaviorsPage(),
+                Section.SectionId.Converters => new ContentPage { Title = "Not Implemented yet" },
+                Section.SectionId.Views => new ContentPage { Title = "Not Implemented yet" },
+                Section.SectionId.TestCases => new ContentPage { Title = "Not Implemented yet" },
+                _ => throw new System.NotImplementedException()
+            };
+
+        public sealed class Section
```
Depechie, line 34: Classes best in a separate file.

XamarinCommunityToolkitSample/ViewModels/BaseViewModel.cs
```diff
@@ -1,56 +1,24 @@
...
-
-            changed.Invoke(this, new PropertyChangedEventArgs(propertyName));
-        }
-
-        #endregion
+        protected virtual void OnPropertyChanged([CallerMemberName] string name = null)
+            => PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(name));
     }
-}
+}
```
Depechie, line 24: I think the BaseViewModel is no longer used? Use it, or drop it I guess.

XamarinCommunityToolkitSample/Views/WelcomePage.xaml
```diff
@@ -0,0 +1,38 @@
...
+    </NavigationPage.TitleView>
+
+    <StackLayout>
+        <Image Source="{extensions:ImageResource xamarin_monkey.png}" HeightRequest="250" />
+        <ScrollView Padding="{StaticResource ContentPadding}"
+                    VerticalOptions="FillAndExpand">
+            <StackLayout Spacing="40">
+                <Label Text="Welcome to XamarinCommunityToolkit" FontSize="Large" />
+                <StackLayout Spacing="10"
+                             BindableLayout.ItemsSource="{Binding Sections, Source={x:Reference Page}}">
```
Depechie, line 22: See remark about BaseViewModel...

XamarinCommunityToolkitSample/Views/WelcomePage.xaml.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        ICommand navigateCommand;
+
+        public WelcomePage()
+            => InitializeComponent();
+
+        public ICommand NavigateCommand => navigateCommand ??= new Command(parameter
+            => Navigation.PushAsync(GetPage((Section.SectionId)parameter)));
+
+        public Section[] Sections { get; } =
```
Depechie, line 16: See remark about BaseViewModel

XamarinCommunityToolkitSample/Views/WelcomePage.xaml.cs
```diff
@@ -0,0 +1,48 @@
...
+                Section.SectionId.TestCases => new ContentPage { Title = "Not Implemented yet" },
+                _ => throw new System.NotImplementedException()
+            };
+
+        public sealed class Section
+        {
+            public SectionId Id { get; set; }
+            public string Text { get; set; }
+
+            public enum SectionId
```
Depechie, line 39: I would add enums in a separate Enums.cs file in Models folder.

AndreiMisiukevich:
@Depechie the sample is updated

XamarinCommunityToolkitSample/Models/WelcomeSectionId.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+
+namespace XamarinCommunityToolkitSample.Models
+{
+    public enum WelcomeSectionId
```
Depechie, line 5: I would still extract this to an Enums.cs file, not sure what others think of this though... @sthewissen  ?

XamarinCommunityToolkitSample/Models/WelcomeSectionId.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        Behaviors,
+        Converters,
+        Extensions,
+        Views,
+        TestCases
+    }
+
+    public static class WelcomeSectionIdExtensions
+    {
```
Depechie, line 15: I would rename the file to include Extension

AndreiMisiukevich: Do you mean extract this class to the separate file?

Depechie: Yeah indeed

XamarinCommunityToolkitSample/Models/WelcomeSectionId.cs
AndreiMisiukevich: My opinion is that we should keep every class/struct/enum in the separate files.
But I'd like to make an exception for Enum extensions cuz that's the only option to add some methods there. (to comment below)

Depechie: Yeah I get it... still even in these cases I separated them. But like I said, not a big problem in this case, so if no other reviewer objects, current code can be kept.

XamarinCommunityToolkitSample/Models/WelcomeSectionId.cs
AndreiMisiukevich: Are you Ok with that as well?

That's like extension in swift :) or partial class in C#

XamarinCommunityToolkitSample/Models/WelcomeSectionId.cs
sthewissen: I'm fine with things being their own class.

PR #139, opened 2020-06-17
Implement IMarkupExtension in converters

AndreiMisiukevich:
### Description of Change ###

Implemented IMarkupExtension for converters.

Now they can be used as:
```xaml
 <Label Text="{Binding MyString, Converter={kit:TextCaseConverter Upper}}" />
```

Since default interface implementations are not supported by nestandard 1 and 2 we **cannot** use them. So I decided to go ahead with the abstract class.
Also, I had to add *System.ComponentModel* nuget for nestandard1.0 **only** due to nestandard1.0 needs (IServiceProvider is presented there)

### Bugs Fixed ###

- Related to issue #133 

### API Changes ###

None

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [ ] Has samples (if omitted, state reason in description)
- [X] Rebased on top of master at time of PR
- [X] Changes adhere to coding standard
- [ ] Updated documentation

PR #144, opened 2020-06-18
Multiconverter

sthewissen:
### Description of Change ###

This is a converter that allows for multiple converters to be applied in series.

#### Sample

```
<ResourceDictionary>
    <xct:MultiConverter x:Key="myMultiConverter">
        <xct:TextCaseConverter />
        <xct:NotEqualConverter />
    </xct:MultiConverter>
    <x:Array x:Key="multiParams" Type="{x:Type xct:MultiConverterParameter}">
        <xct:MultiConverterParameter ConverterType="{x:Type xct:TextCaseConverter}" Value="{x:Static xct:TextCaseConverterType.Upper}" />
        <xct:MultiConverterParameter ConverterType="{x:Type xct:NotEqualConverter}" Value="ANDREI" />
    </x:Array>
</ResourceDictionary>
```

In your control e.g.:

 ```
<Label Text="{Binding X, Converter={StaticResource myMultiConverter}, ConverterParameter={StaticResource multiParams}}"
                VerticalOptions="CenterAndExpand" 
                HorizontalOptions="CenterAndExpand" />
```

With a binding context of:

```
BindingContext = new JustATest { X = "Andrei" };
```

### Bugs Fixed ###

- Related to issue #137 

### API Changes ###

Added: 
 
- `MultiConverter`

### PR Checklist ###

- [ ] Updated documentation

### DIFF
--- a/XamarinCommunityToolkit/Converters/MultiConverter.shared.cs
+++ b/XamarinCommunityToolkit/Converters/MultiConverter.shared.cs
@@ -1,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Xamarin.Forms;
+
+namespace XamarinCommunityToolkit.Converters
+{
+    /// <summary>
+    /// Converts an incoming value using all of the incoming converters in sequence.
+    /// </summary>
+    public class MultiConverter : List<IValueConverter>, IValueConverter
+    {
+        /// <summary>
+        /// Uses the incoming converters to convert the value.
+        /// </summary>
+        /// <param name="value">Value to convert.</param>
+        /// <param name="targetType">The type of the binding target property.</param>
+        /// <param name="parameter">Parameter to pass into subsequent converters.</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>The converted value.</returns>
+        public object Convert(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
+            => this.Aggregate(value, (current, converter) => converter.Convert(current, targetType, parameter, culture));
+
+        public object ConvertBack(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
+            => throw new NotImplementedException();
+    }
+}
--- a/XamarinCommunityToolkit/Converters/MultiConverter.shared.cs
+++ b/XamarinCommunityToolkit/Converters/MultiConverter.shared.cs
@@ -19,9 +19,28 @@ namespace XamarinCommunityToolkit.Converters
         /// <param name="culture">The culture to use in the converter.</param>
         /// <returns>The converted value.</returns>
         public object Convert(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
-            => this.Aggregate(value, (current, converter) => converter.Convert(current, targetType, parameter, culture));
+        {
+            // If it's a list of MultiConverterParameter, use those and propagate them into the converters.
+            if (parameter is IList<MultiConverterParameter> parameters)
+            {
+                return this.Aggregate(value, (current, converter) => converter.Convert(current, targetType,
+                    parameters.FirstOrDefault(x => x.ConverterType == converter.GetType())?.Value, culture));
+            }
+
+            return this.Aggregate(value, (current, converter) => converter.Convert(current, targetType, parameter, culture));
+        }
 
         public object ConvertBack(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
             => throw new NotImplementedException();
+    }
+
+    /// <summary>
+    /// Represents a parameter to be used in the MultiConverter.
+    /// </summary>
+    public class MultiConverterParameter : BindableObject
+    {
+        public Type ConverterType { get; set; }
+
+        public object Value { get; set; }
     }
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: XamarinCommunityToolkit/Converters/MultiConverter.shared.cs [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+using Xamarin.Forms;
+
+namespace XamarinCommunityToolkit.Converters
+{
+    /// <summary>
+    /// Converts an incoming value using all of the incoming converters in sequence.
+    /// </summary>
+    public class MultiConverter : List<IValueConverter>, IValueConverter
+    {
+        /// <summary>
```
AndreiMisiukevich, line 13: We should merge https://github.com/xamarin/XamarinCommunityToolkit/pull/139 and then add a property "Parameters" as well

XamarinCommunityToolkit/Converters/MultiConverter.shared.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <param name="targetType">The type of the binding target property.</param>
+        /// <param name="parameter">Parameter to pass into subsequent converters.</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>The converted value.</returns>
+        public object Convert(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
+        {
+            // If it's a list of MultiConverterParameter, use those and propagate them into the converters.
+            if (parameter is IList<MultiConverterParameter> parameters)
+            {
+                return this.Aggregate(value, (current, converter) => converter.Convert(current, targetType,
```
AndreiMisiukevich, line 26: **this** can be omitted )

XamarinCommunityToolkit/Converters/MultiConverter.shared.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <summary>
+        /// Uses the incoming converters to convert the value.
+        /// </summary>
+        /// <param name="value">Value to convert.</param>
+        /// <param name="targetType">The type of the binding target property.</param>
+        /// <param name="parameter">Parameter to pass into subsequent converters.</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>The converted value.</returns>
+        public object Convert(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
+        {
```
AndreiMisiukevich, line 22: I believe we can make this method in line

```csharp
=> parameter is IList<MultiConverterParameter> parameters
    ? // First option
    : // Second one
```

XamarinCommunityToolkit/Converters/MultiConverter.shared.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
+            => throw new NotImplementedException();
+    }
+
+    /// <summary>
+    /// Represents a parameter to be used in the MultiConverter.
+    /// </summary>
+    public class MultiConverterParameter : BindableObject
```
AndreiMisiukevich, line 40: Perhaps it would be a good practice to store every class in a separate file (even if it's a small guy)

sthewissen:
> I was thinking if it's possible to make parameter as ResourceDictionary and handle parameters via key mappings by attached properties but it seems no. (Converters are not bindable objects).
>
> Probably your solution is a single proper one

The `MultiConverterParameter` class also still inherits from `BindableObject` as I was trying to have `Value` be a bindable property for maximum flexibility. I don't think that will work though, or at least I couldn't get it to work. Do you perhaps have any idea why?

AndreiMisiukevich:
> > I was thinking if it's possible to make parameter as ResourceDictionary and handle parameters via key mappings by attached properties but it seems no. (Converters are not bindable objects).
> > Probably your solution is a single proper one
> 
> The `MultiConverterParameter` class also still inherits from `BindableObject` as I was trying to have `Value` be a bindable property for maximum flexibility. I don't think that will work though, or at least I couldn't get it to work. Do you perhaps have any idea why?

They don't receive BindingContext, I guess... Not sure we can handle it like with Behaviors...

XamarinCommunityToolkit/Converters/MultiConverterParameter.shared.cs
```diff
@@ -0,0 +1,15 @@
...
+using Xamarin.Forms;
+
+namespace XamarinCommunityToolkit.Converters
+{
+    /// <summary>
+    /// Represents a parameter to be used in the MultiConverter.
+    /// </summary>
+    public class MultiConverterParameter : BindableObject
+    {
+        public Type ConverterType { get; set; }
```
AndreiMisiukevich, line 11: probably we should static BindableProperty ConverterTypeProperty & ValueProperty if we'd like to try to make them bindable

AndreiMisiukevich: But I am not sure we can make them bindable in deed :(

AndreiMisiukevich:
@sthewissen now you may inherit the MultyConverter from ValueConverterExtension

sthewissen:
> @sthewissen now you may inherit the MultyConverter from ValueConverterExtension

I'm wondering though since I inherit from `List<IValueConverter>` what the best way to do so is. We would need a different `ValueConverterExtension` for it that implements `IMarkupExtension<List<IValueConverter>>`?

Issue #145, opened 2020-06-20
[Enhancement] SafeAreaInsets effect


sthewissen:
## Summary
On iOS we can set the `SafeAreaInsets` usage on a page level. This means that either you get additional padding on the top/bottom of your layout on devices with a notch or no padding on devices without a notch. 

Sometimes, however, you want to use the full screen but that also means that individual elements can end up going behind the notch. We want to be able to use `SafeAreaInsets` on individual elements. This effect fixes that and is applicable to any element, adding the padding from `SafeAreaInsets` to individual elements on the page.

## API Changes
In order to facilitate this, an effect is added. The effect itself is only implemented on iOS.

sthewissen:
I already have this code in a few of my projects and will investigate PRing it in.

PR #146, opened 2020-06-20
SafeAreaEffect helper for iOS specific styling

sthewissen:
### Description of Change ###

Added the SafeArea effect. This is used to be able to add safe area padding onto each element on the page instead of just on the page level.

### Bugs Fixed ###

- Related to issue #145 

### API Changes ###

Added: 
 
- `SafeAreaPaddingEffect`
- `SafeAreaPaddingEffectRouter`
- `SafeAreaPaddingEffectRouter` implementation for iOS
- `SafeAreaPadding` struct

### PR Checklist ###

- [ ] Updated documentation


sthewissen:
There's a few things left here, hence the draft status.

- Probably need a TypeConverter to handle this nicely on the XAML side.
- Need to whip up a bit of a sample + scenarios for testing.

XamarinCommunityToolkit/Effects/SafeAreaPaddingEffect.shared.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            if (!(bindable is View view))
+                return;
+
+            var hasSafeArea = !((SafeAreaPadding)newValue).IsEmpty;
+
+            if (hasSafeArea)
+            {
+                view.Effects.Add(new SafeAreaPaddingEffectRouter());
+            }
```
AndreiMisiukevich, line 30: Can we add **return** and remove **else** ?

XamarinCommunityToolkit/Effects/SafeAreaPaddingEffect.shared.cs
```diff
@@ -0,0 +1,42 @@
...
+            else
+            {
+                var toRemove = view.Effects.FirstOrDefault(e => e is SafeAreaPaddingEffectRouter);
```
AndreiMisiukevich, line 33: Probably it would be slightly better to keep a class field and store effect there

XamarinCommunityToolkit/Effects/SafeAreaPaddingEffectRouter.iOS.cs
```diff
@@ -0,0 +1,53 @@
...
+                        initialPadding.Top + (safeArea.Top ? insets.Top : 0),
+                        initialPadding.Right + (safeArea.Right ? insets.Right : 0),
+                        initialPadding.Bottom + (safeArea.Bottom ? insets.Bottom : 0)
+                    );
+
+                    return;
+                }
+            }
+
+            element.Padding = initialPadding;
```
AndreiMisiukevich, line 42: if the device has iOS 10 for example this code will set "Zero" padding (cuz initial padding never set)

XamarinCommunityToolkit/Effects/SafeAreaPaddingEffectRouter.iOS.cs
```diff
@@ -0,0 +1,53 @@
...
+    public class SafeAreaPaddingEffectRouter : PlatformEffect
+    {
+        Thickness initialPadding;
+
+        protected override void OnAttached()
+        {
+            if (!(Element is Layout element))
+                return;
+
+            if (UIDevice.CurrentDevice.CheckSystemVersion(11, 0))
```
AndreiMisiukevich, line 21: too much nesting.

could you please refactor to **guard** pattern?

```
if(!UIDevice.CurrentDevice.CheckSystemVersion(11, 0))
    return;

initialPadding = element.Padding;

if( insets.Top <= 0)
    return;

...
```

XamarinCommunityToolkit/Effects/SafeAreaPaddingEffectRouter.iOS.cs
```diff
@@ -0,0 +1,53 @@
...
+                }
+            }
+
+            element.Padding = initialPadding;
+        }
+
+        protected override void OnDetached()
+        {
+            if (Element is Layout element)
+            {
```
AndreiMisiukevich, line 48: braces can be omitted

XamarinCommunityToolkit/Helpers/SafeAreaPadding.shared.cs
```diff
@@ -0,0 +1,82 @@
...
+        bool Equals(SafeAreaPadding other)
+        {
+            if (!isParameterized && !other.isParameterized)
+                return true;
+
+            return Left == other.Left &&
+                Top == other.Top &&
+                Right == other.Right &&
+                Bottom == other.Bottom;
+        }
```
AndreiMisiukevich, line 46: ?
```suggestion
        => !isParameterized && 
            !other.isParameterized || 
            Left == other.Left &&
            Top == other.Top &&
            Right == other.Right &&
            Bottom == other.Bottom;
        }
```

XamarinCommunityToolkit/Helpers/SafeAreaPadding.shared.cs
```diff
@@ -0,0 +1,82 @@
...
+            Left = left;
+            Top = top;
+            Right = right;
+            Bottom = bottom;
+        }
+
+        public static implicit operator SafeAreaPadding(bool uniformSafeAreaPadding)
+            => new SafeAreaPadding(uniformSafeAreaPadding);
+
+        bool Equals(SafeAreaPadding other)
```
AndreiMisiukevich, line 37: probably we should implement **IEquatable** instead private method approach.

XamarinCommunityToolkit/Helpers/SafeAreaPadding.shared.cs
```diff
@@ -0,0 +1,82 @@
...
+                Bottom == other.Bottom;
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj))
+                return false;
+
+            return obj is SafeAreaPadding safeAreaPadding && Equals(safeAreaPadding);
+        }
```
AndreiMisiukevich, line 54:
```suggestion
        => !ReferenceEquals(null, obj) &&
            obj is SafeAreaPadding safeAreaPadding && 
            Equals(safeAreaPadding);
```

XamarinCommunityToolkit/Effects/SafeAreaPaddingEffect.shared.cs
```diff
@@ -0,0 +1,39 @@
...
+            if (hasSafeArea)
+            {
+                view.Effects.Add(new SafeAreaPaddingEffectRouter());
+                return;
+            }
+
+            var toRemove = view.Effects.FirstOrDefault(e => e is SafeAreaPaddingEffectRouter);
+
+            if (toRemove != null)
+                view.Effects.Remove(toRemove);
```
AndreiMisiukevich, line 36: We should always remove the effect or check if any effect already exists (cuz possibly we can add several effects now)

```suggestion
            var oldEffect = view.Effects.FirstOrDefault(e => e is SafeAreaPaddingEffectRouter);
            if (oldEffect != null)
                view.Effects.Remove(oldEffect);
                
            if (((SafeAreaPadding)newValue).IsEmpty)
                return;
                
            view.Effects.Add(new SafeAreaPaddingEffectRouter());
```

XamarinCommunityToolkit/Effects/SafeAreaPaddingEffect.shared.cs
AndreiMisiukevich: Forget about it. That's a static class... my mistake.

XamarinCommunityToolkit/Helpers/SafeAreaTypeConverter.shared.cs
```diff
@@ -0,0 +1,47 @@
...
+                }
+                else
+                {
+                    // Single uniform SafeArea
+                    if (bool.TryParse(value, out var l))
+                        return new SafeArea(l);
+                }
+            }
+
+            throw new InvalidOperationException($"Cannot convert \"{value}\" into {typeof(SafeArea)}");
```
AndreiMisiukevich, line 44: I believe it can be more laconic :)
```suggestion
            try
            {
                var safeAreaItems = value?.Trim().Split(',').Select(bool.Parse);
                return safeAreaItems.Length switch
                {
                    1 => new SafeArea(l),
                    2 => new SafeArea(h, v),
                    4 => new SafeArea(l, t, r, b),
                    _ => throw new Exception()
                };
            }
            catch
            {
                throw new InvalidOperationException($"Cannot convert \"{value}\" into {typeof(SafeArea)}");
            }
```

sthewissen: The problem I have is that sometimes shorter is not necessarily more readable 😂

sthewissen: Also, in your suggestion, where are all the variables coming from like `l`, `h`, `v`, `t`, `r` and `b`? I would suppose the LINQ query select creates a dynamic list, but I'd have to grab them by index? Not sure if I like that more.

AndreiMisiukevich: > Also, in your suggestion, where are all the variables coming from like `l`, `h`, `v`, `t`, `r` and `b`? I would suppose the LINQ query select creates a dynamic list, but I'd have to grab them by index? Not sure if I like that more.

yes, I meant getting by the index of course )

AndreiMisiukevich: Then

```csharp
public override object ConvertFromInvariantString(string value)
{
    var items = ParseArray(value);
    return items.Length switch
    {
        1 => new SafeArea(items[0]),
        2 => new SafeArea(items[0], items[1]),
        4 => new SafeArea(items[0], items[1], items[2], items[3]),
        _ => throw new InvalidOperationException($"Cannot convert \"{value}\" into {typeof(SafeArea)}")
    };
}

bool[] ParseArray(string value)
{
    var source = value?.Trim().Split(',');
    var items = new bool[source?.Length ?? 0];
    for (var i = 0; i < items.Length; i++)
    {
        if (!bool.TryParse(source[i], out var item))
            return null;

        items[i] = item;
    }
    return items;
}
```

XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs
```diff
@@ -0,0 +1,48 @@
...
+    public class SafeAreaEffectRouter : PlatformEffect
+    {
+        Thickness initialMargin;
+
+        protected override void OnAttached()
+        {
+            if (!(Element is View element))
+                return;
+
+            if (!UIDevice.CurrentDevice.CheckSystemVersion(11, 0) || !UIApplication.SharedApplication.Windows.Any())
```
jfversluis, line 22: Since we're using Forms anyway. Isn't there a Forms.IsiOS11OrNewer or something? Or is that something internal again...

XamarinCommunityToolkit/Helpers/SafeArea.shared.cs
```diff
@@ -0,0 +1,79 @@
+﻿using System;
+using Xamarin.Forms;
+
+namespace XamarinCommunityToolkit.Helpers
+{
+    [TypeConverter(typeof(SafeAreaTypeConverter))]
+    public struct SafeArea
+    {
+        readonly bool isParameterized;
```
jfversluis, line 9: Should this be `_isParameterized`?

XamarinCommunityToolkit/Helpers/SafeAreaTypeConverter.shared.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using Xamarin.Forms;
+
+namespace XamarinCommunityToolkit.Helpers
+{
+    [Xamarin.Forms.Xaml.TypeConversion(typeof(SafeArea))]
```
jfversluis, line 6: Move `Xamarin.Forms.Xaml` to using

XamarinCommunityToolkit/Helpers/SafeAreaTypeConverter.shared.cs
jfversluis: I think I agree with Steven for this. Yes, the solution is technically accurate and will work fine, but I think what's in there now is more readable for the time being. Let's just start out simple and engineer whenever we have a need for it not because we might need to have to in the future.

XamarinCommunityToolkit/Helpers/SafeArea.shared.cs
AndreiMisiukevich: > Should this be `_isParameterized`?

No, it shouldn't according to our code style :)

XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs
sthewissen:
```
internal static bool IsiOS11OrNewer
{
	get
	{
		if (!s_isiOS11OrNewer.HasValue)
			s_isiOS11OrNewer = UIDevice.CurrentDevice.CheckSystemVersion(11, 0);
		return s_isiOS11OrNewer.Value;
	}
}
```

Not quite sure why though 🤷

sthewissen:
LGTM!

Issue #149, opened 2020-06-29
[Enhancement] TextValidationFlags


AndreiMisiukevich:
## Summary
We should update existing **decoration** approach for TextValidationBehavior.
I propose to replace several bindable properties by a single one (flags enum).

## API Changes
```csharp

public TextValidationFlags Flags { get; set; }

[Flags]
public enum TextValidationFlags {
    None = 0,
    TrimStart = 1,
    TrimEnd = 2,
    Trim = 3,
    NullToEmpty = 4
}
```

## Intended Use Case
It will make text-decoration more clear for users (You have only one property for that, no need to look for others).

PR #151, opened 2020-07-03
TextDecorationFlags for TextValidationBehavior

AndreiMisiukevich:
### Description of Change ###

Replaces a bunch of properties by one "flags" property.

### Ticket Fixed ###
- Related to issue #149 

### API Changes ###

Added: 
 
- `TextDecorationFlags TextValidationBehavior.DecorationFlags { get; set; }`
- `TextDecorationFlags`

Removed:

 - `bool TextValidationBehavior.ShouldTrimValue`
 - `bool TextValidationBehavior.ShouldConvertNullToEmptyValue`
 
```csharp
    [Flags]
    public enum TextDecorationFlags
    {
        None = 0,
        TrimStart = 1,
        TrimEnd = 2,
        Trim = 3,
        NullToEmpty = 4,
        ReduceWhiteSpaces = 8
    }
```

### Behavioral Changes ###

More convenient for use (need to set up only one property with necessary flags instead of several)

### PR Checklist ###

- [ ] Updated documentation

### DIFF
--- a/XamarinCommunityToolkit/Behaviors/TextValidationBehavior.shared.cs
+++ b/XamarinCommunityToolkit/Behaviors/TextValidationBehavior.shared.cs
@@ -1,5 +1,6 @@
 ﻿using Xamarin.Forms;
 using System.Text.RegularExpressions;
+using System.Text;
 
 namespace XamarinCommunityToolkit.Behaviors
 {
@@ -11,11 +12,8 @@ namespace XamarinCommunityToolkit.Behaviors
         public static readonly BindableProperty MaximumLengthProperty =
             BindableProperty.Create(nameof(MaximumLength), typeof(int), typeof(TextValidationBehavior), int.MaxValue, propertyChanged: OnValidationPropertyChanged);
 
-        public static readonly BindableProperty ShouldTrimValueProperty =
-            BindableProperty.Create(nameof(ShouldTrimValue), typeof(bool), typeof(TextValidationBehavior), false, propertyChanged: OnValidationPropertyChanged);
-
-        public static readonly BindableProperty ShouldConvertNullToEmptyValueProperty =
-            BindableProperty.Create(nameof(ShouldConvertNullToEmptyValue), typeof(bool), typeof(TextValidationBehavior), true, propertyChanged: OnValidationPropertyChanged);
+        public static readonly BindableProperty DecorationFlagsProperty =
+            BindableProperty.Create(nameof(DecorationFlags), typeof(TextDecorationFlags), typeof(TextValidationBehavior), TextDecorationFlags.None, propertyChanged: OnValidationPropertyChanged);
 
         public static readonly BindableProperty RegexPatternProperty =
             BindableProperty.Create(nameof(RegexPattern), typeof(string), typeof(TextValidationBehavior), defaultValueCreator: GetDefaultRegexPattern, propertyChanged: OnRegexPropertyChanged);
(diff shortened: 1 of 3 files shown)


XamarinCommunityToolkit/Behaviors/TextValidationBehavior.shared.cs
```diff
@@ -112,5 +114,21 @@ void OnRegexPropertyChanged()
             => regex = RegexPattern != null
                 ? new Regex(RegexPattern, RegexOptions)
                 : null;
+
+        string ReduceWhiteSpaces(string value)
```
jfversluis, line 118: What does this do? Remove all whitespaces? Why not simply say `string.Replace(' ', '');`?

jfversluis: OK, so yeah, `IsWhitespace` does more than just spaces. See [here](https://docs.microsoft.com/en-us/dotnet/api/system.char.iswhitespace?view=netcore-3.1#remarks). But looks like we can simply call [`Trim`](https://docs.microsoft.com/en-us/dotnet/api/system.string.trim?view=netcore-3.1) (see under Note to Callers)

jfversluis: Forget I said anything. Trim is just begin and end of course

AndreiMisiukevich: > OK, so yeah, `IsWhitespace` does more than just spaces. See [here](https://docs.microsoft.com/en-us/dotnet/api/system.char.iswhitespace?view=netcore-3.1#remarks). But looks like we can simply call [`Trim`](https://docs.microsoft.com/en-us/dotnet/api/system.string.trim?view=netcore-3.1) (see under Note to Callers)

this method reduces spaces quantity to 1
Input: ```"____Hi there!_________Who are you?______"```
Output ```"_Hi there!_Who are you?_"```

``string.Split``` + ```string.Join``` won't work for us cuz this will remove spaces from start and end of the string.

jfversluis: Will be good and very important to have some tests for this though. I also see some cases for removing spaces altogether or maybe replace it with another character.

AndreiMisiukevich: > Will be good and very important to have some tests for this though. I also see some cases for removing spaces altogether or maybe replace it with another character.

Hmm, could you please provide an example? I thought about adding such flag but didn't realize where it can be used.
Now if you want to remove all spaces from start and end of the string you can use the "Trim" option (As I use in email test-case).

```"____myemail@gmail.com____"``` will be updated to ```"myemail@gmail.com"```

But when do we need to remove all spaces?

jfversluis: I mean, people have weird systems, right? And even if it's only for design purposes. Maybe you're showing barcode data or a phone number and you (or your product owner... ;)) wants to show it as one number but it is stored as `12 34 56 789`.

Or maybe you are showing a MAC address and you want to replace all spaces with colons or the other way around. But maybe we should wait for someone to actually have a need for these cases.

AndreiMisiukevich: > I mean, people have weird systems, right? And even if it's only for design purposes. Maybe you're showing barcode data or a phone number and you (or your product owner... ;)) wants to show it as one number but it is stored as `12 34 56 789`.
> 
> Or maybe you are showing a MAC address and you want to replace all spaces with colons or the other way around. But maybe we should wait for someone to actually have a need for these cases.

Got it, thanks.
Looks like we should expose one more flag "RemoveAllNonDigits"? :)

PR #156, opened 2020-07-07
Added Missing Converter Tests

Char0394:
# Description of Change
Added missing converter tests.

Added:
- IntToBoolConverter_Tests
- InvertedBoolConverter_Tests
- IsNotNullOrEmptyConverter_Tests
- ListIsNotNullOrEmptyConverter_Tests
- ListIsNullOrEmptyConverter_Tests
- MultiConverter_Tests
- NotEqualConverter_Tests
- TextCaseConverter_Tests

NOTE: With this, we have all the converters tests ready, we need to identify what else we want to test.


XamarinCommunityToolkit.UnitTests/Converters/ListIsNullOrEmptyConverter_Tests.cs
```diff
@@ -1,10 +1,43 @@
...
+using System.Linq;
+using XamarinCommunityToolkit.Converters;
+using Xunit;
+
 namespace XamarinCommunityToolkit.UnitTests.Converters
 {
     public class ListIsNullOrEmptyConverter_Tests
     {
-        public ListIsNullOrEmptyConverter_Tests()
+        public static IEnumerable<object[]> GetData()
```
jfversluis, line 12: Not that it really matters, but is this required to be in a method? Can't it just be a collection that is declared as a variable?

jfversluis:
Great work @Char0394 ! Thank you so much!

> NOTE: With this, we have all the converters tests ready, we need to identify what else we want to test.

I think we already have some other things in our codebase already. And I noticed this one lost converter here: https://github.com/xamarin/XamarinCommunityToolkit/blob/main/XamarinCommunityToolkit/Helpers/SafeAreaTypeConverter.shared.cs

Not sure if it should be moved into the converters namespace

AndreiMisiukevich:
> Not sure if it should be moved into the converters namespace

I believe, it shouldn't. SafeAreaTypeConverter is subclass of **TypeConverter** (not implementing **IValueConverter**). So this one is used for xaml purposes. (Determines how a string should be parsed to SafeArea)

XamarinCommunityToolkit.UnitTests/Converters/ListIsNullOrEmptyConverter_Tests.cs
Char0394: Yes, you are right. Will change it in the next PR.

Issue #158, opened 2020-07-12
[Bug] EventToCommandBehavior throws ArgumentException with ListView events


almirvuk:
### Description
When using `EventToCommandBehaviour` implementation from `XamarinCommunityToolkit` with for example ListView and its events `ItemTapped` or `ItemSelected` event. Same approach and the same sample code is working with other implementation of `EventToCommandBehavior`, for example, the one from Prism library.

### Steps to Reproduce

1. Add behaviour to ListView and use it with ItemTapped event.
2. Debug the EventToCommandBehavior implementation

### Expected Behavior
To not throw an exception.

### Actual Behavior
It throws an exception on line 79, after calling:
`eventInfo.AddEventHandler(View, eventHandler);`

### Basic Information

```
System.ArgumentException has been thrown:

Object of type 'System.EventHandler' cannot be converted to type 'System.EventHandler`1[Xamarin.Forms.ItemTappedEventArgs]'.
```

### Reproduction Link
I have a reproduction sample and it is available on a pull request for ItemTappedeEventArgsConverter in ConvertersGallery page converter [here](https://github.com/xamarin/XamarinCommunityToolkit/pull/157).

jfversluis:
Fixed by #159

PR #159, opened 2020-07-12
[Bug] EventToCommandBehavior throws ArgumentException

AndreiMisiukevich:
### Description of Change ###
We should create event handler for observing the event according to **eventInfo.EventHandlerType**.

```csharp
eventHandler = eventHandlerMethodInfo.CreateDelegate(eventInfo.EventHandlerType, this)
```

### Bugs Fixed ###
- Related to issue #158 

### API Changes ###
None

### Behavioral Changes ###
No crash when used with ListView events

### PR Checklist ###

- [ ] Updated documentation

### DIFF
--- a/XamarinCommunityToolkit.UnitTests/Behaviors/EventToCommandBehavior_Tests.cs
+++ b/XamarinCommunityToolkit.UnitTests/Behaviors/EventToCommandBehavior_Tests.cs
@@ -1,0 +1,103 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Xamarin.Forms;
+using Xamarin.Forms.Internals;
+using XamarinCommunityToolkit.Behaviors;
+using Xunit;
+
+namespace XamarinCommunityToolkit.UnitTests.Behaviors
+{
+    public class EventToCommandBehavior_Tests
+    {
+        public EventToCommandBehavior_Tests()
+            => Device.PlatformServices = new MockPlatformServices();
+
+        [Fact]
+        public void ArgumentExceptionIfSpecifiedEventDoesNotExist()
+        {
+            var listView = new ListView();
+            var behavior = new EventToCommandBehavior
+            {
+                EventName = "Wrong Event Name"
+            };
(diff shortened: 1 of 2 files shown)


jfversluis:
Can we write a test for this so we make sure this doesn't happen again?

AndreiMisiukevich:
> Is it covered by the one from Almir?
No, it isn't covered by any unit test

> Can we write a test for this so we make sure this doesn't happen again?
RegisterEvent is a private method that is internally when the behavior is attached to some view.
Such stuff should be covered by UI tests. Also, we can expose a public static method with this logic and cover it by unit tests

jfversluis:
OK, RegisterEvent is, but what if we write a test for the entire behavior? We should be able to write a test that creates some dummy `View` element, attach the behavior, unattach the behavior etc. and see if throws something or not? Or am I missing something?

AndreiMisiukevich:
> OK, RegisterEvent is, but what if we write a test for the entire behavior? We should be able to write a test that creates some dummy `View` element, attach the behavior, unattach the behavior etc. and see if throws something or not? Or am I missing something?

Got it, yes we can do it.
But I thought **Unit** tests should be as simple as possible.

jfversluis:
Since the thing we want to test is private, testing the public part is as simple as we can go 😉 

You're right, but I would rather have one test which technically might not be a unit as per the definition than no test at all.

AndreiMisiukevich:
@jfversluis 

:(

image

AndreiMisiukevich:
Well, found a workaround.

@jfversluis I added tests

jfversluis:
Awesome work @AndreiMisiukevich thanks! Where was that exception coming from exactly?

XamarinCommunityToolkit.UnitTests/Behaviors/EventToCommandBehavior_Tests.cs
```diff
@@ -0,0 +1,103 @@
...
+            {
+                EventName = nameof(ListView.ItemTapped)
+            };
+            listView.Behaviors.Add(behavior);
+        }
+    }
+
+    #region Mock Services
+
+    class MockPlatformServices : IPlatformServices
```
jfversluis, line 44: Aah I see. This is something that is done in Forms too :) Great!

PR #164, opened 2020-07-16
Sample UI update

sthewissen:
### Description of Change ###

Changes the sample app to look more like this:

image

### Bugs Fixed ###

- Related to issue #129

### PR Checklist ###

- [ ] Updated documentation


sthewissen:
The lowest level of pages (individual samples) haven't been touched yet. Those are next on the list. The gallery pages should all be in there and should work atm.

XamarinCommunityToolkitSample.iOS/Renderers/NoLineNavigationRenderer.cs
```diff
@@ -0,0 +1,32 @@
...
+[assembly: ExportRenderer(typeof(BaseNavigationPage), typeof(NoLineNavigationPageRenderer))]
+namespace XamarinCommunityToolkitSample.iOS.Renderers
+{
+    public class NoLineNavigationPageRenderer : NavigationRenderer
+    {
+        public override void ViewWillAppear(bool animated)
+        {
+            base.ViewWillAppear(animated);
+
+            if (Element is NavigationPage)
```
AndreiMisiukevich, line 16: I guess this check can be omitted since NavigationRenderer is designed for work with NavigationRenderer :)

Or

```
if (!(Element is NavigationPage))
    return;
```

XamarinCommunityToolkitSample.iOS/Renderers/NoLineNavigationRenderer.cs
```diff
@@ -0,0 +1,32 @@
...
+                NavigationBar.SetBackgroundImage(new UIKit.UIImage(), UIKit.UIBarMetrics.Default);
+                NavigationBar.ShadowImage = new UIKit.UIImage();
+                NavigationBar.ClipsToBounds = true;
+
+                try //Newest iOS version fix - trycatch isn't optimal
+                {
+                    NavigationBar.ScrollEdgeAppearance.ShadowImage = new UIKit.UIImage();
+                    NavigationBar.ScrollEdgeAppearance.ShadowColor = null;
+                }
+                catch (Exception) { }
```
AndreiMisiukevich, line 28: Should we log such exceptions? (Console.WriteLine) or Debug.Log.

Usually, when I need to do it I write a comment

```csharp
try
{
    //Exception
}
catch
{
    // Suppressed
}

Comment (AndreiMisiukevich) [reply XamarinCommunityToolkitSample.iOS/Renderers/NoLineNavigationRenderer.cs:28]: Ah, sorry, that's sample app :) So it doesn't matter then

File: XamarinCommunityToolkitSample/Models/SectionModel.cs [hunk 3]
```diff
@@ -0,0 +1,28 @@
...
+            Description = description;
+        }
+
+        public SectionModel(Type type, string title, Color color, string description)
+        {
+            Type = type;
+            Title = title;
+            Description = description;
+            Color = color;
+        }
```
AndreiMisiukevich, line 21:
```suggestion
        public SectionModel(Type type, string title, string description)
            : this(type, title, Color.Default, description) { }

        public SectionModel(Type type, string title, Color color, string description)
        {
            Type = type;
            Title = title;
            Description = description;
            Color = color;
        }
```

AndreiMisiukevich:
Awesome work!

Char0394:
Good work @sthewissen :)

XamarinCommunityToolkitSample/ViewModels/Converters/ConvertersGalleryViewModel.cs
```diff
@@ -1,15 +1,19 @@
...
     {
-        public IEnumerable<ConverterSectionModel> Items { get; } =
-            ((ConverterSectionId[])Enum
-            .GetValues(typeof(ConverterSectionId)))
-            .Select(id => new ConverterSectionModel(id));
+        public IEnumerable<SectionModel> Items { get; } = new List<SectionModel> {
+            new SectionModel(
+                typeof(ItemTappedEventArgsPage),
+                "ItemTappedEventArgs",
+                Color.FromHex("#498205"),
```
Char0394, line 14: @sthewissen What do you think if instead of having those colors hardcoded in the ViewModel, we have kind of a converter, that returns random colors (based on a list of colors pre-defined)? So if tomorrow we add more items to the list we don't have to think about what color to assign.

XamarinCommunityToolkitSample/ViewModels/Behaviors/BehaviorsGalleryViewModel.cs
```diff
@@ -1,15 +1,28 @@
...
+            ),
+            new SectionModel(
+                typeof(EventToCommandBehaviorPage),
+                "EventToCommandBehavior",
+                "Turns any event into a command that can be bound to."
+            ),
+            new SectionModel(
+                typeof(NumericValidationBehaviorPage),
+                "NumericValidationBehavior",
+                "Changes an Entry's text color when an invalid numeric value is provided."
```
Char0394, line 24: _(Not something that have to be changed now)_
Are we planning to add localization to the app? Maybe we can have those strings in a Resource file, so if we want to support different languages later is easier to add.

jfversluis: That's a great idea! We might borrow some inspiration from GitTrends

Char0394: I guess we should probably open a new issue for it.

jfversluis: Yes! Please do so we can track it better

XamarinCommunityToolkitSample/ViewModels/Converters/ConvertersGalleryViewModel.cs
maximusfatigue: Char0394 @ this is a fantastic idea, Can you share some code for this as I need something like this for my project

AndreiMisiukevich: @maximusfatigue
https://github.com/xamarin/XamarinCommunityToolkit/blob/main/XamarinCommunityToolkit/Views/AvatarView.shared.cs#L204-L206

Issue #154, opened 2020-07-07
[Bug] Rename namespaces


jfversluis:
Based of #114 I guess we need someone to rename all the things in terms of namespaces.

So, everything from `XamarinCommunityToolkit.*` to `Microsoft.Toolkit.Xamarin.Forms.*`

almirvuk:
I can do this. :)

bartlannoeye:
Was about to pick it up, but you beat me to it ;)

jfversluis:
Go for it!

jfversluis:
Note to self: we also probably want to rename the NuGet to `Microsoft.Toolkit.Xamarin.Forms`. Either make that part of this change or a separate one.

almirvuk:
I will probably take this one on Friday night or Saturday morning. :)

PR #166, opened 2020-07-18
RangeSlider

AndreiMisiukevich:
### Description of Change ###

**RangeSlider** works as a Slider with two thumbs allowing select two values (Lower and Upper) which represent "Range".
Could be very useful for some apps.

Jul-14-2020 01-16-03

**Video**:
https://twitter.com/Andrik_Just4Fun/status/1283181731584053249

**TODO:**
I'd like to enhance this control by adding new properties
- Orientation (Horizontal | Vertical)
- ValueLabelPosition (Start | End) ~ (Top | Bottom for Horizontal Orientation and Left | Right for Vertical one)

### Bugs Fixed ###

- Related to issue #6

### API Changes ###
Supported Properties

| Name                        | Default Value | Description |
| --------------------------- | ----------- | ----------- |
| LowerValue                  | 0.0         | Current lower value |
| UpperValue                  | 1.0         | Current upper value |
| MinimumValue                | 0.0         | Minimum value |
| MaximumValue                | 1.0         | Maximum value |
| ThumbSize                   | 30.0        | Thumb size *(if you want to set the same value for Lower & Upper Thumbs)* |
| LowerThumbSize              | -1.0        | Lower Thumb size |
| UpperThumbSize              | -1.0        | Upper Thumb size |
| TrackSize                   | 3.0         | Track size |
| ThumbColor                  | Default     | Thumb color *(if you want to set the same value for Lower & Upper Thumbs)* |
| LowerThumbColor             | Default     | Lower Thumb color |
| UpperThumbColor             | Default     | Upper Thumb color |
| TrackColor                  | Default     | Track color |
| TrackHighlightColor         | Default     | Track highlight color |
| ThumbBorderColor            | Default     | Thumb border color *(if you want to set the same value for Lower & Upper Thumbs)* |
| LowerThumbBorderColor       | Default     | Lower Thumb border color |
| UpperThumbBorderColor       | Default     | Upper Thumb border color |
| TrackBorderColor            | Default     | Track border color |
| TrackHighlightBorderColor   | Default     | Track highlight border color |
| ValueLabelStyle             | Default     | Value label style *(if you want to set the same value for Lower & Upper Thumbs)* |
| LowerValueLabelStyle        | null        | Lower Value label style |
| UpperValueLabelStyle        | null        | Upper Value label style |
| ValueLabelStringFormat      | {0:0.##}    | Value Label String Format |
| LowerThumbView              | null        | Lower Thumb View |
| UpperThumbView              | null        | Upper Thumb View |
| ValueLabelSpacing           | 5.0         | Value label spacing *(from thumb to value label)* |
| ThumbRadius                 | -1.0        | Thumb radius *(if you want to set the same value for Lower & Upper Thumbs)* |
| LowerThumbRadius            | -1.0        | Lower Thumb radius |
| UpperThumbRadius            | -1.0        | Upper Thumb radius |
| TrackRadius                 | -1.0        | Track radius |

### Behavioral Changes ###
This is a new control

### PR Checklist ###

- [ ] Updated documentation

### DIFF
--- a/XamarinCommunityToolkit/Views/RangeSlider.shared.cs
+++ b/XamarinCommunityToolkit/Views/RangeSlider.shared.cs
@@ -1,0 +1,615 @@
+﻿using System.Collections.Generic;
+using Xamarin.Forms;
+using Xamarin.Forms.Internals;
+using static System.Math;
+using static Xamarin.Forms.AbsoluteLayout;
+using System.Runtime.CompilerServices;
+
+namespace XamarinCommunityToolkit.Views
+{
+    public class RangeSlider : TemplatedView
+    {
+        const double enabledOpacity = 1;
+
+        const double disabledOpacity = .6;
+
+        public static BindableProperty MinimumValueProperty
+            = BindableProperty.Create(nameof(MinimumValue), typeof(double), typeof(RangeSlider), .0, propertyChanged: OnMinimumMaximumValuePropertyChanged);
+
+        public static BindableProperty MaximumValueProperty
+            = BindableProperty.Create(nameof(MaximumValue), typeof(double), typeof(RangeSlider), 1.0, propertyChanged: OnMinimumMaximumValuePropertyChanged);
+
+        public static BindableProperty LowerValueProperty
+            = BindableProperty.Create(nameof(LowerValue), typeof(double), typeof(RangeSlider), .0, BindingMode.TwoWay, propertyChanged: OnLowerUpperValuePropertyChanged, coerceValue: CoerceValue);
+
+        public static BindableProperty UpperValueProperty
+            = BindableProperty.Create(nameof(UpperValue), typeof(double), typeof(RangeSlider), 1.0, BindingMode.TwoWay, propertyChanged: OnLowerUpperValuePropertyChanged, coerceValue: CoerceValue);
+
+        public static BindableProperty ThumbSizeProperty
+            = BindableProperty.Create(nameof(ThumbSize), typeof(double), typeof(RangeSlider), 30.0, propertyChanged: OnLayoutPropertyChanged);
+
+        public static BindableProperty LowerThumbSizeProperty
+            = BindableProperty.Create(nameof(LowerThumbSize), typeof(double), typeof(RangeSlider), -1.0, propertyChanged: OnLayoutPropertyChanged);
+
+        public static BindableProperty UpperThumbSizeProperty
+            = BindableProperty.Create(nameof(UpperThumbSize), typeof(double), typeof(RangeSlider), -1.0, propertyChanged: OnLayoutPropertyChanged);
+
+        public static BindableProperty TrackSizeProperty
+            = BindableProperty.Create(nameof(TrackSize), typeof(double), typeof(RangeSlider), 3.0, propertyChanged: OnLayoutPropertyChanged);
+
+        public static BindableProperty ThumbColorProperty
+            = BindableProperty.Create(nameof(ThumbColor), typeof(Color), typeof(RangeSlider), Color.Default, propertyChanged: OnLayoutPropertyChanged);
+
+        public static BindableProperty LowerThumbColorProperty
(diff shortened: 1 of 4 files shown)


XamarinCommunityToolkit/Views/RangeSlider.shared.cs
```diff
@@ -0,0 +1,625 @@
...
+            => color == Color.Default
+                ? defaultColor
+                : color;
+
+        double GetDoubleOrDefault(double value, double defaultSize)
+            => value < 0
+                ? defaultSize
+                : value;
+
+        sealed class RangeSliderLayout : AbsoluteLayout
```
jfversluis, line 583: Should we move this into its own file maybe? Have a folder per control in the project and group files in there? This class isn't too big though. But if we get more controls like these that might help readability and discoverability

AndreiMisiukevich: Mmm, I don't think so. This layout doesn't make any sense without RangeSlider. And it even exists only because that's only one way I found how we can set content to TemplatedView.


This is how I worked around it for Expander in XF 
https://github.com/xamarin/Xamarin.Forms/blob/719fc7a604ff0cce8922d717c99bfb0fa17e35e0/Xamarin.Forms.Core/Expander.cs#L59

but InternalChildren collection is internal indeed :) So we cannot put control there directly, that is why I created this layout and put inside RangeSlider as a private class.

AndreiMisiukevich: Hmmm, just realized that child is added as soon as we register control template 
```ControlTemplate = new ControlTemplate(typeof(RangeSliderLayout));``

So probably we can remove this internal class at all

Comment (AndreiMisiukevich) [reply XamarinCommunityToolkit/Views/RangeSlider.shared.cs]: @jfversluis got rid of useless SliderLayout

File: XamarinCommunityToolkit/Views/RangeSlider.shared.cs [hunk 2]
```diff
@@ -0,0 +1,615 @@
+﻿using System.Collections.Generic;
+using Xamarin.Forms;
+using Xamarin.Forms.Internals;
+using static System.Math;
+using static Xamarin.Forms.AbsoluteLayout;
+using System.Runtime.CompilerServices;
+
+namespace XamarinCommunityToolkit.Views
```
jfversluis, line 8: After this is merged and #154 we need to move this to `Microsoft.Toolkit.Xamarin.Forms.UI.Controls`

jfversluis:
Played with the sample for a little on the stream last night. Works great, lots of options, I love it! Came across a couple of bugs. Not sure if the bugs are in the sample or the control, check it out here: https://www.twitch.tv/videos/685508393?t=01h03m16s it should get you to the right time in the stream, if not, it starts around 1 hour and 3 minutes.

AndreiMisiukevich:
> Played with the sample for a little on the stream last night. Works great, lots of options, I love it! Came across a couple of bugs. Not sure if the bugs are in the sample or the control, check it out here: https://www.twitch.tv/videos/685508393?t=01h03m16s it should get you to the right time in the stream, if not, it starts around 1 hour and 3 minutes.

Interesting... I didn't catch such bugs during testing in my repo. Perhaps it's related to XF version. Anyway, I will check, thanks!

AndreiMisiukevich:
@jfversluis ah... thumb size is fixed (it was sample issue)

AndreiMisiukevich:
As for "Shape issue"
I know it's kind of a mess, but there is no bug there.
Actually there is no property that responds for "Shape", but there is "CornerRadius" property. By default, it has "-1.0" value, which means that the corner radius should depend on thumb size (to make thumb as a circle).

 So "circle" is a default one.

When you disabled switch (and made thumbs as squares) basically you set "CornerRadius" to 0.
So, "LowerThumbRadius" and "UpperThumbRadius" are still "-1.0" (default value). That means that 0 should be applied to all.

I guess I should update the sample to make it possible to control thumbs radiuses

AndreiMisiukevich:
@jfversluis sample is updated

Issue #114, opened 2020-05-29
[Discussion] Project branding / namespace naming


sthewissen:
So here we are at the reboot of the Community Toolkit. Let's get this show on the road! There are a few things I'm wondering about that we might want to tackle first though, so I'm opening some issues to have these discussions. So let's get to it!

**Namespace naming** 
I think we can all agree that the "Community Toolkit" part of the name is fine. It's the other parts that I'm wondering about both from a public naming perspective and a namespace naming perspective. With .NET MAUI on the horizon, we can choose to either go with a bit more of a platform-independent namespace naming scheme or face the fact that we will have to adjust somewhere in the future :) I personally don't mind the latter as I don't think it'll be a big effort.

The current namespace of the existing codebase seems to be `Xamarin.Toolkit`, which I personally _would_ consider changing to `Xamarin.CommunityToolkit` to align it a bit better. 

**NuGet naming**
With regards to the actual NuGet package we will be putting out there are also a few different routes we could take:

- Xamarin.CommunityToolkit
- Community Toolkit for Xamarin
- FormsCommunityToolkit (I believe is the current name)
- Xamarin.Toolkit
- etc..

For discoverability, I would include the Xamarin name in there somewhere. Regarding the author, I wonder if we should put Microsoft there or something different. As it's "only" an endorsed project but not specifically made by Microsoft I can imagine that MS wouldn't want that there though. Perhaps @davidortinau can chime in on that one as he might have some insights regarding the customers that only want to use Microsoft packages 😀

**Logo!**
Every project starts with a cool logo. We all know this. I'm just experimenting a bit at this point, but here's a suggestion to take the Essentials icon as a base and change it onto something that indicates the group/community feeling by connecting all the separate orbs:

![image](https://user-images.githubusercontent.com/2419439/83231986-ad1be000-a18c-11ea-81b4-d555df4cb559.png)

Depechie:
For reference in regards of the Namespace and Nuget naming, we can always look over the wall at what is being done with the UWP community toolkit.
Their namespace is: Microsoft.Toolkit.Uwp as is their Nuget package.
The repo is underneath WindowsCommunityToolkit.

So, yes having a namespace with Microsoft reference seems a good suggestion.

AndreiMisiukevich:
Since it's **Xamarin.Forms** addition, perhaps it would make sense to extend Xamarin.Forms namespace.

```csharp
namespace Xamarin.Forms.CommunityKit
{
}
[...]

PR #167, opened 2020-07-18
Microsoft.Toolkit.Xamarin.Forms namespace rename.

almirvuk:
### Description of Change ###

This pull request is based on bug #154.

Everything from `XamarinCommunityToolkit.*` a namespace is changed to `Microsoft.Toolkit.Xamarin.Forms.*`

After this rename whole solution is building successfully. 

Note: Still need to rename folders, tried that already but it makes all the files removed and added in sense on git. I want to talk with other community members about this question.


XamarinCommunityToolkit.UnitTests/Converters/ItemSelectedEventArgsConverter_Tests.cs
```diff
@@ -1,11 +1,11 @@
-﻿using XamarinCommunityToolkit.Converters;
+﻿using Microsoft.Toolkit.Xamarin.Forms.Converters;
```
bartlannoeye, line 1: Do we care about namespace ordering? If so log an issue and I could fix with R#

almirvuk: Usually, I have the habit to order them. In my last two PRs, I wanted to order them from longer to shorten ones. 😊 Not sure if there is the policy on that for the entire project. But I am ok with idea to have them ordered. 😊

almirvuk:
@bartlannoeye I will try with GitKraken for rename, currently, I am using the GitHub Desktop client and it registers the removed/added changes for rename of the folder names.

XamarinCommunityToolkit.UnitTests/Converters/ItemSelectedEventArgsConverter_Tests.cs
bartlannoeye: Typically it's all System namespaces and then other namespaces, ordered alphabetically

jfversluis:
Looks like the fail was because of the renamed csproj file which wasn't changed in the yaml :) fixed that for you.

We also might want to consider renaming the filesystem folders

jfversluis:
image

:(

Do you have time for this @almirvuk ? I was planning on doing a live stream later today anyway to test some stuff. So else I can go through this later today/tomorrow.

almirvuk:
:( I see it, this is not even with the main branch anymore... I can resolve them but later this night around 8-9 PM (EU time)

jfversluis:
Yeah I just merged Steven's Sample app UI. So it's probably just picking the right namespace and making changes in what he did. Nothing too exciting. Let's coordinate around then on who will do it!

almirvuk:
@jfversluis it should be ok now. 😄

jfversluis:
Some unexpected behavior because we use the Xamarin.Forms namespace which conflicts with the regular Xamarin.Forms namespace. We need to choose something different :(

jfversluis:
Whoops :D Not make these changes too often... 

@almirvuk and others, if you can take the NuGet from [here](https://dev.azure.com/xamarin/_apis/resources/Containers/7206868/nuget?itemPath=nuget%2FMicrosoft.Toolkit.Xamarin.Forms.1.0.0-preview1206.nupkg) and see that no namespace collisions happen with regular `Xamarin.Forms` then it's all good for me!

AndreiMisiukevich:
> Whoops :D Not make these changes too often...
> 
> @almirvuk and others, if you can take the NuGet from [here](https://dev.azure.com/xamarin/_apis/resources/Containers/7206868/nuget?itemPath=nuget%2FMicrosoft.Toolkit.Xamarin.Forms.1.0.0-preview1206.nupkg) and see that no namespace collisions happen with regular `Xamarin.Forms` then it's all good for me!

image

jfversluis:
Oh hm interesting. Can you click on the status checks here in GitHub then to azure DevOps and see the artifacts that were produced?

AndreiMisiukevich:
> Oh hm interesting. Can you click on the status checks here in GitHub then to azure DevOps and see the artifacts that were produced?

yes, that works

almirvuk:
https://giphy.com/gifs/hero0fwar-karmawhore-rhyming-g9582DNuQppxC

almirvuk:
image

@jfversluis @AndreiMisiukevich

clairernovotny:
Hey folks -- Microsoft.Toolkit.* is an issue in that the namespace is already used by a different project and it will cause confusion. 

Can you please rename to something else before you release?

jfversluis:
I don't see how that can be. We've gone through the proper channels for this, has been approved and I feel Microsoft.Toolkit is way too generic and it can easily be identified by the platform that comes after it.

michael-hawker:
Hey @jfversluis, I setup a meeting for us Monday, didn't realize there was a big time zone difference when I tried to set something up today. Please feel free to forward any info you have ahead of time about who you cleared this with, as we haven't been involved in any of these discussions.

As was [called out in the original thread already](https://github.com/xamarin/XamarinCommunityToolkit/issues/114#issuecomment-636378119), there could be confusion around if 'Forms' in out existing naming is WinForms or Xamarin Forms, but _it's also about project ownership and communities_.

Outside of what's technical feasible or not, if we both ship multiple things under the namespace `Microsoft.Toolkit.*` and with similar named packages, _folks will start reporting issues for both of our projects on each others repos_.

It looks like [this project already tried to own](https://www.nuget.org/packages/Xamarin.Toolkit.Effects/) `Xamarin.Toolkit` in the past for naming, was there a problem with going back to use that? It seems like a great more direct name.

jfversluis:
Hey @michael-hawker! Sorry for the confusion. I did not see your meeting, sorry for not being clear and/or the timezone confusion. I was meaning to get back to this on Monday as it's about to come up on bedtime here.

I (and we) mean no harm. I'm sure we can get this resolved and I'd hate to go back and forth out in the open like this so let's make sure we have a meeting about this. 

The thing that surprises me the most is that I have been able to reserve this name and prefix. In fact, I notice that "we" have the verified badge while you haven't. So it seems that your registration might not be done properly?

Anyway. Let's set something up and get it resolved. I'm not out to rub anyone the wrong way :)

michael-hawker:
Thanks @jfversluis for some of the background! Not sure if my Teams messages are going through as I see an 'x' in your status, but let me know if the meeting time I setup works for you on Monday. Definitely just trying to make sure we're all on the same page and get up-to-speed on all the project goals. 🙂

The Windows Community Toolkit is supposed to have the package space reserved on NuGet, but we're a .NET Foundation project, so the NuGet status here being also under `Microsoft.*` in general is an odd space with some issues, so there was definitely a hiccup here (as called out in the thread with the NuGet team).

To clarify you're saying 'Xamarin.Toolkit' was tossed as in the future Xamarin will be known as MAUI, eh? So then something like `MAUI.Toolkit` would work but be more confusing now? We have a similar type of issue with WinUI 3 approaching for us too. Anyway, we can follow-up on the call; as you say it'll be quicker that way.

I definitely see opportunities to collaborate. We're doing a lot more with .NET Standard in the Windows Community Toolkit, so it would make sense to contribute those in one space. For instance, we have a new MVVM library and would love to have some Xamarin expertise/feedback in building out a sample.

On the other-side, we're headed down the WinUI 3 Reunion path vs. MAUI, so I don't know how well-aligned those goals would be; lots to talk about, but let's kick things off Monday. Looking forward to chatting.

PR #175, opened 2020-07-31
PasswordValidationBehavior

LucioMSP:
### Description of Change ###

Added a password  behavior. The code that I share below is to validate the password format. Which must conform to the following structure: At least 8 characters, at least 1 number, 1 lowercase, 1 uppercase, 1 special character [for example: R3dDr@g0n]

### Usage ###

Its operation is similar to the other Behaviors, as well as that of electronic mail and numeric.

### PR Checklist ###

- [ ] Updated documentation

### DIFF
--- a/XamarinCommunityToolkit/Behaviors/PasswordValidationBehavior.shared.cs
+++ b/XamarinCommunityToolkit/Behaviors/PasswordValidationBehavior.shared.cs
@@ -1,0 +1,12 @@
+﻿using System.Text.RegularExpressions;
+
+namespace Microsoft.Toolkit.Xamarin.Forms.Behaviors
+{
+    public class PasswordValidationBehavior : TextValidationBehavior
+    {
+        protected override string DefaultRegexPattern
+            => @"^(?=.*[A-Za-z])(?=.*\d)(?=.*[$@$!%*#?&])[A-Za-z\d$@$!%*#?&]{8,}$";
+
+        protected override RegexOptions DefaultRegexOptions => RegexOptions.IgnoreCase;
+    }
+}
--- a/XamarinCommunityToolkit/Microsoft.Toolkit.Xamarin.Forms.csproj
+++ b/XamarinCommunityToolkit/Microsoft.Toolkit.Xamarin.Forms.csproj
@@ -23,7 +23,11 @@
     <RepositoryUrl>https://github.com/xamarin/XamarinCommunityToolkit</RepositoryUrl>
     <PackageReleaseNotes>
       
+      
+      
       <!-- See if we can host these on MS docs? -->
+    
+    
     
     </PackageReleaseNotes>
     <DefineConstants>$(DefineConstants);</DefineConstants>
@@ -111,5 +115,10 @@
     <ItemGroup Condition=" $(TargetFramework.StartsWith('tizen')) ">
       <Compile Include="**\*.tizen.cs" />
       <Compile Include="**\*.tizen.*.cs" />
+    </ItemGroup>
+    <ItemGroup>
+      <Compile Update="Behaviors\PasswordValidationBehavior.shared.cs">
+        <SubType></SubType>
+      </Compile>
     </ItemGroup>
 </Project>
--- a/XamarinCommunityToolkitSample/Pages/Behaviors/PasswordValidationBehaviorPage.xaml
+++ b/XamarinCommunityToolkitSample/Pages/Behaviors/PasswordValidationBehaviorPage.xaml
@@ -1,0 +1,25 @@
+﻿<?xml version="1.0" encoding="UTF-8"?>
+<pages:BasePage xmlns="http://xamarin.com/schemas/2014/forms"
+                xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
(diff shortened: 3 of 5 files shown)


sthewissen:
First of all, thanks for taking your time to contribute!

I'm not against a behavior that does something like this, but a hardcoded password policy to me at least isn't something we should want. Not every user's policy is the same and security is a whole world on its own. I don't think the world even agrees on a singular password policy to rule them all. Forcing a specific one is not something we should probably want to get into.

This is also the reason a similar behavior was declined in a different PR. To supply a custom password policy we have the `TextValidationBehavior` that lets users choose their own validation RegEx.

LucioMSP:
I understand, but hey, at least here in Mexico, software developers do allow us this type of format, but I understand what you are saying. In the same way I will look for how to contribute.

AndreiMisiukevich:
Hi, @LucioMSP Thanks for your contribution. I really appreciate it!
But I think password-policy is a project-specific thing. I'm not sure if that's possible to create a common behavior that covers most of possible cases.

Perhaps we could go ahead with the next approach.

1) We add a new enum type **PasswordFlags**
PasswordFlags { None, HasDigit, HasSpecialCharacter, HasLowercaseCharacter, HasUppercaseCharacter, AllowSpaces }
2) Add a bindable property PasswordFlagsProperty

I think such behavior we can merge, cuz it will be flexible :) @sthewissen your thought

sthewissen:
My question would then become, if you want to offer flexibility for passwords, wouldn't you then also add properties to specify quantities of what you want? E.g. at least 3 uppercase, 2 special characters etc. That's why I said I think something like this complicates manners to do right :)

Depechie:
I think the remark of @sthewissen to promote a text validation through regex, also for me seems a better suggestion.

roubachof:
If I may, I also would personally go for regex but don't forget that most developers are not comfortable with regex. So I think this kind of validator with custom properties could be nice.

AndreiMisiukevich:
@sthewissen @Depechie @roubachof 

- RequiredDigitsNumber (bindable, 0 is default)
- RequiredLowercaseCharactersNumber (bindable, 0 is default)
- RequiredUppercaseCharactersNumber (bindable, 0 is default)
- RequiredSpecialCharactersNumber (bindable, 0 is default)
- AllowSpaces (bindable, false is default)

How about this approach then?

I am trying to find a quite flexible and common solution 👍

LucioMSP:
Excellent, the truth seems perfect to me, and I think that if it would help many developers

jfversluis:
Just to add my thoughts in the mix.

Like a few already mentioned, thank you so much @LucioMSP for being one of the first to jump on this project and contribute. Very awesome :)

I think hard-coding the policy is not what we want. The other options seem good to me, can we somehow have both?

For the people not being familiar with regex or people who don't require it, they can use the flags. People with other needs will be able to use the regex? In fact, you kind of automatically do that now because of the regex pattern being available through the `TextValidationBehavior`, right?

If we do offer both, we might want to establish some kind of priority though. If one property is set, we don't do the other. Or maybe even throw an exception in this case to make clear that you can't use both at the same time.

AndreiMisiukevich:
https://github.com/xamarin/XamarinCommunityToolkit/issues/202

jfversluis:
@LucioMSP are you still up for doing work for this? Maybe we should think about it some more in an issue or something first? :)

LucioMSP:
@jfversluis You are absolutely right, I think I better review any error or improvement, my purpose is to support and see how I do it.

jfversluis:
Awesome! Let us know if you need any help or input :)

AndreiMisiukevich:
Hi @LucioMSP :) Are you still interested in contributing this feature?

LucioMSP:
Yes, but nothing occurred to me anymore, how do you think I can contribute? cc @AndreiMisiukevich

AndreiMisiukevich:
> @sthewissen @Depechie @roubachof
> 
> * RequiredDigitsNumber (bindable, 0 is default)
> * RequiredLowercaseCharactersNumber (bindable, 0 is default)
> * RequiredUppercaseCharactersNumber (bindable, 0 is default)
> * RequiredSpecialCharactersNumber (bindable, 0 is default)
> * AllowSpaces (bindable, false is default)
> 
> How about this approach then?
> 
> I am trying to find a quite flexible and common solution 👍

@LucioMSP how about this approach?

AndreiMisiukevich:
I guess this feature should be achieved via MultiValidationBehavior with nested validators. So every user will be able to define the desired bunch of validations.

So we should add several new validators
- HasDigitsValidationBehavior (DigitsNumber)
- ContainsLowercaseCharactersValidationBehavior (LowercaseCharactersNumber)
- ContainsUppercaseCharactersValidationBehavior (UppercaseCharactersNumber)
- ContainsSpecialCharactersValidationBehavior (SpecialCharactersNumber)
- PreventSpacesValidationBehavior

CC @jfversluis @sthewissen @LucioMSP

AndreiMisiukevich:
Thanks for your contribution!
But unfortunately, I have to close this PR due to it been inactive for a long time

PR #182, opened 2020-08-02
Implement ByteArrayToImageSourceConverter in converters.

almirvuk:
`ByteArrayToImageSourceConverter` is a very simple converter that allows you to convert an object of type `byte[]` to an object of a type `ImageSource`. 

Use-case: Sometimes in our projects, we are dealing with `byte[]` values for images and we need to convert them to `ImageSource` object in our view-model or code-behind but using this converter we don't need to write that logic all over again.

### Added:

`ByteArrayToImageSourceConverter`

### Sample

```
<c:ByteArrayToImageSourceConverter x:Key="ByteArrayToImageSourceConverter" />

<views:AvatarView Size="300"
                  Source="{Binding AvatarByteArray, Converter={StaticResource ByteArrayToImageSourceConverter}}"
                  HorizontalOptions="Center" VerticalOptions="Center" />
```

### PR Checklist ###

- [X] Has tests (if omitted, state reason in the description)
- [X] Has samples (if omitted, state reason in the description)
- [X] Changes adhere to the coding standard
- [X] Updated documentation


AndreiMisiukevich:
Thank you for your contribution. Just a few tweaks and we will merge it ;)

XamarinCommunityToolkit/Converters/ByteArrayToImageSourceConverter.shared.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>An object of type ImageSource.</returns>
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+             => value is byte[] imageBytes
+               ? ImageSource.FromStream(() => new MemoryStream(imageBytes))
+               : null;
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+            => throw new NotImplementedException();
```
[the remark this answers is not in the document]
almirvuk: Do these changes in the Convert method makes more sense now? @AndreiMisiukevich 😄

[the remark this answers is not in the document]
AndreiMisiukevich: Oh, sorry for the mess. int/ulong etc. was more as a question than change request. I'm not even sure that this is necessary.
Since this converter has the name "ByteToImageSourceConverter", probably it should work with byte arrays only.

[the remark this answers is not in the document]
almirvuk: It is ok, I needed to handle null and throwing exception in the right way 😄

almirvuk:
@roubachof Thank you for the suggestion, to be honest, I would keep this one and add another one called `ByteArrayToFileImageSourceConverter` with behaviour which you described in your comment. Not sure what others think about this idea. Later on, in the docs, we can describe what are the benefits and downsides of using any of these two. 😄

XamarinCommunityToolkitSample/ViewModels/Converters/ByteArrayToImageSourceViewModel.cs
```diff
@@ -0,0 +1,51 @@
...
+                using var response = await client.GetAsync(avatarUrl);
+                var imageBytes = await response.Content.ReadAsByteArrayAsync().ConfigureAwait(false);
+
+                Avatar = imageBytes;
+
+                IsBusy = false;
+            }
+            catch
+            {
+            }
```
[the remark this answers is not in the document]
almirvuk: This should be fixed by the latest commit.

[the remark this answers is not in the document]
almirvuk: This should be fixed by the latest commit.

XamarinCommunityToolkit.UnitTests/Converters/ByteArrayToImageSourceConverter_Tests.cs
```diff
@@ -13,20 +13,64 @@ public class ByteArrayToImageSourceConverter_Tests
...
+            {
+                var aByte = a.ReadByte();
+                var bByte = b.ReadByte();
+                if (aByte.CompareTo(bByte) != 0)
+                {
+                    return false;
+                }
+            }
+
+            return true;
```
AndreiMisiukevich, line 73:
```suggestion
            if (a == b)
                return true;
                
            if (a == null ||
                b == null ||
                a.Length != b.Length)
                return false;
                
            for (var i = 0; i < a.Length; i++)
                if (a.ReadByte() != b.ReadByte())
                    return false;
                    
            return true;
```

XamarinCommunityToolkit/Converters/ByteArrayToImageSourceConverter.shared.cs
```diff
@@ -21,34 +21,25 @@ public class ByteArrayToImageSourceConverter : ValueConverterExtension, IValueCo
...
+            throw new ArgumentException("Expected value to be of type byte[].", nameof(value));
+        }
 
-                case int[] intArray:
-                    imageBytes = new byte[intArray.Length * sizeof(int)];
-                    Buffer.BlockCopy(intArray, 0, imageBytes, 0, imageBytes.Length);
-                    break;
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            var streamImageSource = (StreamImageSource)value;
```
AndreiMisiukevich, line 35: I guess we should check if the value is null or StreamImageSource

If null then return null
if not is StreamImageSource throw an exception

XamarinCommunityToolkit/Converters/ByteArrayToImageSourceConverter.shared.cs
```diff
@@ -21,34 +21,25 @@ public class ByteArrayToImageSourceConverter : ValueConverterExtension, IValueCo
...
-                    imageBytes = new byte[ulongArray.Length * sizeof(ulong)];
-                    Buffer.BlockCopy(ulongArray, 0, imageBytes, 0, imageBytes.Length);
-                    break;
+            var streamFromImageSource = streamImageSource.Stream(System.Threading.CancellationToken.None).Result;
 
-                default:
-                    throw new ArgumentException("Expected value to be of type byte[].", nameof(value));
-            }
+            using var memoryStream = new MemoryStream();
+            streamFromImageSource.CopyTo(memoryStream);
```
AndreiMisiukevich, line 40: streamFromImageSource can be null

XamarinCommunityToolkit.UnitTests/Converters/ByteArrayToImageSourceConverter_Tests.cs
AndreiMisiukevich: I don't think we should raise ArgumentNullException if a == null || b == null
(btw we don't raise if both equal null cuz null == null // TRUE)

XamarinCommunityToolkit/Converters/ByteArrayToImageSourceConverter.shared.cs
almirvuk: Should be done in the latest commit.

XamarinCommunityToolkit/Converters/ByteArrayToImageSourceConverter.shared.cs
almirvuk: Should be done in the latest commit.

XamarinCommunityToolkit/Converters/ByteArrayToImageSourceConverter.shared.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            if (value is null)
+                return null;
+
+            if (value is StreamImageSource streamImageSource)
+            {
+                var streamFromImageSource  = streamImageSource.Stream(System.Threading.CancellationToken.None).Result;
```
AndreiMisiukevich, line 40:
```suggestion
                var streamFromImageSource = streamImageSource.Stream(System.Threading.CancellationToken.None).Result;
```

almirvuk: This happens when you are in a hurry and want to finish things asap :D... thanks for the review! @AndreiMisiukevich

XamarinCommunityToolkit/Converters/ByteArrayToImageSourceConverter.shared.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            if (value is null)
+                return null;
+
+            if (value is StreamImageSource streamImageSource)
+            {
+                var streamFromImageSource = streamImageSource.Stream(System.Threading.CancellationToken.None).Result;
```
Depechie, line 40: Any reason on why System.Threading is not added as an _using_?

almirvuk: Fixed in the last commit, thanks!

XamarinCommunityToolkitSample/Resx/AppResources.resx
```diff
@@ -171,6 +171,11 @@
               <value>A converter that allows you to extract the value from ItemSelectedEventArgs that can be used in combination with EventToCommandBehavior.</value>
           </data>
        
+         <!--ItemTappedSelectedEventArgs-->
+  
+         <data name="ByteArrayToImageSourceShortDescription" xml:space="preserve">
+              <value>A converter that allows you to convert byte array to an object of a type ImageSource.</value>
+          </data>
```
AndreiMisiukevich, line 178:
```suggestion
         </data>
```

XamarinCommunityToolkit/Converters/ByteArrayToImageSourceConverter.shared.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// </summary>
+        /// <param name="value">The value to convert.</param>
+        /// <param name="targetType">The type of the binding target property.</param>
+        /// <param name="parameter">Additional parameter for the converter to handle. Not implemented.</param>
+        /// <param name="culture">The culture to use in the converter.</param>
+        /// <returns>An object of type ImageSource.</returns>
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            if (value is null)
+                return default(ImageSource);
```
AndreiMisiukevich, line 26:
```suggestion
                return null;
```

sthewissen:
> @roubachof Thank you for the suggestion, to be honest, I would keep this one and add another one called `ByteArrayToFileImageSourceConverter` with behaviour which you described in your comment. Not sure what others think about this idea. Later on, in the docs, we can describe what are the benefits and downsides of using any of these two. 😄

LGTM, only thing I'm wondering is how we can keep track of things like "Later on, in the docs" to actually do them 😂

@jfversluis any suggestions?

almirvuk:
> > @roubachof Thank you for the suggestion, to be honest, I would keep this one and add another one called `ByteArrayToFileImageSourceConverter` with behaviour which you described in your comment. Not sure what others think about this idea. Later on, in the docs, we can describe what are the benefits and downsides of using any of these two. 😄
> 
> LGTM, only thing I'm wondering is how we can keep track of things like "Later on, in the docs" to actually do them 😂
> 
> @jfversluis any suggestions?

I was planning to do the docs for the stuff which I made in repo... and to cover those topics.

PR #183, opened 2020-08-02
Add observable collection extensions

krdmllr:
### Description of Change ###

Adds helpfull extension methods for `ObservableCollection<T>`. 

### Bugs Fixed ###

none

### API Changes ###

Added: 
 
- `public static void AddRange(this ObservableCollection<T> collection, IEnumerable<T> items)`
- `public static void SetRange(this ObservableCollection<T> collection, IEnumerable<T> items)`
 
### Behavioral Changes ###

none 

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description) -> Can add some if this is interesting
- [ ] Updated documentation

### DIFF
--- a/XamarinCommunityToolkit/Extensions/ObservableCollectionExtension.shared.cs
+++ b/XamarinCommunityToolkit/Extensions/ObservableCollectionExtension.shared.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace Microsoft.Toolkit.Xamarin.Forms.Extensions
+{
+    /// <summary>
+    /// Extension methods for <see cref="ObservableCollection{T}"/>.
+    /// </summary>
+    public static class ObservableCollectionExtension
+    {
+        /// <summary>
+        /// Adds a range of items to a <paramref name="collection"/>.
+        /// </summary>
+        /// <param name="collection">Collection where the items will be added to.</param>
+        /// <param name="items">Items to add to the collection.</param>
+        public static void AddRange<T>(this ObservableCollection<T> collection, IEnumerable<T> items)
+        {
+            foreach (var item in items)
+            {
+                collection.Add(item);
+            }
+        }
+
+        /// <summary>
+        /// Sets a range of items in a <paramref name="collection"/> and clears existing items.
+        /// </summary>
+        /// <param name="collection">Collection where the items will be added to.</param>
+        /// <param name="items">Items to set in the collection.</param>
+        public static void SetRange<T>(this ObservableCollection<T> collection, IEnumerable<T> items)
+        {
+            collection.Clear();
+            collection.AddRange(items);
+        }
+    }
+}


XamarinCommunityToolkit/Extensions/ObservableCollectionExtension.shared.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <summary>
+        /// Adds a range of items to a <paramref name="collection"/>.
+        /// </summary>
+        /// <param name="collection">Collection where the items will be added to.</param>
+        /// <param name="items">Items to add to the collection.</param>
+        public static void AddRange<T>(this ObservableCollection<T> collection, IEnumerable<T> items)
+        {
+            foreach (var item in items)
+            {
+                collection.Add(item);
```
Depechie, line 21: With this implementation each add would still generate a notification... I always tend to use a subclassed RangeObservableCollection that overrides the OnCollectionChanged to suppress the single add notifications.

Depechie, line 21: Add null checks.

IeuanWalker:
I use @jamesmontemagno `ObservableRangeCollection` from mvvm-helpers - 
https://github.com/jamesmontemagno/mvvm-helpers/blob/master/MvvmHelpers/ObservableRangeCollection.cs

I know its not a set extension methods, but not sure if the class in this PR is needed when the class in MVVM-Helpers does the same function and more :)

krdmllr:
Lets discuss if we want this extension at all. I can add tests, null checks any anything else then :)

AndreiMisiukevich:
Hi, thanks for your contribution :) But I tend to agree with @Depechie 
I'd like to see the RangeObservableCollection as a subclass of ObservableCollection.

Something like this

```csharp
	public class RangeObservableCollection<TEntity> : ObservableCollection<TEntity>
	{
		public RangeObservableCollection() { }

		public GalleyObservableCollection(IEnumerable<TEntity> collection) : base(collection) { }

		public void AddRange(IEnumerable<TEntity> collection)
			=> PerformBatchUpdate(collection, item => Items.Add(item));

		public void RemoveRange(IEnumerable<TEntity> collection)
			=> PerformBatchUpdate(collection, item => Items.Remove(item));

		public void ReplaceRange(IEnumerable<TEntity> collection)
		{
			Items.Clear();
			AddRange(collection);
		}

		void PerformBatchUpdate(IEnumerable<TEntity> collection, Action<TEntity> action)
		{
			if (collection == null)
				return;

			foreach (var item in collection)
				action.Invoke(item);
				
		    OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
		}
	}
```

sthewissen:
Also, namingwise I would perhaps prefer to call it `ObservableRangeCollection`. It's the alternative that most people use (the one by @jamesmontemagno) so seeing the same name will at least give those users the impression that it's the same thing.

XamarinCommunityToolkit/Extensions/ObservableCollectionExtension.shared.cs
jamesmontemagno: That is correct. You can just steal my ObservableRangeCollection. Steal all of Mvvmhelpers please!

sthewissen:
We also now have #190 so we probably need to figure out a course of action here :)

krdmllr:
I will close this one :) If the ObservableRangeCollection does not come from #190 I can help creating a new PR for another implementation.

PR #193, opened 2020-08-08
GTK support

AndreiMisiukevich:
### Description of Change ###
Should we support GTK as well?

### API Changes ###
Possibility to write custom classes for GTK

### Behavioral Changes ###
None

### PR Checklist ###

- [ ] Updated documentation


jfversluis:
While I like to support as much platforms as possible this adds 2MB to our library which will probably be useless for most people? If we do, I want to be sure that those libraries do not end up in peoples apps and make their apps unnecessarily bigger.

Also, I am not familiar with GTK, but isn't there a better way to link that library instead of putting DLLs in our repo?

AndreiMisiukevich:
> While I like to support as much platforms as possible this adds 2MB to our library which will probably be useless for most people? If we do, I want to be sure that those libraries do not end up in peoples apps and make their apps unnecessarily bigger.
> 
> Also, I am not familiar with GTK, but isn't there a better way to link that library instead of putting DLLs in our repo?

it's referenced only by GTK projects.
As for dlls, I know, that's odd :) but seems it's only one way how to do it (Forms does the same way, no?)

But probably we can postpone with this action for now

jfversluis:
For that last bit probably @jsuarezruiz knows

AndreiMisiukevich:
Do we decline this PR for now?

PR #191, opened 2020-08-07
Big Rename (again...)

jfversluis:
I think we all know what this is about

PR #195, opened 2020-08-08
Fail builds on formatting errors w/ dotnet format

jfversluis:
Added [`dotnet format`](https://github.com/dotnet/format) to the build pipeline. If there is any error in formatting it will fail the build.

Is this something we would want?

Ideally I wanted to setup something that runs this as a GitHub action and adds a commit to the PR with the format fixes. But it seems nothing is out there (that is working). I would want to create something for it, but in the meantime maybe we can use this.

PS. This build will probably fail until #191 is merged which has the formatting fixes over the code that is on `main` now. Adding DNM until that one is merged. In the meantime, let me hear your thoughts on this.


IeuanWalker:
@jfversluis be good to add this config file too - https://github.com/Xavalon/XamlStyler/wiki/External-Configurations

We use it in our team, so anyone who has the VS extension installed automatically uses the same XAML formatting.

jfversluis:
Thanks @IeuanWalker ! I'm not entirely familiar with all this yet... This is not included in the `.editorconfig` then?

jfversluis:
Closed in favor of #207

PR #206, opened 2020-08-17
Prevent designer file of being comitted

jfversluis:
Each simple change potentially commits the Android designer file with thousands of lines.

Added it to the `.gitignore` file and removed the currently committed one.


AndreiMisiukevich:
awesome work :)

PR #245, opened 2020-08-26
Added some unit tests to check namespaces

jfversluis:
### Description of Change ###

Don't know if this makes any sense, but digging into the code the other day I noticed how easy it is to mix up namespaces. This will be especially true when working with extra subfolders like with `CameraView`.

Samples not added because... Tests.

If this does not make any sense to anyone, please tell me as well :D 
Or if you have ideas on how to add more checks like these.

### Bugs Fixed ###

N/A

### Behavioral Changes ###

N/A

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description) doh
- [ ] Updated documentation

### DIFF
--- a/XamarinCommunityToolkit.UnitTests/ConvertersNamespace_Tests.cs
+++ b/XamarinCommunityToolkit.UnitTests/ConvertersNamespace_Tests.cs
@@ -1,0 +1,61 @@
+﻿using System.Linq;
+using System.Reflection;
+using Xamarin.CommunityToolkit.Behaviors;
+using Xamarin.Forms;
+using Xunit;
+
+namespace Xamarin.CommunityToolkit.UnitTests
+{
+	public class Namespace_Tests
+	{
+		[Fact]
+		public void MakeSureConvertersAreInTheRightNamespace()
+		{
+			var allTheTypes = Assembly.GetAssembly(typeof(BaseBehavior)).GetTypes();
+			foreach (var type in allTheTypes.Where(t => t.Name.EndsWith("Converter") && t.GetInterface("IValueConverter") != null))
+			{
+				Assert.True(type.Namespace.Equals("Xamarin.CommunityToolkit.Converters"), $"{type.FullName} not in Xamarin.CommunityToolkit.Converters namespace");
+			}
+		}
+
+		[Fact]
+		public void MakeSureEffectsAreInTheRightNamespace()
+		{
+			var allTheTypes = Assembly.GetAssembly(typeof(BaseBehavior)).GetTypes();
+			foreach (var type in allTheTypes.Where(t => t.Name.EndsWith("Effect") && t.IsClass && t.IsSealed && t.IsAbstract))
+			{
+				Assert.True(type.Namespace.Equals("Xamarin.CommunityToolkit.Effects"), $"{type.FullName} not in Xamarin.CommunityToolkit.Effects namespace");
+			}
+		}
+
+		[Fact]
+		public void MakeSureMarkupExtensionsAreInTheRightNamespace()
+		{
+			var allTheTypes = Assembly.GetAssembly(typeof(BaseBehavior)).GetTypes();
+			foreach (var type in allTheTypes.Where(t => t.Name.EndsWith("Extension") && t.GetInterface("IMarkupExtension") != null))
+			{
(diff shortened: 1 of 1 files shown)


XamarinCommunityToolkit.UnitTests/ConvertersNamespace_Tests.cs
```diff
@@ -0,0 +1,61 @@
...
+using Xunit;
+
+namespace Xamarin.CommunityToolkit.UnitTests
+{
+	public class Namespace_Tests
+	{
+		[Fact]
+		public void MakeSureConvertersAreInTheRightNamespace()
+		{
+			var allTheTypes = Assembly.GetAssembly(typeof(BaseBehavior)).GetTypes();
```
AndreiMisiukevich, line 14: I guess would be slightly better to use some "converter" if one day we decide to split behaviors and converter into several assemblies.

XamarinCommunityToolkit.UnitTests/ConvertersNamespace_Tests.cs
```diff
@@ -0,0 +1,61 @@
...
+			foreach (var type in allTheTypes.Where(t => t.Name.EndsWith("Converter") && t.GetInterface("IValueConverter") != null))
+			{
```
AndreiMisiukevich, line 16: Please don't use braces for one line loops.

Dura lex sed lex (c)

XamarinCommunityToolkit.UnitTests/ConvertersNamespace_Tests.cs
```diff
@@ -0,0 +1,61 @@
...
+				Assert.True(type.Namespace.Equals("Xamarin.CommunityToolkit.Behaviors"), $"{type.FullName} not in Xamarin.CommunityToolkit.Behaviors namespace");
+			}
+		}
+
+		[Fact]
+		public void MakeSureViewsAreInTheRightNamespace()
+		{
+			var allTheTypes = Assembly.GetAssembly(typeof(BaseBehavior)).GetTypes();
+			foreach (var type in allTheTypes.Where(t => t.Name.EndsWith("View") && t.IsSubclassOf(typeof(View))))
```
AndreiMisiukevich, line 55: Mmm, e.g. Expander won't have "View" postfix.
Will we add it? Or Should we workaround it somehow?

XamarinCommunityToolkit.UnitTests/ConvertersNamespace_Tests.cs
jfversluis: Good point! Done

XamarinCommunityToolkit.UnitTests/ConvertersNamespace_Tests.cs
jfversluis: I've removed the check to end in view and just see if it inherits from it

XamarinCommunityToolkit.UnitTests/ConvertersNamespace_Tests.cs
jfversluis: Whoops. Done

XamarinCommunityToolkit.UnitTests/ConvertersNamespace_Tests.cs
```diff
@@ -0,0 +1,59 @@
...
+namespace Xamarin.CommunityToolkit.UnitTests
+{
+	public class Namespace_Tests
+	{
+		[Fact]
+		public void MakeSureConvertersAreInTheRightNamespace()
+		{
+			var allTheTypes = Assembly.GetAssembly(typeof(InvertedBoolConverter)).GetTypes();
+
+			foreach (var type in allTheTypes.Where(t => t.Name.EndsWith("Converter") && t.GetInterface("IValueConverter") != null))
```
AndreiMisiukevich, line 19: Sorry, I don't want to be so picky, but I guess that we should use **nameof** as often as possible (for types and namespaces if it's possible)

e.g. nameof(IValueConverter), nameof(Xamarin.CommunityToolkit.Converters) etc.

jfversluis: No worries, that's what reviews are for!

For the `IValueConverter` yes. For the whole namespace it won't work because it will only return to you the last bit. I.e. `nameof(Xamarin.CommunityToolkit.Converters` returns only `Converters` and thus the assert will fail

AndreiMisiukevich (commented):
LGTM
Well done!

PR #244, opened 2020-08-26
CameraView iOS

jfversluis:
### Description of Change ###

Adds the iOS part (and shared part) of the `CameraView`

### Bugs Fixed ###

- Related to issue #211 

### API Changes ###

Added: 

- `public enum CameraOptions` enum for selecting front, back, default or external camera
- `public enum CameraFlashMode` enum for enabling, disabling, torch or auto the flash
- `public enum CameraCaptureOptions` enum for setting the camera to photo, video or default

```csharp
public class CameraMediaSource : MediaSource
{
	public static MediaSource FromStream(Func<Stream> stream) ;
	public static MediaSource FromResource(string resource, Type resolvingType);
	public static MediaSource FromResource(string resource, Assembly sourceAssembly = null);
}
```

```csharp
public class StreamMediaSource : MediaSource, IStreamImageSource
{
	protected CancellationTokenSource CancellationTokenSource { get; }
	public virtual Func<CancellationToken, Task<Stream>> Stream { get; set; } // Bindable
	protected override void OnPropertyChanged(string propertyName);
	protected void OnLoadingCompleted(bool cancelled);
	protected void OnLoadingStarted();
	public virtual Task<bool> Cancel();
}
```

```csharp
public class MediaCapturedEventArgs : EventArgs
{
         internal MediaCapturedEventArgs();
	public object Data { get; }
	public ImageSource Image { get; }
	public MediaSource Video { get; }
}
```

```csharp
public class CameraView : View
{
	public event EventHandler<MediaCapturedEventArgs> MediaCaptured;
	public event EventHandler<string> MediaCaptureFailed;
	public event EventHandler<bool> OnAvailable;

	public bool IsBusy { get; set; } // Bindable
	public bool IsAvailable { get; set; } // Bindable
	public CameraOptions CameraOptions { get; set; } // Bindable
	public bool SavePhotoToFile { get; set; } // Bindable
	public CameraCaptureOptions CaptureOptions { get; set; } // Bindable
	public bool KeepScreenOn { get; set; } // Bindable
	public bool VideoStabilization { get; set; } // Bindable
	public CameraFlashMode FlashMode { get; set; } // Bindable
	public Aspect PreviewAspect { get; set; } // Bindable
	public double Zoom { get; set; } // Bindable
	public double MaxZoom  { get; set; } // Bindable
	public void Shutter();

         internal event EventHandler ShutterClicked;
	internal void RaiseMediaCaptured(MediaCapturedEventArgs args);
[...]

Issue #211, opened 2020-08-19
[Enhancement] CameraView


jfversluis:
## Summary

There has been a PR on the Xamarin.Forms repo for [CameraView](https://github.com/xamarin/Xamarin.Forms/pull/4084). Now that Forms will not see any major new versions for a while, they have asked if it can come stay with us for a little. I have always wanted to see this control, so I'm very happy to take it in.

In the longer run, it might be cool to hook into the raw data feed and let people hook up barcode scanning or computer vision stuff, but that's just a thing I'm daydreaming about. First, let's get this in. I've copied some things from the PR linked above into this issue so you know what it's about. This is all based on just taking the code as is for now from said PR.

||Android|iOS|UWP|
|---|:---:|:---:|:---:|
|Preview|✅|✅|✅|
|Preview Aspect|✅|🚫|🚫|
|Mirror preview|✅|🚫|🚫|
|Zoom|✅|✅|✅|
|Take photo|✅|✅|✅|
|Flash|✅|✅|✅|
|Take video|✅|✅|✅|
|Record audio|✅|✅|✅|
|Video stabilization|✅|✅|✅|

## API Changes

Added:
- CameraView class
- CameraOptions enum
- CameraCaptureOptions enum
- CameraFlashMode enum

### Required permissions ###

**Android**
`AndroidManifest.xml` should include the following lines inside `manifest` tag
```XML
	<uses-permission android:name="android.permission.CAMERA" />
	<uses-permission android:name="android.permission.RECORD_AUDIO" />
	<uses-permission android:name="android.permission.FLASHLIGHT" />
```

**iOS**
`Info.plist` should include the following lines inside `dict` tag
```XML
	<key>NSCameraUsageDescription</key>
	<string>This app would like to access the camera</string>
	<key>NSMicrophoneUsageDescription</key>
	<string>This app would like to access the microphone to record movies</string>
```

**UWP**
`Package.appxmanifest` should include the following lines inside `Capabilities`
```XML
	<DeviceCapability Name="webcam" />
	<DeviceCapability Name="microphone" />
```

### Platforms Affected ### 

- Core/XAML (all platforms)
- iOS
- Android
- UWP

### Behavioral/Visual Changes ###

This is a new control

### Before/After Screenshots ### 

**iPhone**
![img_0067](https://user-images.githubusercontent.com/27482193/46955864-96e70c00-d09c-11e8-9f13-9b22d6015e11.PNG)

**Android**
[...]

PR #252, opened 2020-08-29
[UWP] CameraView

bmacombe:
### Description of Change ###

UWP Implementation of #211 
Follow Up to #244 

Add's UWP CameraView Renderer

### API Changes ###

Removes from CameraView 
```csharp
public bool KeepScreenOn { get; set; } // Bindable
public Aspect PreviewAspect {get; set; } // Bindable
```
KeepScreenOn was Not implemented on iOS and causes an exception in some cases in UWP.  Suggest using Xamarin Essentials for this as it's already implemented there.

PreviewAspect was only available on Android, we decided to remove until platform specifics are available again.

https://docs.microsoft.com/en-us/dotnet/api/xamarin.essentials.devicedisplay.keepscreenon?view=xamarin-essentials#Xamarin_Essentials_DeviceDisplay_KeepScreenOn

### PR Checklist ###

- [ ] Updated documentation


bmacombe:
@pictos @jfversluis After some rebase (slash giving up and just making a new branch :) ) fun....it's ready for review!

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
```diff
@@ -0,0 +1,397 @@
...
+				return;
+
+			IsBusy = true;
+			switch (Element.CaptureOptions)
+			{
+				default:
+				case CameraCaptureOptions.Default:
+				case CameraCaptureOptions.Photo:
+					if (mediaRecording != null)
+						goto case CameraCaptureOptions.Video;
```
pictos, line 109: `go to` LoL

bmacombe: Lol, yeah I noticed that too, I didn't even know you could do that!

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
```diff
@@ -0,0 +1,397 @@
...
+				await mediaCapture.InitializeAsync(new MediaCaptureInitializationSettings
+				{
+					VideoDeviceId = device.Id,
+					MediaCategory = MediaCategory.Media,
+					StreamingCaptureMode = selectedAudioDevice == null ? StreamingCaptureMode.Video : StreamingCaptureMode.AudioAndVideo,
+					AudioProcessing = Windows.Media.AudioProcessing.Default,
+					AudioDeviceId = selectedAudioDevice
+				});
+				flash = await Lamp.GetDefaultAsync();
+				if (mediaCapture.VideoDeviceController.ZoomControl.Supported)
```
pictos, line 331: Null reference exception if we click too fast in the `Frontal camera` switch
```suggestion
				if (mediaCapture?.VideoDeviceController.ZoomControl.Supported ?? false)
```

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
pictos: Me neither kkkk

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
bmacombe: Interestingly I get this exception if I toggle Front Camera to fast.

System.Runtime.InteropServices.COMException: 'The request is invalid in the current state.
The request is invalid in the current state.'

pictos: Sometimes I got this error as well `System.Runtime.InteropServices.COMException`

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
```diff
@@ -0,0 +1,397 @@
...
+			{
+				Element?.RaiseMediaCaptureFailed($"The app was denied access to the camera or microphone. {ex.Message}");
+				IsBusy = false;
+				return;
+			}
+
+			try
+			{
+				Control.Source = mediaCapture;
+				await mediaCapture.StartPreviewAsync();
```
bmacombe, line 346: This causes a com exception for me when switching the camera type to fast.  The try block doesn't catch it either

bmacombe: It's too late...that's because it's only it's not a file load exception

bmacombe: added a catch for the com exception which raises a capture failed message in CameraView.

pictos: Looks like this `FileLoadException` is a special one
https://docs.microsoft.com/en-us/windows/uwp/audio-video-camera/camera

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
```diff
@@ -0,0 +1,397 @@
...
+			if (isPreviewing)
+				await mediaCapture.StopPreviewAsync();
+
+			if (mediaRecording != null)
+				Element.RaiseMediaCaptured(new MediaCapturedEventArgs(video: await StopRecord()));
+
+			await Dispatcher.RunAsync(CoreDispatcherPriority.Normal, () =>
+			{
+				Control.Source = null;
+				displayRequest?.RequestRelease();
```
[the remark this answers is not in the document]
bmacombe: Decided to remove the keep screen on which required this call, so it's no longer an issue

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
```diff
@@ -0,0 +1,399 @@
...
+		{
+			get => busy;
+			set
+			{
+				if (busy != value)
+					Element.IsBusy = busy = value;
+			}
+		}
+
+		bool Available
```
pictos, line 57: I realized that we just set this property, but never use it to verify anything. Do you have a guess where we want to check if the camera is Available?

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
```diff
@@ -0,0 +1,399 @@
...
+		}
+
+		void UpdateZoom()
+		{
+			var zoomControl = mediaCapture.VideoDeviceController.ZoomControl;
+			if (!zoomControl.Supported)
+				return;
+
+			// Added here since it's an internal method to XF
+			static float Clamp(double value, float min, float max)
```
pictos, line 247: Can you move this to the end of the method?

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
```diff
@@ -0,0 +1,399 @@
...
+			{
+				case nameof(CameraView.CameraOptions):
+					await CleanupCameraAsync();
+					await InitializeCameraAsync();
+					break;
+				case nameof(CameraView.FlashMode):
+					if (flash != null)
+						flash.IsEnabled = Element.FlashMode == CameraFlashMode.Torch || Element.FlashMode == CameraFlashMode.On;
+					break;
+				case nameof(CameraView.PreviewAspect):
```
pictos, line 230: Do we will support this? I did a quick look and I didn't find it on iOS, but we have it on android.
cc: @jfversluis

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
```diff
@@ -0,0 +1,399 @@
...
+					UpdateZoom();
+					break;
+			}
+			base.OnElementPropertyChanged(sender, e);
+		}
+
+		void UpdateZoom()
+		{
+			var zoomControl = mediaCapture.VideoDeviceController.ZoomControl;
+			if (!zoomControl.Supported)
```
pictos, line 243: Just wondering if we should create a BindableProperty to expose this to the dev. To makes sense to have a boolean `IsZoomSupported`. What do you think? @bmacombe @jfversluis

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
bmacombe: I can see, but doesn't it need to be declared before it's used?  Could just make it a private method instead of a local function.  Just wanted to keep it close incase we can remove it if we get access to XF internals.

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
bmacombe: We can remove it if it's not used.  I'll double check that.

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
pictos: Good point. Well, I like to put my local functions at the end of the method... But we can move it later in another PR.

bmacombe: Guess you can declare them anywhere, I'll move it to the bottom.  Didn't know you could do, guess I've always assumed they had to be declared before you can use them.

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
bmacombe: I wonder if it at one point it was used like IsBusy, but it's not used now.  Removed

bmacombe: Actually that was used to set the IsAvailable on the CameraView so I added it back.

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
bmacombe: There is a MaxZoom property, which returns 1 when no zoom is available.  Could probably add that just in CameraView, true if MaxZoom >1 otherwise false.

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
pictos: I missed that, I'm sorry 🙈

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
jfversluis: Brrrr unfortunately I know you can... Can we get that out?

bmacombe: I'm sure it could be refactored to not use GoTo, I'll take a look

bmacombe: Replaced with a local function called inside the various case statements

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
jfversluis: As indicated [here](https://github.com/xamarin/XamarinCommunityToolkit/issues/211) (by the original author) indeed this is only for UWP. Strange that this is not a platform specific then while the rest is. I'll keep it for now, but might be worth to take it out and keep this is something platform specific to not confuse users.

XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj
```diff
@@ -51,7 +51,7 @@
   </PropertyGroup>
   <ItemGroup Condition=" '$(Configuration)'=='Debug' ">
     <!-- Manage Xamarin.Forms version for development -->
-    <PackageReference Include="Xamarin.Forms" Version="4.5.0.356" />
+    <PackageReference Include="Xamarin.Forms" Version="4.5.0.396" />
```
jfversluis, line 54: Do we actually need to upgrade this? Why?

XamarinCommunityToolkitSample.UWP/Xamarin.CommunityToolkit.Sample.UWP.csproj
```diff
@@ -143,7 +143,7 @@
     </Page>
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Xamarin.Forms" Version="4.6.0.1141" />
+    <PackageReference Include="Xamarin.Forms" Version="4.8.0.1269" />
```
jfversluis, line 146: Again, is this necessary? Although I don't mind using a newer version for the sample app, but can you then also upgrade iOS and Android?

jfversluis (changes requested):
Some very minor things. 

Could you maybe also get this one out while we're at it?
https://github.com/xamarin/XamarinCommunityToolkit/pull/252/files#r480900480

Maybe just comment it out for now with a TODO and turn it into a platform specific if we decide how to do that.

XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj
bmacombe: Not sure why that one was changed...I'll revert it

XamarinCommunityToolkitSample.UWP/Xamarin.CommunityToolkit.Sample.UWP.csproj
bmacombe: Ios and Android were already bumped...check main.

I had to bump uwp to match because the sample app used the new online grid row/column Def and the converter isn't in 4.6

jfversluis: In that case, no problem!

bmacombe: it's those dang Mac users, not updating UWP 😄

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
bmacombe: I commented it out and left a note that it's only supported by Android and we can revisit when we have the platform specifics.  In the original spec, there is also MirrorPreview that is only android, but I didn't see any code in CameraView.

@pictos You might want to look into the MirrorPreview on the Android renderer to see if it's needed https://github.com/xamarin/Xamarin.Forms/blob/ae045fe259a23984ea5790b6551fc1e3fb3f5159/Xamarin.Forms.Platform.Android/Renderers/CameraViewRenderer.cs#L74-L76

bmacombe:
@jfversluis @pictos Hopefull all the final tweaks are made!

XamarinCommunityToolkitSample.UWP/Xamarin.CommunityToolkit.Sample.UWP.csproj
jfversluis: Guilty...

jfversluis (approved):
LGTM

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.uwp.cs
pictos: @bmacombe, yes on android this method is used with PlatformSpecific, I did a spike and you find it [here](https://github.com/jfversluis/XamarinCommunityToolkit-1/commit/faf782d8d28193c3f3764505022fda1ce9ffde3c). I'm just awaiting the feedback from the team to see how we will handle it. But if until my PR we don't have a decision we can remove it and add later as enchantment.

bmacombe: @pictos Cool sounds good!

jfversluis: @pictos if you're comfortable bring up that spike link in the forms channel on Discord. If you'd rather have me do it, let me know :)

pictos: @jfversluis I can do it (:

Issue #223, opened 2020-08-24
[Enhancement] UserStoppedTypingBehavior


jBijsterboschNL:
## Summary

- After a certain threshold has been reached after the user has stopped typing, a given command can be executed
- Stopped Typing Threshold will be adjustable
- Auto Dismiss Keyboard will be adjustable when user has stopped typing

See this repository for an example: https://github.com/jBijsterboschNL/Xamarin.Forms-UserStoppedTypingBehavior

And this blog post about the implementation:
https://blog.johndotnet.nl/automatic-search-trigger-when-user-stopped-typing-in-xamarin-forms/

## Intended Use Case
If you want a command (f.e. a search command) to be automatically executed after the user has stopped typing, but you don't want it to be triggered on each text changed event (triggering too many and also obsolete search calls).

jfversluis:
I am Gerald and I approve of this idea!

sthewissen:
Sounds cool!

jBijsterboschNL:
Great! :) will try to get this implemented into the Xamarin Community Toolkit in the next weeks.

jBijsterboschNL:
PR has been merged, so this issue can be closed :) thanks all for your feedback

AndreiMisiukevich:
Closed by #253

PR #253, opened 2020-08-30
User stopped typing behavior

jBijsterboschNL:
### Description of Change ###

Implemented UserStoppedTypingBehavior. This behavior waits for the user to stop typing and then executes a Command. The waiting threshold is adjustable, as well as there is an option to dismiss the keyboard after the user has stopped typing.

### Bugs Fixed ###

- Related to issue #223

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

- [ ] Updated documentation


XamarinCommunityToolkit/Behaviors/UserStoppedTypingBehavior.shared.cs
```diff
@@ -0,0 +1,81 @@
...
+        #region Bindable Properties
+
+        public static readonly BindableProperty CommandProperty = BindableProperty.Create(nameof(Command), typeof(ICommand), typeof(UserStoppedTypingBehavior));
+        public ICommand Command
+        {
+            get => (ICommand)GetValue(CommandProperty);
+            set => SetValue(CommandProperty, value);
+        }
+
+        public static readonly BindableProperty StoppedTypingThresholdProperty = BindableProperty.Create(nameof(StoppedTypingThreshold), typeof(int), typeof(UserStoppedTypingBehavior), defaultValue: 1000);
```
AndreiMisiukevich, line 21: no need to write ```defaultValue: 1000``` just ```1000``` )

XamarinCommunityToolkit/Behaviors/UserStoppedTypingBehavior.shared.cs
```diff
@@ -0,0 +1,81 @@
...
+        public int StoppedTypingThreshold
+        {
+            get => (int)GetValue(StoppedTypingThresholdProperty);
+            set => SetValue(StoppedTypingThresholdProperty, value);
+        }
+
+        public static readonly BindableProperty AutoDismissKeyboardProperty = BindableProperty.Create(nameof(AutoDismissKeyboard), typeof(bool), typeof(UserStoppedTypingBehavior), defaultValue: false);
```
AndreiMisiukevich, line 28: same here

XamarinCommunityToolkit/Behaviors/UserStoppedTypingBehavior.shared.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        public static readonly BindableProperty CommandProperty = BindableProperty.Create(nameof(Command), typeof(ICommand), typeof(UserStoppedTypingBehavior));
+        public ICommand Command
+        {
+            get => (ICommand)GetValue(CommandProperty);
+            set => SetValue(CommandProperty, value);
+        }
+
+        public static readonly BindableProperty StoppedTypingThresholdProperty = BindableProperty.Create(nameof(StoppedTypingThreshold), typeof(int), typeof(UserStoppedTypingBehavior), defaultValue: 1000);
+        public int StoppedTypingThreshold
```
AndreiMisiukevich, line 22: I think it makes sense to clarify what this means.
StoppedTypingTimeThreshold or StoppedTypingMillisecondsThreshold

XamarinCommunityToolkit/Behaviors/UserStoppedTypingBehavior.shared.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            get => (int)GetValue(StoppedTypingThresholdProperty);
+            set => SetValue(StoppedTypingThresholdProperty, value);
+        }
+
+        public static readonly BindableProperty AutoDismissKeyboardProperty = BindableProperty.Create(nameof(AutoDismissKeyboard), typeof(bool), typeof(UserStoppedTypingBehavior), defaultValue: false);
+        public bool AutoDismissKeyboard
```
AndreiMisiukevich, line 29: @jfversluis @sthewissen @Char0394 
Perhaps it's only my personal preference, but I prefer to use "Is", "Are", "Should" as infix for all boolean properties.

e.g. ShouldDismissKeyboardAutomatically or IsKeyboardAutoDismissing

XamarinCommunityToolkit/Behaviors/UserStoppedTypingBehavior.shared.cs
```diff
@@ -0,0 +1,81 @@
...
+        protected override void OnDetachingFrom(View view)
+        {
+            base.OnDetachingFrom(view);
+
+			if (!(view is InputView inputView)) return;
+			
+			inputView.TextChanged -= InputView_TextChanged;
+        }
+
+        void InputView_TextChanged(object sender, TextChangedEventArgs e)
```
AndreiMisiukevich, line 55: OnInputViewTextChanged

But I propose you to override existing OnViewPropertyChanged. There you will be able to cast "View" and retrieve necessary property

XamarinCommunityToolkit/Behaviors/UserStoppedTypingBehavior.shared.cs
```diff
@@ -0,0 +1,81 @@
...
+            tokenSource = new CancellationTokenSource();
+
+#pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
+			PerformTextChanged(sender as Entry, e.NewTextValue, tokenSource.Token);
+#pragma warning restore CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
+		}
+
+        async Task PerformTextChanged(Entry entry, string newTextValue, CancellationToken token)
+        {
+            await Task.Delay(StoppedTypingThreshold);
```
AndreiMisiukevich, line 69: Just for better readability during the review please update the formatting in this file :)

XamarinCommunityToolkit/Behaviors/UserStoppedTypingBehavior.shared.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        void InputView_TextChanged(object sender, TextChangedEventArgs e)
+        {
+            if (tokenSource != null)
+                tokenSource.Cancel();
+            
+            tokenSource = new CancellationTokenSource();
+
+#pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
+			PerformTextChanged(sender as Entry, e.NewTextValue, tokenSource.Token);
```
AndreiMisiukevich, line 63: I think we needn't have to pass tokenSource.Token here. We can get token inside the PerformTextChanged's body

XamarinCommunityToolkit/Behaviors/UserStoppedTypingBehavior.shared.cs
```diff
@@ -0,0 +1,81 @@
...
+#pragma warning restore CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
+		}
+
+        async Task PerformTextChanged(Entry entry, string newTextValue, CancellationToken token)
+        {
+            await Task.Delay(StoppedTypingThreshold);
+
+            if (token.IsCancellationRequested) return;
```
AndreiMisiukevich, line 71: How about re-writing this method to 
```
await Task.Delay(StoppedTypingThreshold, cancellationToken.Token)
.ContinueWith(task => { 
    if (task.Exception != null) return;
    /// Proceed
});
```
Or even try-catch all method (but anyway I suggest to pass token as a parameter to Task.Delay method).

AndreiMisiukevich (changes requested):
Hi, thank you for this awesome contribution. I appreciate any desire to make XF World better and more convenient. Especially when the contribution is covered by unit tests and has a sample 👍

XamarinCommunityToolkitSample.sln
```diff
@@ -9,11 +9,11 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Xamarin.CommunityToolkit.Sa
 EndProject
```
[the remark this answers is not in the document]
jBijsterboschNL: Not sure where these changes came from. After merging in latest version from main, this got changed as well. When I undo this, and open the solution again, Visual Studio for Windows automatically does the same changes again. Not sure if it does any harm? Also not sure how I could prevent this?

XamarinCommunityToolkit/Behaviors/UserStoppedTypingBehavior.shared.cs
jBijsterboschNL: Hi @AndreiMisiukevich, but why would you want to override the OnViewPropertyChanged for this, as the InputView already has a specific event (TextChanged) for this? Otherwise the OnViewPropertyChanged code will be called for all property changes, not only for the Text property to change, and we need to add a check if the Text property has changed. I don't see the benefits of it and why that would be a better solution. The naming change has been made (once I commit my changes).

XamarinCommunityToolkit/Behaviors/UserStoppedTypingBehavior.shared.cs
jBijsterboschNL: Hi @AndreiMisiukevich, this is definitely needed like so. As otherwise the token cancellation won't work properly no more. After applying this change requested by you, one of the unit tests failed (thanks for writing unit tests, yay!).

jBijsterboschNL:
Hi @AndreiMisiukevich, thank you for your quick code review action. I've applied changes that you requested and replied to some change request to discuss further.

XamarinCommunityToolkit.UnitTests/Behaviors/UserStoppedTypingBehavior_Tests.cs
```diff
@@ -0,0 +1,93 @@
...
+using Xunit;
+
+namespace Xamarin.CommunityToolkit.UnitTests.Behaviors
+{
+	public class UserStoppedTypingBehavior_Tests
+	{
+		const int defaultThreshold = 1000;
+
+		[Fact]
+		public async void CommandExecutesAfterTimeThresholdExpires()
```
pictos, line 14: Any reason to don't use `async Task` here and in the others methods?

XamarinCommunityToolkit/Behaviors/UserStoppedTypingBehavior.shared.cs
```diff
@@ -0,0 +1,81 @@
...
+
+		void InputView_TextChanged(object sender, TextChangedEventArgs e)
+		{
+			if (tokenSource != null)
+				tokenSource.Cancel();
+
+			tokenSource = new CancellationTokenSource();
+
+#pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
+			PerformTextChanged(sender as Entry, e.NewTextValue, tokenSource.Token);
```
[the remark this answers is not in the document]
pictos: Also, instead of using `sender as Entry` can you use `(Entry)sender`. Because if the View isn't an `Entry` we can face a null exception at `entry.Unfocu() //line 74`. So I think it's better to fail here because of the sync/async behavior. Do you agree?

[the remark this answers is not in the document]
jBijsterboschNL: Hi @pictos, actually the type check has already been done in the OnAttached() method. So that is not possible. But I did see that I was casting to an Entry instead of an InputView, so will change that here.

jBijsterboschNL:
@AndreiMisiukevich and @pictos, I've applied all requested changed from your code reviews. Could you guys re-check and approve this PR once you guys agree with the code changes that have been made? Cheers.

XamarinCommunityToolkit/Behaviors/UserStoppedTypingBehavior.shared.cs
AndreiMisiukevich: in that case, you won't have to take care of unsubscribing from your event.

```csharp
protected override void OnViewPropertyChanged(object sender, PropertyChangedEventArgs e) {
    if (e.PropertyName == InputView.TextProperty.PropertyName) {
        var text = ((InputView)View).Text;
    }
}
```

but uh

XamarinCommunityToolkit/Behaviors/UserStoppedTypingBehavior.shared.cs
```diff
@@ -0,0 +1,82 @@
...
+		protected override void OnDetachingFrom(View view)
+		{
+			base.OnDetachingFrom(view);
+
+			if (!(view is InputView inputView)) return;
+
+			inputView.TextChanged -= InputView_TextChanged;
+		}
+
+		void InputView_TextChanged(object sender, TextChangedEventArgs e)
```
AndreiMisiukevich, line 55: please rename it to OnInputTextChanged

XamarinCommunityToolkitSample/ViewModels/Behaviors/UserStoppedTypingBehaviorViewModel.cs
```diff
@@ -0,0 +1,33 @@
...
+			set => Set(ref performedSearches, value);
+		}
+
+		public ICommand SearchCommand { get; }
+
+		#endregion Properties
+
+		public UserStoppedTypingBehaviorViewModel()
+		{
+			SearchCommand = new Command<string>(PerformSearch);
```
AndreiMisiukevich, line 25: how about

```=> SearchCommand = new Command<string>(PerformSearch);```
For this ctor and for PerformSearch as well

sthewissen (approved):
Awesome! Let's merge this!

AndreiMisiukevich (approved):
LGTM
Good work with comments :) 👍

jfversluis:
I've updated the PR template to automatically also resolve the issue when the PR is merged. That is typically what we want.

Great work @jBijsterboschNL !

PR #255, opened 2020-08-30
Enable C# 8 Null check 

### LINKED ISSUE #254: [Enhancement] Enable Nullable check for the project
## Summary

Since we are using the coolest new C# 8 would be good to enable the NullableReference warnings, to make the lib safer.

## API Changes

Add <Nullable>enable</Nullable> into the csproj.

## Intended Use Case
Avoid `NullReferenceExceptions`

pictos:
### Description of Change ###

I Added the ` <Nullable>enable</Nullable>` into the csproj and added the `?` and `!` validators. 

For now, I don't add this to the `renderers`.

### Bugs Fixed ###

- fixes #254

### Behavioral Changes ###

None

### PR Checklist ###

- [ ] Updated documentation


AndreiMisiukevich:
@pictos thanks for your contribution :) 
Swift is coming 😆 

@jfversluis do we actually want it?

jfversluis:
@AndreiMisiukevich personally I think this is a good move and makes the quality of our code better. Do you have a different view?

AndreiMisiukevich:
> @AndreiMisiukevich personally I think this is a good move and makes the quality of our code better. Do you have a different view?

No, but then we should restrict force unwrapping by "!"

pictos:
> No, but then we should restrict force unwrapping by "!"

Yeah, of course. I'm using just in places that Roslyn can't know that the `object` will not be null. For example
```csharp
if (string.IsNullOrEmpty(myString)
{
   return;
}
var p = myString.Trim();
```

In this case, we know that `myString` isn't null, but Roslyn doesn't, so we need to put the `!`.

Also, I skip renderers because of it, because are a lot of places that `object` isn't null but Roslyn doesn't know, for example

```csharp
if (Control == null)
{
     SetNativeControl(new MyControl());
     Control.BackgroundColor = Color.Red.ToMyPlatform(); 
}
```
There, we know that the Control will not be null after `SetNativeControl` call, but Roslyn doesn't.

pictos:
I'm closing this one because this branch is too behind the current `main`. In the future, we can open this in another PR

PR #269, opened 2020-09-05
Added BaseTemplatedView

### LINKED ISSUE #258: [Housekeeping] Add BaseTemplatedView
## Summary
In order to get rid of code duplications, I propose to add a BaseTemplatedView.
All views witch are inherited from TempateView, they will be inherited from BaseTemplatedView instead.

## API Changes

```csharp
	public class BaseTemplatedView<T> : TemplatedView where T : View, new()
	{
		T Content { get; set; }

		public BaseTemplatedView()
			=> ControlTemplate = new ControlTemplate(typeof(T));

		protected override void OnBindingContextChanged()
		{
			base.OnBindingContextChanged();
			Content.BindingContext = BindingContext;
		}

		protected override void OnChildAdded(Element child)
		{
			if (Content == null && child is T content)
			{
				Content = content;
				OnContentAdded(Content);
			}

			base.OnChildAdded(child);
		}

		protected override void OnChildRemoved(Element child)
		{
			if (child == Content)
			{
				OnContentDestroyed(Content);
				Content = null;
			}
			base.OnChildRemoved(child);
		}

		protected virtual void OnContentAdded(T content)
		{
		}

		protected virtual void OnContentDestroyed(T content)
		{
		}
	}
```

AndreiMisiukevich:
### Description of Change ###

Added hidden abstract BaseTemplatedView class that combines ControlTemplate/BindingContext setting logic inside.
It would keep more common properties/methods in the future.

Also, removed useless "ChildRemoved" overrides in templated views because we choose ControlTemplate only once and don't change it.

### How to test
Run the sample app and make sure all templated views work as expected.

### Bugs Fixed ###
- Fixes #258 

### API Changes ###
None
 
### Behavioral Changes ###
None

### PR Checklist ###

- [ ] Has tests 
- [X] Has samples
- [ ] Updated documentation


XamarinCommunityToolkit/Views/BaseTemplatedView.shared.cs
```diff
@@ -0,0 +1,33 @@
+using System.ComponentModel;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.UI.Views
+{
+	[EditorBrowsable(EditorBrowsableState.Never)]
```
pictos, line 6: Any reason to hide this for the lib users? I see this as a great helper if any dev. needs to create a templated view control

AndreiMisiukevich: I don't want that this control appears as a hint in XAML. 
Also, usually, users inherit custom controls directly from "TControl" class

e.g.
```
class Expander : StackLayout {
}
```

pictos: I see, so can we make this `internal` instead of using the `EditorBrowsableState.Never`?

AndreiMisiukevich: No, we can't inherit **public** class from **internal**.

pictos: Hm... That makes sense and this also explains why we have this attribute in other places, tks.

AndreiMisiukevich:
CC @jfversluis and this one please)

jfversluis (approved):

GreenTea66:
Ich brauche auch ein solches Element. In WPF fand ich das immer eine tolle Sache. Jetzt arbeite ich an einem grossen Xamarin.Forms Projekt und vermisse dieses Steuerelement !!!

jfversluis:
@JKDeveloper92 I think you might get get more/better answers when you ask in English if possible :)

Issue #287, opened 2020-09-10
[Enhancement] [CameraView] MediaCapturedEventArgs should return file path of saved file when saving file


acuntex:
## Summary
When taking a photo or recording a video, the CameraView currently raises the event MediaCaptured with a parameter of type MediaCapturedEventArgs.
This class contains the following properties:
```
		/// <summary>
		/// Raw image data, only filled when taking a picture
		/// </summary>
		public object Data { get; }
		public ImageSource Image { get; }
		public MediaSource Video { get; }
```
Data gets filled with the content of a byte array (the jpeg) whether SavePhotoToFile is true or false.
ImageSource gets filled with the content of a byte array whether SavePhotoToFile is true or false. 
MediaSource gets filled with `MediaSource.FromFile(outputFileUrl.Path)` yet I believe the file does not exist anymore at this path because a few lines above it was moved:
```
						creationRequest.AddResource(
							PHAssetResourceType.Video,
							outputFileUrl,
							new PHAssetResourceCreationOptions
							{
								ShouldMoveFile = true
							});
```

The problem now is that a developer could need the path of the saved file for further processing. e.g. I want to pass it to TOCropViewController (iOS) or ArthurHub/Android-Image-Cropper (Android).
While the iOS component could work with an in-memory-solution, the Android component does not.

Furthermore I'm then questioning the need of the ImageSource and the MediaSource since it's just a waste of memory/cpu if they aren't used.

## API Changes
I'm proposing the following change for the class `MediaCapturedEventArgs`:

```
		/// <summary>
		/// Raw image data, only filled when taking a picture and SavePhotoToFile is false
		/// </summary>
		public byte[] ImageData { get; }
		/// <summary>
		/// Path of the saved file, only filled when taking a picture and SavePhotoToFile is true
		/// </summary>
		public string Path { get; }
```
Is there anything else necessary besides these two properties?

## Intended Use Case
Whenever the developer needs the file path of the saved file.

PR #291, opened 2020-09-11
CameraView: Implemented returning of path of saved file

acuntex:
### Description of Change ###

As stated in https://github.com/xamarin/XamarinCommunityToolkit/issues/287 it is sometimes necessary to access the path of the saved file.
Currently it's also impossible to access the video file because it has been moved and the path given to the MediaSource is invalid.
Furthermore I added a lazy creation of ImageSource/MediaSource since they are not always needed.

### API Changes ###

Added in MediaCapturedEventArgs: 
 
- `string Path { get; }`

Changed in MediaCapturedEventArgs:

 - `object Data { get; }` => `byte[] ImageData { get; }`
 - `ImageSource Image` is now lazily created
 - `MediaSource Video` is now lazily created
 - `internal MediaCapturedEventArgs(object data = null, ImageSource image = null, MediaSource video = null)` => `internal MediaCapturedEventArgs(string path = null, byte[] imageData = null)`

### Behavioral Changes ###

 - ImageSource and MediaSource in MediaCapturedEventArgs are now lazily created and not provided in the constructor
 - MediaCapturedEventArgs returns now the path of the saved file for videos and for photos if SavePhotoToFile is true

### PR Checklist ###

- [ ] Updated documentation

### DIFF
--- a/XamarinCommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs
+++ b/XamarinCommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs
@@ -303,13 +303,13 @@ namespace Xamarin.CommunityToolkit.UI.Views
 			}
 		}
 
-		void OnPhoto(object sender, byte[] data) =>
+		void OnPhoto(object sender, Tuple<string, byte[]> tuple) =>
 			Device.BeginInvokeOnMainThread(() =>
-				Element?.RaiseMediaCaptured(new MediaCapturedEventArgs(data, ImageSource.FromStream(() => new MemoryStream(data)))));
+				Element?.RaiseMediaCaptured(new MediaCapturedEventArgs(tuple.Item1, tuple.Item2)));
 
-		void OnVideo(object sender, string data) =>
+		void OnVideo(object sender, string path) =>
 			Device.BeginInvokeOnMainThread(() =>
-				Element?.RaiseMediaCaptured(new MediaCapturedEventArgs(video: MediaSource.FromFile(data))));
+				Element?.RaiseMediaCaptured(new MediaCapturedEventArgs(path)));
 
 		void SetupImageReader()
 		{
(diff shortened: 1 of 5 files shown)


XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.ios.cs
```diff
@@ -158,7 +198,8 @@ async void HandleShutter(object sender, EventArgs e)
 			{
 				case CameraCaptureOptions.Default:
 				case CameraCaptureOptions.Photo:
-					await Control?.TakePhoto();
+					if (Control != null)
+						await Control.TakePhoto();
```
jfversluis, line 202: Is there any particular reason to do this?

jfversluis (approved):
One nitty thing. All the rest looks good to me :)

XamarinCommunityToolkit/Views/CameraView/CameraViewRenderer.ios.cs
acuntex: Previously it was
`await Control?.TakePhoto();`
If Control is null, it would throw an exception because you can't await null, only Tasks.

AndreiMisiukevich: BTW should we make all awaits "configure await - FALSE" ?

acuntex: I would not say all... Only if the rest of the method should not be run on the captured thread (e.g. the UI thread).

jfversluis: Oh interesting! I would say it knows what to do with that null. Thanks!

Yeah, the ConfigureAwait is something we should probably more aware of as well.

jfversluis:
Could you rebase/fix the conflicts @acuntex ? I think after that we're good to go

acuntex:
@jfversluis Rebased/Fixed and implemented the android part.
Also added the replacement for the <None ...> in the csproj.

jfversluis (approved):

Issue #323, opened 2020-09-23
[Bug] Safe area effect restoring issue


Tommigun1980:
Hi and thanks for a great project.

When I was browsing the code I noticed the following in https://github.com/xamarin/XamarinCommunityToolkit/blob/main/XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs - when the effect is detached it assigns a margin that in some cases has not been assigned (old iOS versions etc), so you might want to keep track of whether the initial margin edit actually happened.

Thanks.

PR #324, opened 2020-09-23
Don't overwrite margin when detaching safe area effect if the effect

Tommigun1980:
was never applied.

### Bugs Fixed ###

- Fixes https://github.com/xamarin/XamarinCommunityToolkit/issues/323

### API Changes ###

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Browsing the [sample app](./XamarinCommunityToolkitSample) is the best place to 
 
 The toolkit is available via NuGet, and should be installed into all your projects:
 
-* NuGet Official Releases: [![NuGet](https://img.shields.io/nuget/vpre/Xamarin.CommunityToolkit.svg?label=NuGet)](https://www.nuget.org/packages/Microsoft.Toolkit.Xamarin.Forms)
+* NuGet Official Releases: [![NuGet](https://img.shields.io/nuget/vpre/Xamarin.CommunityToolkit.svg?label=NuGet)](https://www.nuget.org/packages/Xamarin.CommunityToolkit)
 * NuGet Nightly Releases: [![NuGet Nightly](https://img.shields.io/badge/NuGet-Nightly-yellow)](https://pkgs.dev.azure.com/xamarin/public/_packaging/XamarinCommunityToolkitNightly/nuget/v3/index.json)
 
 Now all you need to do is use it! 
--- a/XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs
+++ b/XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs
@@ -12,7 +12,7 @@ namespace Xamarin.CommunityToolkit.iOS.Effects
 {
 	public class SafeAreaEffectRouter : PlatformEffect
 	{
-		Thickness initialMargin;
+		Thickness? initialMargin;
 
 		protected override void OnAttached()
 		{
@@ -24,24 +24,27 @@ namespace Xamarin.CommunityToolkit.iOS.Effects
 
 			var insets = UIApplication.SharedApplication.Windows[0].SafeAreaInsets;
 
-			initialMargin = element.Margin;
-
 			if (insets.Top <= 0)
 				return;
+
+			initialMargin = element.Margin;
 
 			var safeArea = SafeAreaEffect.GetSafeArea(element);
 
 			element.Margin = new Thickness(
-				initialMargin.Left + (safeArea.Left ? insets.Left : 0),
-				initialMargin.Top + (safeArea.Top ? insets.Top : 0),
-				initialMargin.Right + (safeArea.Right ? insets.Right : 0),
-				initialMargin.Bottom + (safeArea.Bottom ? insets.Bottom : 0));
+				element.Margin.Left + (safeArea.Left ? insets.Left : 0),
+				element.Margin.Top + (safeArea.Top ? insets.Top : 0),
+				element.Margin.Right + (safeArea.Right ? insets.Right : 0),
+				element.Margin.Bottom + (safeArea.Bottom ? insets.Bottom : 0));
 		}
 
 		protected override void OnDetached()
 		{
-			if (Element is Layout element)
-				element.Margin = initialMargin;
+			if (Element is Layout element && initialMargin.HasValue)
+			{
+				element.Margin = initialMargin.Value;
+				initialMargin = null;
+			}
 		}
 	}
 }


README.md
```diff
@@ -18,7 +18,7 @@ Browsing the [sample app](./XamarinCommunityToolkitSample) is the best place to
 
 The toolkit is available via NuGet, and should be installed into all your projects:
 
-* NuGet Official Releases: [![NuGet](https://img.shields.io/nuget/vpre/Xamarin.CommunityToolkit.svg?label=NuGet)](https://www.nuget.org/packages/Microsoft.Toolkit.Xamarin.Forms)
+* NuGet Official Releases: [![NuGet](https://img.shields.io/nuget/vpre/Xamarin.CommunityToolkit.svg?label=NuGet)](https://www.nuget.org/packages/Xamarin.CommunityToolkit)
```
AndreiMisiukevich, line 21: Should these lines be discarded now?

XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs
```diff
@@ -24,24 +24,27 @@ protected override void OnAttached()
...
 			var insets = UIApplication.SharedApplication.Windows[0].SafeAreaInsets;
 
-			initialMargin = element.Margin;
-
 			if (insets.Top <= 0)
 				return;
 
+			initialMargin = element.Margin;
+
 			var safeArea = SafeAreaEffect.GetSafeArea(element);
```
AndreiMisiukevich, line 33: How about adding "isApplied" boolean filed instead.
We can set it to true in this line and then check if it's true in the **OnDeatached** method

XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs
```diff
@@ -24,24 +24,27 @@ protected override void OnAttached()
...
-				initialMargin.Bottom + (safeArea.Bottom ? insets.Bottom : 0));
+				element.Margin.Left + (safeArea.Left ? insets.Left : 0),
+				element.Margin.Top + (safeArea.Top ? insets.Top : 0),
+				element.Margin.Right + (safeArea.Right ? insets.Right : 0),
+				element.Margin.Bottom + (safeArea.Bottom ? insets.Bottom : 0));
 		}
 
 		protected override void OnDetached()
 		{
-			if (Element is Layout element)
```
AndreiMisiukevich, line 43:
```csharp
if (isApplied)
    element.Margin = initialMargin;

isApplied = false;
```

XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs
Tommigun1980: You could. But the entire point with conditional value types is that you don’t need two values for this.

Tommigun1980: You could. But the entire reason for billable value types existing is exactly to avoid having two variables to accomplish this. Why not use all language features?

XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs
Tommigun1980: Why? Nullable value types exist exactly for avoiding that extra state flag.

README.md
Tommigun1980: Yeah. I closed this PR as another was created fixing this. The other one doesn’t have this line anymore.

Tommigun1980: Yeah. Sorry about that, another pull request fixed this so I closed the pull request with this commit but my branch still had this apparently.

pictos:
@Tommigun1980 can you resolve the conflicts?

XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs
```diff
@@ -24,24 +24,27 @@ protected override void OnAttached()
...
+			initialMargin = element.Margin;
+
 			var safeArea = SafeAreaEffect.GetSafeArea(element);
 
 			element.Margin = new Thickness(
-				initialMargin.Left + (safeArea.Left ? insets.Left : 0),
-				initialMargin.Top + (safeArea.Top ? insets.Top : 0),
-				initialMargin.Right + (safeArea.Right ? insets.Right : 0),
-				initialMargin.Bottom + (safeArea.Bottom ? insets.Bottom : 0));
+				element.Margin.Left + (safeArea.Left ? insets.Left : 0),
```
pictos, line 35: Why not use the `initialMargin`? Here we are calling two get methods per call, if we use the `initialMargin` will be just one call.

Tommigun1980: @pictos
There would be the same number of “calls” with initialMargin.Value.Xxx. And even if there was one more “call”, the compiler is pretty good at taking care of micro optimizations like that, not to mention that I’d gladly trade like one cpu cycle for legibility (as in store initial value, then modify value). But in this case the cost is even the same. If you really want to go down to the nitty gritty:

element.Margin.Left vs initialValue.Value.Left should be equivalent
There is no method call involved in any of these, just a property fetcher, that surely will be inlined anyway
I think discussing individual cycles takes away time from more important matters, especially if it’s at the cost of legibility.

pictos: @Tommigun1980
> element.Margin.Left vs initialValue.Value.Left should be equivalent

Why `initialMargin` is nullable? If I'm not wrong the way that XF works, we can guarantee that `initialMargin` wouldn't be null.

> There is no method call involved in any of these, just a property fetcher, that surely will be inlined anyway

I did a test on my side and this isn't true. Using a local variable to get the `Margin` and access the values will be away faster than using your approach. I ran against .NETCore, Mono should follow the same results but with different values.

``` ini

BenchmarkDotNet=v0.12.1.1356-nightly, OS=Windows 10.0.19041.508 (2004/?/20H1)
Intel Core i5-7300HQ CPU 2.50GHz (Kaby Lake), 1 CPU, 4 logical and 4 physical cores
  [Host]     : .NET Framework 4.8 (4.8.4220.0), X64 RyuJIT
  DefaultJob : .NET Framework 4.8 (4.8.4220.0), X64 RyuJIT


```
|                     Method |     Mean |   Error |  StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------- |---------:|--------:|--------:|------:|--------:|-------:|------:|------:|----------:|
| UsingLocalToAccessProperty | 205.6 ns | 4.13 ns | 7.24 ns |  1.00 |    0.00 | 0.0153 |     - |     - |      48 B |
|     AccessPropertyDirectly | 329.8 ns | 6.42 ns | 6.00 ns |  1.65 |    0.07 | 0.0153 |     - |     - |      48 B |

You can run by yourself [here](https://github.com/pictos/XamarinCommunityToolkit/tree/benchmark). If I did something wrong in my preparation please let me know.

>  if it’s at the cost of legibility.
I'm sorry but I can't see the cost of legibility to using a local variable instead of using a property directly.

README.md
AndreiMisiukevich: Can you please fix the merge conflict? (You can remove this commit or just merge the latest "main" code)

XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs
AndreiMisiukevich: Every time you call getter of "Element.Margin" the unboxing happens (because Thickness is a struct, but all bindable properties stores "object")

XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs
AndreiMisiukevich: Well, we can leave nullable, but I don't think we should check "Element is Layout"
Just leave "if initialMargin.HasValue"

XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs
Tommigun1980: > @Tommigun1980
> 
> > element.Margin.Left vs initialValue.Value.Left should be equivalent
> 
> Why `initialMargin` is nullable? If I'm not wrong the way that XF works, we can guarantee that `initialMargin` wouldn't be null when we need to use it. And `Thickness` is a `struct`.
> 
> > There is no method call involved in any of these, just a property fetcher, that surely will be inlined anyway
> 
> I did a test on my side and this isn't true. Using a local variable to get the `Margin` and access the values will be away faster than using your approach. I ran against .NETFramework, Mono should follow the same results but with different values.
> 
> ```ini
> BenchmarkDotNet=v0.12.1.1356-nightly, OS=Windows 10.0.19041.508 (2004/?/20H1)
> Intel Core i5-7300HQ CPU 2.50GHz (Kaby Lake), 1 CPU, 4 logical and 4 physical cores
>   [Host]     : .NET Framework 4.8 (4.8.4220.0), X64 RyuJIT
>   DefaultJob : .NET Framework 4.8 (4.8.4220.0), X64 RyuJIT
> ```
> 
> Method	Mean	Error	StdDev	Ratio	RatioSD	Gen 0	Gen 1	Gen 2	Allocated
> UsingLocalToAccessProperty	205.6 ns	4.13 ns	7.24 ns	1.00	0.00	0.0153	-	-	48 B
> AccessPropertyDirectly	329.8 ns	6.42 ns	6.00 ns	1.65	0.07	0.0153	-	-	48 B
> You can run by yourself [here](https://github.com/pictos/XamarinCommunityToolkit/tree/benchmark). If I did something wrong in my preparation please let me know.
> 
> > if it’s at the cost of legibility.
> 
> I'm sorry but I can't see the cost of legibility to using a local variable instead of using a property directly.

Why it’s nullable? What do you mean, the FIX in this PR is to make it nullable. It’s not set in all cases when disabled (old iOS etc).

Also I’m closing this, REALLY don’t have time for these kinds of discussions.

Bye.

pictos: @Tommigun1980, I'm sorry if I sounded unfriendly, English isn't my main language. I'm doing a code review and I don't have all context even reading the PR description and issue. Said that let me rephrase my concerns.

We're building here a lib that will be used for many developers in many applications, which different contexts, so performance is important to preserve especially if it's an easy win like this.

Making a struct nullable implies adding checks to make sure that we don't face a null exception, this increases complexity to the code, so if we can avoid that is better.

And maybe I'm wrong and I'm ok with that, it's always a chance to learning something new. And please remember this is an open-source project so this kind of discussion can happen in all PRs.

AndreiMisiukevich:
@Tommigun1980 Sorry, I didn't want to seem so picky. I just wanted to clarify some unclear points and probably make something better. I don't insist on my point of view or push my ideas as a 100% proper approach and always open to conversation.

I appreciate your contribution and I will be glad to continue our conversation if you wish :)

jfversluis:
Hey @Tommigun1980 I realize I'm a bit late to the party here, but I still wanted to reach out and reiterate what has been said by the other friends in here. Sorry that you felt that this was being micro-managed or however it came across. I'm sure that wasn't the goal and it certainly wasn't the goal to agitate you in any way, if that is what happened.

Just wanted to say that your efforts are certainly appreciated and I hope you're willing to reconsider and stick with us. If you don't have the time or willingness to see it through, that is totally fine, just let us know and we'll take it from there. Knowing the people involved with this issue I know that they are not native English speakers (just like me btw :D) and are just passionate about building high performing code.

I am in no way saying that you are not that passionate and don't care about any of it, but I think this might have been the cause of a misunderstanding with all of this. At the very least I wanted you to know we appreciate what you did and let you know there is no hard feelings. I hope you'll find this Toolkit useful! Have a nice day!

PR #334, opened 2020-09-28
Added Commands to CameraView control

### LINKED ISSUE #285: [Enhancement] [CameraView] Commands for Events (MVVM)
## Summary
In order to use the CameraView in am MVVM architecture with a view model it would be helpful to have commands for the existing events.

## API Changes

The following commands would be necessary to have a clean MVVM architecture:

```
		public static readonly BindableProperty MediaCapturedCommandProperty = BindableProperty.Create(nameof(MediaCapturedCommand), typeof(ICommand), typeof(CameraView));
		public static readonly BindableProperty MediaCaptureFailedCommandProperty = BindableProperty.Create(nameof(MediaCaptureFailedCommand), typeof(ICommand), typeof(CameraView));
		public static readonly BindableProperty OnAvailableCommandProperty = BindableProperty.Create(nameof(OnAvailableCommand), typeof(ICommand), typeof(CameraView));
		public static readonly BindableProperty ShutterCommandProperty = BindableProperty.Create(nameof(ShutterCommand), typeof(ICommand), typeof(CameraView), defaultBindingMode: BindingMode.OneWayToSource);

		public ICommand MediaCapturedCommand
		{
			get => (ICommand)GetValue(MediaCapturedCommandProperty);
			set => SetValue(MediaCapturedCommandProperty, value);
		}
		
		public ICommand MediaCaptureFailedCommand
		{
			get => (ICommand)GetValue(MediaCaptureFailedCommandProperty);
			set => SetValue(MediaCaptureFailedCommandProperty, value);
		}
		
		public ICommand OnAvailableCommand
		{
			get => (ICommand)GetValue(OnAvailableCommandProperty);
			set => SetValue(OnAvailableCommandProperty, value);
		}
		
		public ICommand ShutterCommand
		{
			get => (ICommand)GetValue(ShutterCommandProperty);
			set => SetValue(ShutterCommandProperty, value);
		}
```
The ShutterCommand is a special case because it is a command that's injected in the ViewModel to trigger the shutter.
The following approach with the code example above would make the commands work on all platforms by modifying the constructor of CameraView:
```
		public CameraView()
		{
			MediaCaptured += (sender, args) => MediaCapturedCommand?.Execute(args);
			MediaCaptureFailed += (sender, s) => MediaCaptureFailedCommand?.Execute(s);
			OnAvailable += (sender, b) => OnAvailableCommand?.Execute(b);
			ShutterCommand = new Command(Shutter);
		}
```

## Intended Use Case
Any MVVM app using CameraView.

pictos:
### Description of Change ###

Added Commands into the CameraView control.

### Bugs Fixed ###

- Fixes #285

Provide links to issues here. Ensure that a GitHub issue was created for your feature or bug fix before sending PR.

### API Changes ###

Added: 
 
- `ICommand ShutterCommand{ get; set; }`
- `object ShutterCommandParameter {get; set; };`

### Behavioral Changes ###

None

### PR Checklist ###

- [ ] Updated documentation


jfversluis (approved):
Looks good! 

Two things:
1. Should be do something with the CanExecute? Since we don't have anything visual attached I don't think we do...
2. How do other controls handle both Clicked and the Command? It seems that you can cause some weird behavior when you set these both now?

XamarinCommunityToolkit/Views/CameraView/CameraView.shared.cs
```diff
@@ -13,6 +14,30 @@ public class CameraView : View
 
 		internal event EventHandler ShutterClicked;
 
+		public static readonly BindableProperty ShutterCommandProperty =
+			BindableProperty.Create(nameof(ShutterCommand),
+						typeof(ICommand),
+						typeof(CameraView),
+						default(ICommand));
```
AndreiMisiukevich, line 21: ```default(ICommand)``` can be removed

XamarinCommunityToolkit/Views/CameraView/CameraView.shared.cs
```diff
@@ -13,6 +14,30 @@ public class CameraView : View
...
+		{
+			get => (ICommand)GetValue(ShutterCommandProperty);
+			set => SetValue(ShutterCommandProperty, value);
+		}
+
+		public static readonly BindableProperty ShutterCommandParameterProperty =
+			BindableProperty.Create(nameof(ShutterCommandParameter),
+						typeof(object),
+						typeof(CameraView),
+						default);
```
AndreiMisiukevich, line 33: default can be removed

pictos:
@jfversluis 
> Should be do something with the CanExecute? Since we don't have anything visual attached I don't think we do...

I think this should be on the dev side. For us, we don't need to check it.


> How do other controls handle both Clicked and the Command? It seems that you can cause some weird behavior when you set these both now?

I used Xamarin.Forms `Button` implementation and looks like they use the same event call. You can see [here](https://github.com/xamarin/Xamarin.Forms/blob/f35ae07a0a8471d255f7a1ebdd51499e10e0a4cb/Xamarin.Forms.Core/ButtonElement.cs#L46)

AndreiMisiukevich:
> @jfversluis
> 
> > Should be do something with the CanExecute? Since we don't have anything visual attached I don't think we do...
> 
> I think this should be on the dev side. For us, we don't need to check it.
> 
> > How do other controls handle both Clicked and the Command? It seems that you can cause some weird behavior when you set these both now?
> 
> I used Xamarin.Forms `Button` implementation and looks like they use the same event call. You can see [here](https://github.com/xamarin/Xamarin.Forms/blob/f35ae07a0a8471d255f7a1ebdd51499e10e0a4cb/Xamarin.Forms.Core/ButtonElement.cs#L46)

As I see they handle CanExecute. Probably we should as well?

pictos:
@AndreiMisiukevich I marked the wrong method, please look at the link again... I think they use the CanExecute to handle with the `IsEnable` property.

pictos:
I see I think they handle `CanExecute` to apply the `IsEnabled = false` design to the button. So we don't need to care about it, since `Shutter` isn't a control

AndreiMisiukevich (approved):

acuntex:
@jfversluis @AndreiMisiukevich @pictos 
Sorry, but currently this implementation makes no sense.

What's currently implemented could be resolved by a EventToCommandBehavior.

The developer would still have to implement a code-behind (xaml.cs) Clicked-Handler to trigger the method Shutter. This does not comply with MVVM-guidelines.

The use case would be the following:
A developer can trigger Shutter from the ViewModel. This can be used e.g. to implement a series of photos or a countdown.

This would be the correct code:

```
		public CameraView()
		{
			ShutterCommand = new Command(Shutter);
		}

		public static readonly BindableProperty ShutterCommandProperty = BindableProperty.Create(nameof(ShutterCommand), typeof(ICommand), typeof(CameraView), defaultBindingMode: BindingMode.OneWayToSource);
		public ICommand ShutterCommand
		{
			get => (ICommand)GetValue(ShutterCommandProperty);
			set => SetValue(ShutterCommandProperty, value);
		}
```

Please note the defaultBindingMode and the creation of the Command in the constructor. This way you can inject a command in your ViewModel that can be executed however the logic of your ViewModel is without any code behind.

AndreiMisiukevich:
> @jfversluis @AndreiMisiukevich @pictos
> Sorry, but currently this implementation makes no sense.
> 
> What's currently implemented could be resolved by a EventToCommandBehavior.
> 
> The developer would still have to implement a code-behind (xaml.cs) Clicked-Handler to trigger the method Shutter. This does not comply with MVVM-guidelines.
> 
> The use case would be the following:
> A developer can trigger Shutter from the ViewModel. This can be used e.g. to implement a series of photos or a countdown.
> 
> This would be the correct code:
> 
> ```
> 		public CameraView()
> 		{
> 			ShutterCommand = new Command(Shutter);
> 		}
> 
> 		public static readonly BindableProperty ShutterCommandProperty = BindableProperty.Create(nameof(ShutterCommand), typeof(ICommand), typeof(CameraView), defaultBindingMode: BindingMode.OneWayToSource);
> 		public ICommand ShutterCommand
> 		{
> 			get => (ICommand)GetValue(ShutterCommandProperty);
> 			set => SetValue(ShutterCommandProperty, value);
> 		}
> ```
> 
> Please note the defaultBindingMode and the creation of the Command in the constructor. This way you can inject a command in your ViewModel that can be executed however the logic of your ViewModel is without any code behind.


Your implementation will cause the invocation of both **event** and **command**.
Well, current ShutterCommand works as an event (triggers when something happens of view side), not as a method (for calling UI code from ViewModel)

jfversluis:
Hm I think I see what you mean. Basically you are saying that this doesn't make much sense because you will still need to call `Shutter()` for the Command to be invoked, right?

And you cannot call `Shutter()` in an MVVM way, so you will still have to do something in your code-behind.

Am I then understanding you correctly? You might have a point :)

acuntex:
@AndreiMisiukevich @jfversluis 
No, in my implementation the ViewModel triggers the Command and the Command triggers the event.

At the moment there is no way to trigger the Shutter from the View Model.

And Again: The current implementation could have been resolved by a EventToCommandBehavior.
There is no practical use case for the current implementation. Shutter is never triggered by the code in XCT. The only way you can trigger shutter is by implementing code-behind (xaml.cs).

If a developer wants to develop according to MVVM guidelines and use a view model and trigger Shutter from his logic there, then this would be the only way.

acuntex:
@AndreiMisiukevich It's basically the same as the ForceCloseCommand in https://github.com/AndreiMisiukevich/ContextMenu/ :-)

AndreiMisiukevich:
> @AndreiMisiukevich It's basically the same as the ForceCloseCommand in https://github.com/AndreiMisiukevich/ContextMenu/ :-)

That makes sense
Yes, you are right) It seems we should make OneWayToSource binding

PR #336, opened 2020-09-28
fix iOS disposing crash

jessejiang0214:
### Description of Change ###

For rateObserver it should remove it first and dispose of it. I got some crashes from iOS native level when App to go the background 
Also, the dispose need to catch the exceptions, otherwise where from the Xamarin.Forms level can catch that.

### Bugs Fixed ###

- Fixes #

https://github.com/xamarin/XamarinCommunityToolkit/issues/335
https://github.com/xamarin/Xamarin.Forms/issues/11412

### API Changes ###

None

### Behavioral Changes ###

The App won't crash after App go to the background when the video is pause or finish.

### PR Checklist ###

- [ ] Updated documentation


XamarinCommunityToolkit/Views/MediaElement/MediaElement.ios.cs
```diff
@@ -164,7 +180,9 @@ void RemoveStatusObserver()
 				try
 				{
 					avPlayerViewController?.Player?.CurrentItem?.RemoveObserver(statusObserver, "status");
+					statusObserver.Dispose();
 				}
+				catch { }
```
AndreiMisiukevich, line 185: Hi, I doubt that such statements are a good idea. Catching all exceptions and do nothing, I mean.

jessejiang0214: The thing is it will throw exceptions when the observer was removed for some reason. I got a few crashing reports regarding this, but I have no idea how it happens.

But if we didn't catch the exception from the PCL level there's no way to catch the exceptions and the App will crash.

AndreiMisiukevich:
@jsuarezruiz @jessejiang0214 can we fix these merge conflicts?

jfversluis:
Just for my understanding, what is the story with this PR? This is based of the branch from Javier? We should first merge that one before we do this one, right?

peterfoot:
Excuse me if I'm missing something but I've been out of the loop with MediaElement. I thought I had submitted a PR with a fix for this some time ago. Was it that the Pause() had to be invoked on the UI thread and if invoked via Dispose it's on a different thread?

jessejiang0214:
@AndreiMisiukevich Merged.

jessejiang0214:
Hi @jfversluis I agree, as the Master doesn't have MediaElement control, so I have to based on that branch.

jessejiang0214:
@peterfoot I raised this issue a long time ago, and I find some code change in Xamarin.Forms repo, but that fix didn't fix the issue completely. Then my customer pushes me to fix that, so I fixed that.

AndreiMisiukevich:
@jessejiang0214 

Can you please rebase onto **xamarin:mediaelement** as well?

Now, as you see, there a lot of changes from other commits

jessejiang0214:
@AndreiMisiukevich Rebased

AndreiMisiukevich:
Not sure why we see all these commits so far.
I guess they shouldn't be there

image

jessejiang0214:
@AndreiMisiukevich I have no idea why it looks like that. But from my memory, it seems @jsuarezruiz push code to this branch. From the comments history, this one is rebased on mediaelement branch.

Screen Shot 2020-10-02 at 4 02 18 pm

AndreiMisiukevich:
@jfversluis how will we handle all these PRs?
Should we wait till others are merged?

jfversluis:
Yeah seems like something weird is going on here. I would propose to first merge `MediaElement` into main before we take in any changes on that one.

Would you be willing to add your fix to that when we get it merged @jessejiang0214 ?

jessejiang0214:
Hi @jfversluis it sounds good to me.

jfversluis:
@jessejiang0214 I merged `MediaElement` could you check what changes need to still go in from you if any? cc: @peterfoot

jessejiang0214:
Hi @jfversluis , I push my changes based on the new master, so how should I create PR for that, need another new issue or reopen this one?

peterfoot:
@jessejiang0214 I opened a fresh PR and referenced the original and the original issue. You may want to check out what I found out and changed in my PR related to Dispose. The base Dispose calls OnElementChanged so there were two sets of cleanup being called. I got rid of the whole Dispose override and just ensured all the required cleanup is in OnElementChanged. I think the reason this became more unreliable is because of a change in behaviour of the base renderer dispose from the original version of Xamarin Forms it was created against. I think all the observers are removed in the right sequence.

peterfoot:
@jessejiang0214 for reference - new PR #381

jessejiang0214:
Hi @peterfoot I think you forget to remove rate observer and I would suggest we call dispose to the observer, cause there's a warning if you didn't call Dispose for the observer. Would you mind I push some code base on your PR?

peterfoot:
@jessejiang0214 Yes of course please go ahead and fix it!

jfversluis:
@jessejiang0214 I've merged Peter's latest PR into main, so you can simply target that now. Thanks :)

jessejiang0214:
@jfversluis new PR https://github.com/xamarin/XamarinCommunityToolkit/pull/392 
BTW, is there any demo code for some property only available in IOS? I want to add a new feature about the MediaElement which only available in IOS.

PR #346, opened 2020-09-30
Fixed ShutterCommand

### LINKED ISSUE #341: Improve CameraView Command support
Following the discussion (comment can be seen below), we should improve the Command support on the `CameraView` that was just merged. The current implementation only has limited use.

---

@jfversluis @AndreiMisiukevich @pictos 
Sorry, but currently this implementation makes no sense.

What's currently implemented could be resolved by a EventToCommandBehavior.

The developer would still have to implement a code-behind (xaml.cs) Clicked-Handler to trigger the method Shutter. This does not comply with MVVM-guidelines.

The use case would be the following:
A developer can trigger Shutter from the ViewModel. This can be used e.g. to implement a series of photos or a countdown.

This would be the correct code:

```
		public CameraView()
		{
			ShutterCommand = new Command(Shutter);
		}

		public static readonly BindableProperty ShutterCommandProperty = BindableProperty.Create(nameof(ShutterCommand), typeof(ICommand), typeof(CameraView), defaultBindingMode: BindingMode.OneWayToSource);
		public ICommand ShutterCommand
		{
			get => (ICommand)GetValue(ShutterCommandProperty);
			set => SetValue(ShutterCommandProperty, value);
		}
```

Please note the defaultBindingMode and the creation of the Command in the constructor. This way you can inject a command in your ViewModel that can be executed however the logic of your ViewModel is without any code behind.

_Originally posted by @acuntex in https://github.com/xamarin/XamarinCommunityToolkit/pull/334#issuecomment-701223975_

pictos:
### Description of Change ###

Now ShutterCommand is a `readonly BindableProperty` and `BindingMode.OneWayToSource`, this way the dev doesn't need to call `CameraView.Shutter()` method.

### Bugs Fixed ###

- Fixes #341

### API Changes ###

Added: 
- `internal static readonly BindablePropertyKey ShutterCommandPropertyKey`

Changed:

- `public static readonly BindableProperty ShutterCommandProperty = ShutterCommandPropertyKey.BindableProperty;`
 - `public ICommand ShutterCommand => (ICommand)GetValue(ShutterCommandProperty);`

Removed :
- `public object ShutterCommandParameter {get; set;}  //BP`

### PR Checklist ###

- [ ] Updated documentation


XamarinCommunityToolkit/Views/CameraView/CameraView.shared.cs
```diff
@@ -14,27 +14,17 @@ public class CameraView : View
...
-			BindableProperty.Create(nameof(ShutterCommandParameter),
-						typeof(object),
-						typeof(CameraView));
+		public static readonly BindableProperty ShutterCommandProperty = ShutterCommandPropertyKey.BindableProperty;
 
-		public object ShutterCommandParameter
-		{
-			get => GetValue(ShutterCommandParameterProperty);
-			set => SetValue(ShutterCommandParameterProperty, value);
-		}
```
ahoefling, line 37: What is the reasoning behind removing the `ShutterCommandParameter`?

I find parameters as implemented to be very useful when you need to pass a parameter to the ViewModel method that is being invoked by the `ICommand`. For example you can bind to non ViewModel objects such as a single item of a collection. 

I think having the `ShutterCommandParameter` will be useful when you need to pass additional information to the ViewModel.

A viewmodel example would look something like this
```c#
public class CameraViewModel
{
    public ICommand Shutter => new Command<object>(OnShutter);

    void OnShutter(object data)
    {
        // do something with the data 
    }
}
```

XamarinCommunityToolkit/Views/CameraView/CameraView.shared.cs
```diff
@@ -14,27 +14,17 @@ public class CameraView : View
...
-		public static readonly BindableProperty ShutterCommandProperty =
-			BindableProperty.Create(nameof(ShutterCommand),
-						typeof(ICommand),
-						typeof(CameraView));
+		internal static readonly BindablePropertyKey ShutterCommandPropertyKey =
+			BindableProperty.CreateReadOnly(nameof(ShutterCommand),
+				typeof(ICommand),
+				typeof(CameraView),
+				default,
+				BindingMode.OneWayToSource,
```
ahoefling, line 22: This is the correct `BindingMode`! This part is very important to get right.

Good Job 🎉

ahoefling (changes requested):
This looks really good! I reviewed how I build ICommand binding implementations and how Xamarin.Forms does and this follows the best practices for making a readonly bindable property. My only concern is the removal of the `ShutterCommandParameter`. I find that useful and I am not sure why it was removed.

XamarinCommunityToolkit/Views/CameraView/CameraView.shared.cs
pictos: The CommandParameter makes sense when you can implement your own Command, like `Button.Command`, this case `CameraView.ShutterCommand` is `readonly`, so the dev can't override it. If you see the code sample, I need to have a Command for the Button that triggers the `CameraView.ShutterCommand`. So, if you want to use some `CommandParameter` you should use the `Button.CommandParameter`. Is this makes sense for you?

ahoefling: Thank you for the explanation, this does make sense. Since the `ICommand` is being instantiated by the toolkit a developer has no control on the method being invoked. This means the ViewModel just creates the `ICommand` property and it is automatically set.

There is a little bit of magic here, which is okay. I think we need to be diligent with our documentation on this. It will be very easy to get confused on how this property is instantiated. Especially if you set it in your ViewModel and it doesn't do anything.

ahoefling (dismissed):
Based on the explanation of how the `ShutterCommand` works, I have no issues approving this PR. I think this looks good, thank you for the contribution

pictos:
@acuntex can you review this PR and see if it fits your expectations?

acuntex:
@pictos Looks good! Thank you!

XamarinCommunityToolkit/Views/CameraView/CameraView.shared.cs
```diff
@@ -14,27 +14,17 @@ public class CameraView : View
...
-		{
-			get => (ICommand)GetValue(ShutterCommandProperty);
-			set => SetValue(ShutterCommandProperty, value);
-		}
-
-		public static readonly BindableProperty ShutterCommandParameterProperty =
-			BindableProperty.Create(nameof(ShutterCommandParameter),
-						typeof(object),
-						typeof(CameraView));
+		public static readonly BindableProperty ShutterCommandProperty = ShutterCommandPropertyKey.BindableProperty;
```
AndreiMisiukevich, line 25: How about getting rid of ShutterCommandPropertyKey? (we can just put ".BindableProperty" at the end of the chain)

acuntex: Is there anything wrong about:
`
public static readonly BindableProperty ShutterCommandProperty = BindableProperty.Create(nameof(ShutterCommand), typeof(ICommand), typeof(CameraView), defaultBindingMode: BindingMode.OneWayToSource);
`
Just curious what's the difference to the BindablePropertyKey approach.

ahoefling: When I was reading the docs and Xamarin.Forms platform code I noticed this is a common place patter for ReadOnly bindable properties. I had the same question the other night when I was reviewing this code and came to the conclusion that I think this is okay.

As far as I understand the `BindablePropertyKey` allows for readonly bindable properties. 

If we use that traditional approach, I think there will be problems. Given the following scenario
- Using BindableProperty.Create
- Using getter/setter
- Using `BindingMode.OneWayToSource`

The code would allow the ViewModel to set the `ICommand` property but it won't be updated on control. This means the ViewModel will lose it's `ICommand` reference to the shutter and the code won't be able to execute it from the ViewModel.

bmacombe: I think the BindablePropertyKey is correct looking at the docs also.  It kind of hurts my head thinking about it, I've never had a need for a VM to use a command to execute something on a control/view, it's always View action to VM execute something.

XamarinCommunityToolkitSample/ViewModels/Views/CameraViewModel.cs
```diff
@@ -0,0 +1,26 @@
...
+			ShutterCommand = new Command(Shutter);
+			CameraShutterCommand = new Command(DoNothing);
+		}
+
+		void Shutter() => CameraShutterCommand?.Execute(null);
+
+		void DoNothing() =>
+			Console.WriteLine("This is just to prove that the user can't override the CameraViewShutterCommand");
+	}
+}
```
AndreiMisiukevich, line 26: I don't think we need to keep both

The CameraView works in OneWayToSource mode
The button works in OneWay mode

It means that this ViewModel doesn't make any sense :)

We can bind Button's command to CameraView's command directly in xaml via x:Source

XamarinCommunityToolkitSample/Pages/Views/CameraViewPage.xaml
```diff
@@ -69,7 +70,7 @@
 
                 <Button
                     x:Name="doCameraThings"
-                    Clicked="DoCameraThings_Clicked"
+                    Command="{Binding ShutterCommand}"
```
AndreiMisiukevich, line 73: ```Command="{Binding ShutterCommand, Source={x:Reference cameraView}}"```

No need for ViewModel

AndreiMisiukevich (changes requested):
Could you please address my comment? :)

Thanks!

XamarinCommunityToolkitSample/Pages/Views/CameraViewPage.xaml
pictos: ###

AndreiMisiukevich (approved):

jfversluis (approved):

PR #347, opened 2020-09-30
Adding GravatarImageSource

### LINKED ISSUE #312: [Spec]  Gravatar Image Source
# GravatarImageSource

It is fairly common for apps to want to use a Gravatar Image based on a User's Email. The Toolkit should include a GravatarImageSource. This is something I've already got in my toolkit but am happy to contribute to the CommunityToolkit. 

# API

### Values

- FileNotFound
- MysteryPerson
- Identicon
- MonsterId
- Wavatar
- Retro
- Robohash
- Blank

### Properties

| API | Description |
| Email | Gets or sets Email. |
| Size | Gets or sets Size. |
| DefaultGravatar | Gets or sets DefaultGravatar. |

### Events

none
 
# Scenarios

You have an authenticated user and would like to display a profile picture of the user using their Gravatar Image

# Platform Compatibility
- Target Frameworks: 
  - iOS:  supported
  - Android: supported
  - UWP: supported
# Backward Compatibility

# Difficulty : low

dansiegel:
### Description of Change ###

Adds a new GravatarImageSource that can be easily added for any image source using a user's email address. This utilizes the image provided by gravatar.com and supports all of their default image results to handle when an email address has no associated Gravatar.

### Issues ###

- closes #312

### Platforms Supported ###

- Android
- iOS
- macOS
- UWP
- WPF
- Tizen

### API Changes ###

Added: 

```csharp
public enum DefaultGravatar
{
	FileNotFound,
	MysteryPerson,
	Identicon,
	MonsterId,
	Wavatar,
	Retro,
	Robohash,
	Blank
}

public class GravatarImageSource : ImageSource
{
    public string Email { get; set; }
    public int Size { get; set; }
    public DefaultGravatar Default { get; set; }
    public boo CachingEnabled { get; set; }
    public TimeSpan CacheValidity { get; set; }
}
```

Note that by default Caching is enabled for 7 days. An additional XAML Extension has been provided to make it even easier to use.

### Behavioral Changes ###

none

### PR Checklist ###

- [ ] Updated documentation


XamarinCommunityToolkit/Views/GravatarImageSource/UWP/GravatarImageSourceHandler.uwp.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Threading.Tasks;
+using Windows.Storage;
+using Windows.UI.Xaml.Media.Imaging;
+using Xamarin.Forms.Platform.UWP;
+
+namespace Xamarin.CommunityToolkit.UI.Views
+{
+	public class GravatarImageSourceHandler : IImageSourceHandler
+	{
+		public async Task<Windows.UI.Xaml.Media.ImageSource> LoadImageAsync(Xamarin.Forms.ImageSource imagesource, CancellationToken cancellationToken = default)
```
pictos, line 13: Can you simplify the name removing `Xamarin` prefix? If it's possible can you use an alias instead of the full namespace, to follow our code style?

XamarinCommunityToolkit/Views/GravatarImageSource/UWP/GravatarImageSourceHandler.uwp.cs
```diff
@@ -0,0 +1,25 @@
...
+		{
+			var fileInfo = await GravatarHandlerUtil.Load(imagesource, 1, ApplicationData.Current.LocalCacheFolder.Path).ConfigureAwait(false);
+
+			if (fileInfo?.Exists ?? false)
+			{
```
pictos, line 18: We don't use braces for the one-line statement. Please fix the others, to match our code style.

XamarinCommunityToolkit/Views/GravatarImageSource/MacOS/GravatarImageSourceHandler.macos.cs
```diff
@@ -0,0 +1,33 @@
...
+
+			return null;
+		}
+
+		static string GetCacheDirectory()
+		{
+			var dirs = NSSearchPath.GetDirectories(NSSearchPathDirectory.CachesDirectory, NSSearchPathDomain.User);
+			if (dirs == null || dirs.Length == 0)
+			{
+				// this should never happen...
```
pictos, line 27: Do we need this comment? If this should never happen, can be a good idea trown an exception?

XamarinCommunityToolkit/Views/GravatarImageSource/Wpf/GravatarImageSourceHandler.wpf.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Threading.Tasks;
+using System.Windows.Media;
+using System.Windows.Media.Imaging;
+using Xamarin.Forms.Platform.WPF;
+
+namespace Xamarin.CommunityToolkit.UI.Views
+{
+	public class GravatarImageSourceHandler : IImageSourceHandler
+	{
+		public async Task<System.Windows.Media.ImageSource> LoadImageAsync(Xamarin.Forms.ImageSource imagesource, CancellationToken cancellationToken = default)
```
pictos, line 13: Can you use an alias instead of full namespace to follow our code style?

XamarinCommunityToolkit/Views/GravatarImageSource/iOS/GravatarImageSourceHandler.ios.cs
```diff
@@ -0,0 +1,33 @@
...
+		{
+			var fileInfo = await GravatarHandlerUtil.Load(imagesource, scale, GetCacheDirectory()).ConfigureAwait(false);
+
+			if (fileInfo?.Exists ?? false)
+				return UIImage.FromFile(fileInfo.FullName);
+
+			return null;
+		}
+
+		static string GetCacheDirectory()
```
pictos, line 22: If I'm not wrong this method is the same as .macOS. Can we avoid code duplication?

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
```diff
@@ -0,0 +1,42 @@
+using System.IO;
+using System.Threading.Tasks;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.UI.Views
+{
+#if !NETSTANDARD
```
pictos, line 7: Is this necessary?

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarImageExtension.shared.cs
```diff
@@ -0,0 +1,29 @@
...
+using Xamarin.Forms.Xaml;
+
+namespace Xamarin.CommunityToolkit.UI.Views
+{
+	[ContentProperty(nameof(Email))]
+	public class GravatarImageExtension : GravatarImageSource, IMarkupExtension<GravatarImageSource>
+	{
+		public GravatarImageSource ProvideValue(IServiceProvider serviceProvider)
+		{
+			if (serviceProvider is null)
```
pictos, line 12: We don't use `is` to check for null, can you please change it (and others) to `==` or `!=`, to follow our code style?

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarImageSource.shared.cs
```diff
@@ -0,0 +1,136 @@
...
+#endif
+
+		internal string CacheFileName(float scale) => $"{GetMd5Hash(Email)}-{Size}@{scale}x.png";
+
+		internal async Task<byte[]> GetGravatarAsync(float scale)
+		{
+#if NETSTANDARD1_0
+			throw new NotImplementedException();
+#else
+			using var client = new HttpClient();
```
pictos, line 85: Is it a good idea, create a new HttpClient for each call?

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarImageSource.shared.cs
```diff
@@ -0,0 +1,136 @@
...
+			if (!response.IsSuccessStatusCode)
+			{
+				return Array.Empty<byte>();
+			}
+
+			return await response.Content.ReadAsByteArrayAsync().ConfigureAwait(false);
+#endif
+		}
+
+		const string requestUriFormat = "https://www.gravatar.com/avatar/{0}?s={1}&d={2}";
```
pictos, line 97: Can you move this const to the top of the file?

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarImageSource.shared.cs
```diff
@@ -0,0 +1,136 @@
...
+			};
+		}
+
+		string GetMd5Hash(string str)
+		{
+#if NETSTANDARD1_0
+			throw new NotImplementedException();
+#else
+			byte[] hash = null;
+			using (var md5 = MD5.Create())
```
pictos, line 117: Can you use the C#8 style? For the HttpClient implementation you use the C#8, would be good to follow the same pattern.

pictos (changes requested):
@dansiegel thanks for this amazing contribution. I didn't run the code yet to test. Do you mind adding some unit tests? If you can't do it don't worry I'll do it in the future (:

XamarinCommunityToolkit/Views/GravatarImageSource/MacOS/GravatarImageSourceHandler.macos.cs
dansiegel: this is the literal copy from Xamarin.Essentials since there is no dependency on Essentials

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
dansiegel: I could make it netstandard1.0 only... since it's not 1.0 compliant... but ultimately this is only useful on an actual platform that could implement an Image Handler

XamarinCommunityToolkit/Views/GravatarImageSource/iOS/GravatarImageSourceHandler.ios.cs
dansiegel: it is... I'll look at how I can share it

pictos: You can create a file with extension `.ios.macos.cs`

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
pictos: But if the user tries to use it in a platform without support, like GTK. Will he see a good error message? I'm guessing here, but I think he will face a runtime exception telling that this method is missing... And a _correct_ message should be something like "We don't support this platform". Thoughts?

XamarinCommunityToolkit/Extensions/GravatarImageExtension.shared.cs
```diff
@@ -0,0 +1,30 @@
...
+	public class GravatarImageExtension : GravatarImageSource, IMarkupExtension<GravatarImageSource>
+	{
+		public GravatarImageSource ProvideValue(IServiceProvider serviceProvider)
+		{
+			if (serviceProvider == null)
+			{
+				throw new ArgumentNullException(nameof(serviceProvider));
+			}
+
+			if (BindingContext is null && !IsSet(BindingContextProperty))
```
pictos, line 18:
```suggestion
			if (BindingContext == null && !IsSet(BindingContextProperty))
```

XamarinCommunityToolkit/Views/GravatarImageSource/iOS/GravatarImageSourceHandler.ios.cs
dansiegel: Yeah. I ended up doing that. Wasn't a huge fan of that as each platform has its own folder and only that method can be shared since Mac uses an NSImage and iOS uses UIImage

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
dansiegel: Well this library took an explicit stance against supporting GTK so that example is a moot point. Both WPF and GTK use the same TFM until the transition to .NET 5. 

The Util class is meant simply to consolidate the logic that belongs in the platform ImageHandlers but would be common between each of them.

XamarinCommunityToolkit/Views/GravatarImageSource/iOS/GravatarImageSourceHandler.ios.cs
pictos: @dansiegel, feel free to disagree with my ideas! My point for it is just to avoid code duplication hehe. In the future, if Apple changes the API we can move back this method as it's today.

dansiegel: Trust me I agree it needed to be deduplicated. Just didn't have great ideas for it.

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
pictos: well, if I'm not wrong, we can't target .NET5, because XF doesn't run on it. And when we have MAUI/.NET6 we need to see what we need to change to support both.

XamarinCommunityToolkit/Views/GravatarImageSource/iOS/GravatarImageSourceHandler.ios.cs
pictos: If I have some insight I'll share with you

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
```diff
@@ -0,0 +1,100 @@
...
+
+				return cacheFileInfo;
+			}
+
+			return null;
+		}
+
+		static void SaveImage(FileInfo cacheFileInfo, byte[] imageBytes)
+		{
+			if (imageBytes.Length < 1)
```
AndreiMisiukevich, line 38: Should we check imageBytes for null?

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
```diff
@@ -0,0 +1,100 @@
...
+				return;
+
+			// Delete Cached File
+			if (cacheFileInfo.Exists)
+				cacheFileInfo.Delete();
+
+			File.WriteAllBytes(cacheFileInfo.FullName, imageBytes);
+		}
```
AndreiMisiukevich, line 46: Should we add some "locks" to this file?

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
```diff
@@ -0,0 +1,100 @@
...
+			var requestUri = GetGravatarUri(email, size, scale, defaultGravatar);
+			using var response = await lazyHttp.Value.GetAsync(requestUri).ConfigureAwait(false);
+
+			if (!response.IsSuccessStatusCode)
+				return Array.Empty<byte>();
+
+			return await response.Content.ReadAsByteArrayAsync().ConfigureAwait(false);
+		}
+
+		static string GetGravatarUri(string email, int size, float scale, DefaultGravatar defaultGravatar) => string.Format(requestUriFormat, GetMd5Hash(email), size * scale, DefaultGravatarName(defaultGravatar));
```
AndreiMisiukevich, line 69:
```suggestion
		static string GetGravatarUri(string email, int size, float scale, DefaultGravatar defaultGravatar) 
			=> string.Format(requestUriFormat, GetMd5Hash(email), size * scale, DefaultGravatarName(defaultGravatar));
```

AndreiMisiukevich (approved):
I like it

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
dansiegel: it shouldn't ever be null since any 200 response will give you an image and any non 200 response gives an empty array

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
dansiegel: other than the sample I would say it would be quite unusual to ever have more than one GravatarImageSource on a page with the same email really the worst that is going to happen is that you download it twice... for reading the file we shouldn't need a lock

XamarinCommunityToolkit/Views/GravatarImageSource/iOS/GravatarImageSourceHandler.ios.macos.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Xamarin.CommunityToolkit.UI.Views
+{
+	public partial class GravatarImageSourceHandler
+	{
+		static string GetCacheDirectory()
+		{
+			var dirs = NSSearchPath.GetDirectories(NSSearchPathDirectory.CachesDirectory, NSSearchPathDomain.User);
+			if (dirs == null || dirs.Length == 0)
+			{
+				// this should never happen...
```
pictos, line 12: Do you mind to add a log here, telling the dev that something odd happened? Just to make easy for him to track the error and report

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
```diff
@@ -0,0 +1,106 @@
...
+		{
+			if (!file.Directory.Exists)
+				file.Directory.Create();
+
+			return cachingEnabled && file.Exists && file.CreationTime.Add(cacheValidity) > DateTime.Now;
+		}
+
+		static string CacheFileName(GravatarImageSource gis, float scale)
+		{
+			return $"{GetMd5Hash(gis.Email)}-{gis.Size}@{scale}x.png";
```
pictos, line 58: Can you use the body expression for methods? To follow our code style.

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
```diff
@@ -0,0 +1,106 @@
...
+
+			if (!response.IsSuccessStatusCode)
+				return Array.Empty<byte>();
+
+			return await response.Content.ReadAsByteArrayAsync().ConfigureAwait(false);
+		}
+
+		static string GetGravatarUri(string email, int size, float scale, DefaultGravatar defaultGravatar)
+		{
+			return string.Format(requestUriFormat, GetMd5Hash(email), size * scale, DefaultGravatarName(defaultGravatar));
```
pictos, line 74: Can you use the body expression for methods? To follow our code style.

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
```diff
@@ -0,0 +1,106 @@
...
+		{
+			using var md5 = MD5.Create();
+			var hash = md5.ComputeHash(Encoding.UTF8.GetBytes(str));
+
+			var sBuilder = new StringBuilder();
+
+			if (hash != null)
+			{
+				for (var i = 0; i < hash.Length; i++)
+				{
```
pictos, line 97: Can you remove the brackets for this `for`?  The if it's ok. To follow our code style.

pictos (changes requested):
@dansiegel we are getting there❣ Most of the comments are about code style.

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
AndreiMisiukevich: Yeah, I see.
But in case of slow internet and navigating from one page (Overview) to another one (Details) we can cause a situation when two threads are writing bytes to the same file.

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
```diff
@@ -0,0 +1,106 @@
...
+			if (!file.Directory.Exists)
+				file.Directory.Create();
+
+			return cachingEnabled && file.Exists && file.CreationTime.Add(cacheValidity) > DateTime.Now;
+		}
+
+		static string CacheFileName(GravatarImageSource gis, float scale)
+		{
+			return $"{GetMd5Hash(gis.Email)}-{gis.Size}@{scale}x.png";
+		}
```
AndreiMisiukevich, line 59:
```suggestion
			=> $"{GetMd5Hash(gis.Email)}-{gis.Size}@{scale}x.png";
```

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
```diff
@@ -0,0 +1,104 @@
...
+			if (!response.IsSuccessStatusCode)
+				return Array.Empty<byte>();
+
+			return await response.Content.ReadAsByteArrayAsync().ConfigureAwait(false);
+		}
+
+		static string GetGravatarUri(string email, int size, float scale, DefaultGravatar defaultGravatar)
+		{
+			return string.Format(requestUriFormat, GetMd5Hash(email), size * scale, DefaultGravatarName(defaultGravatar));
+		}
```
AndreiMisiukevich, line 73:
```suggestion
			=> string.Format(requestUriFormat, GetMd5Hash(email), size * scale, DefaultGravatarName(defaultGravatar));
```

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
```diff
@@ -0,0 +1,102 @@
...
+			var hash = md5.ComputeHash(Encoding.UTF8.GetBytes(str));
+
+			var sBuilder = new StringBuilder();
+
+			if (hash != null)
+			{
+				for (var i = 0; i < hash.Length; i++)
+				{
+					sBuilder.Append(hash[i].ToString("x2"));
+				}
```
AndreiMisiukevich, line 95:
```suggestion
					sBuilder.Append(hash[i].ToString("x2"));
```

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
```diff
@@ -0,0 +1,100 @@
...
+
+		static string DefaultGravatarName(DefaultGravatar defaultGravatar)
+		{
+			return defaultGravatar switch
+			{
+				DefaultGravatar.FileNotFound => "404",
+				DefaultGravatar.MysteryPerson => "mp",
+				_ => $"{defaultGravatar}".ToLower(),
+			};
+		}
```
AndreiMisiukevich, line 81:
```suggestion
			=> defaultGravatar switch
			{
				DefaultGravatar.FileNotFound => "404",
				DefaultGravatar.MysteryPerson => "mp",
				_ => $"{defaultGravatar}".ToLower(),
			};
```

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
```diff
@@ -0,0 +1,98 @@
...
+			if (cacheFileInfo.Exists)
+				cacheFileInfo.Delete();
+
+			File.WriteAllBytes(cacheFileInfo.FullName, imageBytes);
+		}
+
+		static bool UseCacheFile(bool cachingEnabled, TimeSpan cacheValidity, FileInfo file)
+		{
+			if (!file.Directory.Exists)
+				file.Directory.Create();
```
AndreiMisiukevich, line 51: @dansiegel the same for the directory.
That's the possible case when "Create" is invoked twice in a row.

XamarinCommunityToolkit/Views/GravatarImageSource/GravatarHandlerUtil.shared.cs
dansiegel: this has been refactored and there are now locks around all file IO operations

AndreiMisiukevich (approved):
Needs others review

jfversluis (changes requested):
Build seems to fail

> Views/GravatarImageSource/Android/GravatarImageSourceHandler.android.cs(20,13): error CS1996: Cannot await in the body of a lock statement

dansiegel:
@jfversluis good catch... I'll update that shortly

jfversluis (approved):

PR #357, opened 2020-10-02
Adding Default XmlnsDefinition

### LINKED ISSUE #345: [Enhancement | Discussion] Look into using XmlnsDefinition
@dansiegel proposed to use `XmlnsDefinition` to make it easier to consume this library. Thoughts? Pros? Cons?

dansiegel:
### Description of Change ###

Adding XmlnsDefinitions for the Xamarin Community Toolkit. This utilizes a namespace based on what we see from Xamarin.Forms `http://xamarin.com/schemas/{year}/{product}`. The sample app has been updated to use the new `http://xamarin.com/schemas/2020/toolkit` namespace consistently using the `xct` prefix. 

**Q.** Why is the XmlnsDefinition Attribute excluded from .NET Standard 1.0?
**A.** Xamarin.Forms makes this API internal for netstandard1.0 due to an issue with Reflection methods that are not available. For those impacted, they can continue to use a traditional explicit xmlns for each clr namespace they require, or they can simply update to netstandard2.0
image
*from the Xamarin.Forms source*

### Issue ###

- closes #345 

### API Changes ###

none

### Behavioral Changes ###

When consuming the Xamarin Community Toolkit you will no longer need to declare a unique xmlns for each CLR namespace you're consuming in XAML (unless you're oddly using netstandard1.0)

### PR Checklist ###

### DIFF
--- a/XamarinCommunityToolkit/AssemblyInfo/AssemblyInfo.shared.cs
+++ b/XamarinCommunityToolkit/AssemblyInfo/AssemblyInfo.shared.cs
@@ -1,3 +1,12 @@
 ﻿using System.Runtime.CompilerServices;
+using Xamarin.Forms;
 
 [assembly: InternalsVisibleTo("Xamarin.CommunityToolkit.UnitTests")]
+
+#if !NETSTANDARD1_0
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2020/toolkit", "Xamarin.CommunityToolkit.Behaviors")]
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2020/toolkit", "Xamarin.CommunityToolkit.Converters")]
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2020/toolkit", "Xamarin.CommunityToolkit.Effects")]
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2020/toolkit", "Xamarin.CommunityToolkit.Extensions")]
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2020/toolkit", "Xamarin.CommunityToolkit.UI.Views")]
+#endif
--- a/XamarinCommunityToolkitSample/Pages/Views/AvatarViewPage.xaml
+++ b/XamarinCommunityToolkitSample/Pages/Views/AvatarViewPage.xaml
@@ -1,11 +1,10 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
-<pages:BasePage x:Class="Xamarin.CommunityToolkit.Sample.Pages.Views.AvatarViewPage"
-                xmlns="http://xamarin.com/schemas/2014/forms"
+<pages:BasePage xmlns="http://xamarin.com/schemas/2014/forms"
                 xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+                xmlns:xct="http://xamarin.com/schemas/2020/toolkit"
                 xmlns:pages="clr-namespace:Xamarin.CommunityToolkit.Sample.Pages"
-                xmlns:rsx="clr-namespace:Xamarin.CommunityToolkit.Extensions;assembly=Xamarin.CommunityToolkit"
                 xmlns:viewmodels="clr-namespace:Xamarin.CommunityToolkit.Sample.ViewModels.Views"
-                xmlns:views="clr-namespace:Xamarin.CommunityToolkit.UI.Views;assembly=Xamarin.CommunityToolkit">
+                x:Class="Xamarin.CommunityToolkit.Sample.Pages.Views.AvatarViewPage">
 
     <pages:BasePage.BindingContext>
         <viewmodels:AvatarViewViewModel />
@@ -22,7 +21,7 @@
                                      Orientation="Horizontal"
                                      Spacing="10">
 
-                            <views:AvatarView ColorTheme="{x:Static views:ColorTheme.Jungle}"
+                            <xct:AvatarView ColorTheme="{x:Static xct:ColorTheme.Jungle}"
                                               FontSize="Medium"
                                               Size="{Binding Value,
                                               Source={x:Reference Slider}}"
@@ -45,7 +44,7 @@
             </CollectionView.ItemTemplate>
         </CollectionView>
         <StackLayout Padding="{StaticResource ContentPadding}">
-            <Label Text="{rsx:Translate AvatarViewSizeText}" />
+            <Label Text="{xct:Translate AvatarViewSizeText}" />
             <Slider x:Name="Slider"
                     Maximum="80"
                     Minimum="40"
--- a/XamarinCommunityToolkitSample/Pages/Views/AvatarViewPage.xaml
+++ b/XamarinCommunityToolkitSample/Pages/Views/AvatarViewPage.xaml
@@ -27,10 +27,10 @@
                                               Source={x:Reference Slider}}"
                                               Text="{Binding Initials}">
 
-                                <views:AvatarView.Source>
+                                <xct:AvatarView.Source>
                                     <UriImageSource Uri="{Binding Source}" />
-                                </views:AvatarView.Source>
-                            </views:AvatarView>
+                                </xct:AvatarView.Source>
+                            </xct:AvatarView>
 
                             <Label FontSize="20"
                                    Text="{Binding Name}"
--- a/XamarinCommunityToolkit/Helpers/PreserveToolkitAttribute.shared.cs
+++ b/XamarinCommunityToolkit/Helpers/PreserveToolkitAttribute.shared.cs
@@ -1,0 +1,11 @@
+using System;
+using System.ComponentModel;
+
+namespace Xamarin.CommunityToolkit.Helpers
+{
+	[EditorBrowsable(EditorBrowsableState.Never)]
+	[AttributeUsage(AttributeTargets.Assembly)]
+	public sealed class PreserveToolkitAttribute : Attribute
+	{
+	}
+}
--- a/XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj
+++ b/XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project Sdk="MSBuild.Sdk.Extras/2.0.54">
+<Project Sdk="MSBuild.Sdk.Extras/2.1.2">
   <PropertyGroup>
     <TargetFrameworks>netstandard1.0;netstandard2.0;netstandard2.1;Xamarin.iOS10;MonoAndroid90;MonoAndroid10.0;Xamarin.TVOS10;Xamarin.WatchOS10;Xamarin.Mac20;tizen40</TargetFrameworks>
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);uap10.0.17763;netcoreapp3.1;net472;net471</TargetFrameworks>
@@ -22,21 +22,7 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <RepositoryUrl>https://github.com/xamarin/XamarinCommunityToolkit</RepositoryUrl>
     <PackageReleaseNotes>
-      
-      
-      
-      
-      
-      
-      
       <!-- See if we can host these on MS docs? -->
-    
-    
-    
-    
-    
-    
-    
     </PackageReleaseNotes>
     <DefineConstants>$(DefineConstants);</DefineConstants>
     <UseFullSemVerForNuGet>false</UseFullSemVerForNuGet>
@@ -120,7 +106,7 @@
     <AndroidResource Include="Resources\**\*.xml" />
     <AndroidResource Include="Resources\**\*.png" />
   </ItemGroup>
-  
+
   <ItemGroup Condition=" $(TargetFramework.StartsWith('MonoAndroid9.0')) ">
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
   </ItemGroup>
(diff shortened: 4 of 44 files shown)


sthewissen (approved):

sthewissen:
@jfversluis @AndreiMisiukevich Do we want to merge this in? Not sure if more debating on taking on the Forms namespace is needed?

dansiegel:
@sthewissen well I wouldn't say merge this yet... really we need to decide whether or not we're good with using the Forms namespace... if @jfversluis & @AndreiMisiukevich agree, then I'll update the sample app to use the namespace

AndreiMisiukevich:
I like this contribution :) But I am not sure about the exact namespace. @jfversluis we need your input here

AndreiMisiukevich (dismissed):

jfversluis:
I can see it go both ways. I certainly see how this makes it easier, especially with people transitioning controls like the Expander, MediaElement, etc. but might indeed also cause confusion as to where issues need to go.

This means you can do both, right? So you can still do both Xamarin.CommunityToolkit... but you will also see the things pop up in your IntelliSense for Xamarin.Forms?

I guess I am kind of onboard with, this. How do Forms members feel about this? @Redth @jsuarezruiz @mattleibow @PureWeen

Redth:
We should not share the same namespace as Xamarin.Forms as it would be a confusing and misleading.  Let's pick a reasonable namespace (ie: `Xamarin.Forms.Toolkit`) and declare a `XmlnsPrefixAttribute` so that intellisense can help sort out missing prefixes.

Redth (changes requested):
We need a unique namespace prefix and not to reuse the Xamarin.Forms one.

dansiegel:
Given the feedback from @Redth... there are two alternatives that I'm thinking:

- Just go stupid simple: `http://xamarincommunitytoolkit.com`
- Follow the default XF schema (`http://xamarin.com/schemas/2014/forms`): `http://xamarin.com/schemas/2020/toolkit`

Keep in mind that regardless of what we ultimately choose here the XAML intellisense will show you the full domain so it isn't like someone would have to remember what the exact uri is. Anyone have any thoughts?

AndreiMisiukevich:
http://xamarin.com/schemas/2020/toolkit -- looks good
Any thoughts?

AndreiMisiukevich:
> http://xamarin.com/schemas/2020/toolkit -- looks good
> Any thoughts?

Or even ```https://github.com/xamarin/XamarinCommunityToolkit```

jfversluis:
Let's go with this one: http://xamarin.com/schemas/2020/toolkit

Just so we are inline with the rest and it looks super interesting 🤓

XamarinCommunityToolkit/AssemblyInfo/AssemblyInfo.shared.cs
```diff
@@ -1,3 +1,12 @@
...
-[assembly: InternalsVisibleTo("Xamarin.CommunityToolkit.UnitTests")]
\ No newline at end of file
+[assembly: InternalsVisibleTo("Xamarin.CommunityToolkit.UnitTests")]
+
+#if !NETSTANDARD1_0
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2014/forms", "Xamarin.CommunityToolkit.Behaviors")]
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2014/forms", "Xamarin.CommunityToolkit.Converters")]
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2014/forms", "Xamarin.CommunityToolkit.Effects")]
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2014/forms", "Xamarin.CommunityToolkit.Extensions")]
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2014/forms", "Xamarin.CommunityToolkit.UI.Views")]
```
AndreiMisiukevich, line 11:
```suggestion
[assembly: XmlnsDefinition("http://xamarin.com/schemas/2020/toolkit", "Xamarin.CommunityToolkit.Behaviors")]
[assembly: XmlnsDefinition("http://xamarin.com/schemas/2020/toolkit", "Xamarin.CommunityToolkit.Converters")]
[assembly: XmlnsDefinition("http://xamarin.com/schemas/2020/toolkit", "Xamarin.CommunityToolkit.Effects")]
[assembly: XmlnsDefinition("http://xamarin.com/schemas/2020/toolkit", "Xamarin.CommunityToolkit.Extensions")]
[assembly: XmlnsDefinition("http://xamarin.com/schemas/2020/toolkit", "Xamarin.CommunityToolkit.UI.Views")]
```

XamarinCommunityToolkit/AssemblyInfo/AssemblyInfo.shared.cs
```diff
@@ -1,3 +1,12 @@
...
-[assembly: InternalsVisibleTo("Xamarin.CommunityToolkit.UnitTests")]
\ No newline at end of file
+[assembly: InternalsVisibleTo("Xamarin.CommunityToolkit.UnitTests")]
+
+#if !NETSTANDARD1_0
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2020/toolkit", "Xamarin.CommunityToolkit.Behaviors")]
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2020/toolkit", "Xamarin.CommunityToolkit.Converters")]
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2020/toolkit", "Xamarin.CommunityToolkit.Effects")]
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2020/toolkit", "Xamarin.CommunityToolkit.Extensions")]
+[assembly: XmlnsDefinition("http://xamarin.com/schemas/2020/toolkit", "Xamarin.CommunityToolkit.UI.Views")]
```
AndreiMisiukevich, line 11: @jfversluis Why are **Views** inside **UI**, but other aren't?

AndreiMisiukevich:
> Let's go with this one: http://xamarin.com/schemas/2020/toolkit
> 
> Just so we are inline with the rest and it looks super interesting 🤓

updated

XamarinCommunityToolkit/AssemblyInfo/AssemblyInfo.shared.cs
jfversluis: Because they are not visible, I guess 😅

AndreiMisiukevich: Should we make all namespaces consistent? (With **UI** or w/out)

dansiegel: if anything I would say drop the views... everything in UI.Views is a VisualElement... Behaviors, Converters, Effects, Extensions are all helpers but none of them are Visual Elements

jfversluis:
Whooo boy! That are some conflicts :D

Sorry, did some work on the sample app... Let me know if you have the time and motivation to fix it, else I'll pick it up :)

dansiegel:
wow... that list sure increased from when I started rebasing... I'll take care of it

dansiegel:
@jfversluis alright everything is rebased and should be good to go now... for the love of all things holy... please be sure to merge this before making more changes to the sample app, or at least get me some more of that delicious Belgian beer and chocolate 🤣

jfversluis:
Will do both! :D 

Yeah I wanted to merge this actually when I noticed that my change made a bazillion conflicts with this one. But I'm keeping my eye on the build progress and merge this in when it all looks good.

Thanks for this!

roubachof:
Guys, be aware that to be able to consume types from the custom namespace, compiler requires that there's a code reference from the assembly that consumes the types, to the assembly that defines the types.

It means that we need to create an Initializer, like `XamarinCommunityToolkit.Initializer.Initialize()`.

Source is here: https://docs.microsoft.com/en-us/xamarin/xamarin-forms/xaml/custom-namespace-schemas#consuming-a-custom-namespace-schema

I just want to warn you cause I ran into that issues with Shadows: https://github.com/roubachof/Sharpnado.Shadows/issues/24

dansiegel:
@roubachof not necessarily... keep in mind that the Community Toolkit multi-targets so you have one output assembly for each target, where Sharpnado uses an older style one assembly per target plus a shared assembly. The result is if someone simply uses anything from the Toolkit such as the AsyncCommand, ObservableRangeCollection, etc it would have the same effect.

Also we could simply add according to that article. Xamarin.Forms itself does this, and ultimately there are a number of reasons why you shouldn't be trying to link out more than the SDK.

```cs
[assembly: Preserve]
```

roubachof:
Multi-target doesn't seem to change anything to the described issue:
Well if someone only use `Behaviors`, `Views` or anything just Xaml based and the toolkit has no code initializer, the user will end up with the same issue...

jfversluis:
Hm if that's the case, that's not ideal.

On the other hand, what happens now? We say we are linker safe, if we only have a reference in XAML that should mean that everything still works, right? Does that change when we implement this? 

I guess it's a matter of verifying

roubachof:
This has nothing to do with the linker here, you will be still linker safe.
It's just a Xaml compiler thing: you need to have a code reference towards the assembly which has custom xml namespace.
Or else:

> Failure to include such a code reference will result in the XAML compiler being unable to locate the assembly containing the custom namespace schema types.

dansiegel:
What this boils down to is... is there a reference to the Xamarin Community Toolkit in code or not... So let's say that we have:

```xml
             xmlns:xct="http://xamarin.com/schemas/2020/toolkit"
             x:Class="XmlnsTest1.MainPage">
  <Image Source="{xct:GravatarImage Email='johndoe@gmail.com'}" />
</ContentPage
```

If this is our only reference to the Toolkit it would get stripped out... now let's say that somewhere else in my code which could be within the same project or a referenced project... let's say that I have:

```cs
public class ViewAViewModel : ObservableObject
{
}
```

Well now I have a reference to a type in code that comes from the Xamarin Community Toolkit assembly and the linker doesn't pull it out before XamlC runs... We could however add a dummy class like:

```cs
public static class CommunityToolkit
{
    public static void Init()
    {
        // Intentionally left blank... stupid linker...
    }
}
```

That way if someone is only using things within XAML and never has a reference in code, there is a simple dummy method that would keep the linker from pulling the reference. This would be used like:

```cs
public partial class App : Application
{
    public App()
    {
        CommunityToolkit.Init();
    }
}
```

pictos:
@dansiege our goal is to don't have this kind of `init` API @jfversluis, please correct me if I'm saying something wrong.

That `[assembly: Preserve]` seems good, we can try this first and see how it goes.

dansiegel:
@pictos after testing I'm not sure that will have the desired effect. I don't think it's the Mono linker that's really the issue here. If there is zero code reference in C# then you lose the reference to the assembly. 

There is something we could do like I did for the Prism Container Extensions where we introduce an assembly attribute that comes from the toolkit. It's totally transparent but can work as enough of a reference to keep you from losing the assembly. My tests showed if I used something like ObservableObject or ObservableRangeCollection then no omit is needed. Hope that makes sense.

pictos:
> There is something we could do like I did for the Prism Container Extensions where we introduce an assembly attribute that comes from the toolkit. The attribute would be added via a hidden build prop.

Is there a code or sample that we can take a look at?

dansiegel:
> our goal is to don't have this kind of init API

@pictos again this assumes that there is an opposition to having an empty Init somewhere. In that case I would do it a little differently here than what I had done in the [Prism Container Extensions](https://github.com/dansiegel/Prism.Container.Extensions/blob/7.2.0/src/Prism.DryIoc.Extensions/build/Prism.DryIoc.Extensions.targets). A little bit of hindsight, experience and just different goals. Remember that the Toolkit is currently marked LinkerSafe so we probably don't want to mess with that. The goal here is to ensure just enough of a reference to keep the reference to the assembly itself alive for XamlC if someone only ever uses types from the toolkit in XAML and not in code. If they have ANY reference whatsoever in code it preserves the reference for XamlC to function.

In the Toolkit we could add this hidden attribute. Note that it's not actually a preserve attribute, we just want something simple to reference.
```cs
[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = false)]
public sealed class PreserveToolkitAttribute : Attribute
{
}
```

Then what we would do is add a build directory to the nuget that would have a targets file and a code file like:

```cs
// PreserveToolkit.cs
[assembly: Xamarin.CommunityToolkit.PreserveToolkit]
```

```xml
<Project>
  <ItemGroup>
    <Compile Include="$(MSBuildThisFileDirectory)PreserveToolkit.cs" />
  </ItemGroup>
</Project>
```

What this would do is transparently add the attribute to the consuming assembly which would ensure that it doesn't get removed before XamlC runs.

pictos:
@dansiegel thank you for this info, today I learned one more thing! I'll take a look at that tomorrow since is very late here. I hope that we can me this work with this attribute.

dansiegel:
@jfversluis tested the CI build and it works as expected. Install the NuGet, and the Xmlns works without any references in C# to the community toolkit. Just made a small tweak to the targets to prevent the PreserveXamarinCommunityToolkit.cs file from showing up as a linked file in the IDE. This is ready to be merged :)

pictos:
@dansiegel I'll test this and review it ASAP.
Thanks for this.

XamarinCommunityToolkitSample/Pages/Views/AvatarViewPage.xaml
```diff
@@ -22,7 +21,7 @@
                                      Orientation="Horizontal"
                                      Spacing="10">
 
-                            <views:AvatarView ColorTheme="{x:Static views:ColorTheme.Jungle}"
+                            <xct:AvatarView ColorTheme="{x:Static xct:ColorTheme.Jungle}"
```
pictos, line 24: You need to change the `views` to `xct`
```xml
    <views:AvatarView.Source>
            <UriImageSource Uri="{Binding Source}" />
    </views:AvatarView.Source>
</views:AvatarView>

pictos (changes requested):
The building is failing, the reason is in my comment bellow

XamarinCommunityToolkitSample/Pages/Views/AvatarViewPage.xaml
dansiegel: good catch... not sure how my local build didn't fail

pictos (approved):
Tested on my end (VSWindows - Android, iOS, UWP, WPF) and works fine. Please test on your end as well.

The UWP project, for some reason, shows blank pages when we try to navigate to any category (Views, Behaviors, etc) but isn't caused by this PR, it's reproducible in the main branch.

jfversluis (approved):

Issue #348, opened 2020-10-01
[Spec] FormattedText Attached Effect for Entry


ahoefling:
# FormattedText Attached Effect for Entry
Add a `FormattedText` attached effect for the `Entry` control. This will allow the Text of an Entry to use `Span` controls to specify custom styling.

# API

## FormattedTextEffect
The `FormattedTextEffect` is an attached effect that applies the formatting to the native `Entry` control.

### Properties

| API | Description |
| ------------- | ------------- |
| FormattedText | The `FormattedString` that will adjust the styling of the text inside of the `Entry` control. |

### Events
Not applicable

# Scenarios
Given that there is an `Entry` control, adding the attached `FormattedTextEffect` property you will be able to assign the formatting just like you can with the `FormattedText` property on a standard `Label` control. 

Performance is very important for this effect, as I think it will be common to use a Binding to update the `FormattedText`. This will give developers a way to ensure their formatting is correct with each in character in the `Entry`

## Rules
- If `FormattedText` is assigned it will override the `Text` property
- When a user types into the `Entry` it will ensure that `Text` and `FormattedText` remain synchronized if the `FormattedText` is not null
- When a user types into the `Entry` where `FormattedText` is not null it will assume the formatting from the last `Span` object

## Code Samples - Simple Example
```xaml
<Entry>
    <local:FormattedTextEffect.FormattedText>
        <FormattedString>
            <Span Text="Hello " TextColor="Orange" />
            <Span Text="World" TextColor="Black" />
        </FormattedString>
    </local:FormattedTextEffect.FormattedText>
</Entry>
```

## Code Sample - ViewModel Example
This specification is being submitted as something I recently have built. This example is how I am using it in production today. Suppose you want to update the `FormattedText` from your ViewModel to define the segments and a converter to manage the UI controls.

```c#
public class MyViewModel
{
    public string[] MyText => new { "Hello ", "World" };
}
```

```c#
public class StringArrayToFormattedTextConverter : IValueConverter
{
    public object ConvertTo(object value, Type targetType, object parameter, CultureInfo culture)
    {
        // This hard-coded to assume that `value` is a string[] with 2 segments. This is done
        // as an example to showcase the technique.
        string[] text = (string[])value;
  
        FormattedText formattedText = new FormattedText();
        formattedText.Spans.Add(new Span
        {
            TextColor = Color.Orange,
            Text = text[0]
        });

        formattedText.Spans.Add(new Span
        {
[...]

PR #360, opened 2020-10-03
[Android] Adds EntryFormattedText Effect

ahoefling:
### Description of Change ###
Adds a FormattedText Effect that can be applied as an attached property to `<Entry>` controls. This is the Android only implementation as I want to focus on getting the shared code correct prior to implementing it any further.

### Bugs Fixed ###

- Fixes: #348 - Android and shared code implementation

### API Changes ###
```csharp
public static class FormattedTextEffect
{
    public static FormattedString GetFormattedText(BindableObject view);
    public static void SetFormattedText(BindableObject view, FormattedString value);
}
```

The `FormattedTextEffect` is implemented as an attached property. This means it can be directly attached at the root level of the `<Entry>` control. 

```xaml
<Entry>
    <xct:FormattedTextEffect.FormattedText>
        <FormattedString>
            <Span Text="Hello " TextColor="Red" />
            <Span Text="World" TextColor="Purple" />
        </FormattedString>
    </xct:FormattedTextEffect.FormattedText>
</Entry>
```

### Behavioral Changes ###
This is a new feature to the Xamarin Community Toolkit

The EntryFormattedText Effect allows developers to add text formatting to the `<Entry>` control. For example a developer can have multiple text colors in a string and update them in place. See the gif under the **screenshots** section

### Screenshots ###
entry-formatted-text

### PR Checklist ###

### DIFF
--- a/XamarinCommunityToolkit.UnitTests/Xamarin.CommunityToolkit.UnitTests.csproj
+++ b/XamarinCommunityToolkit.UnitTests/Xamarin.CommunityToolkit.UnitTests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFrameworks>netcoreapp3.1;net461</TargetFrameworks>
 
     <IsPackable>false</IsPackable>
   </PropertyGroup>
--- a/XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
+++ b/XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
@@ -54,9 +54,7 @@ namespace Xamarin.CommunityToolkit.Android.Effects
 			base.OnElementPropertyChanged(args);
 
 			if (args.PropertyName == FormattedTextEffect.FormattedTextProperty.PropertyName)
-			{
 				UpdateFormattedText();
-			}
 		}
 
 		void OnTextChanged(object sender, global::Android.Text.TextChangedEventArgs args)
--- a/XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
+++ b/XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
@@ -2,6 +2,7 @@
 using Android.Text;
 using Android.Text.Style;
 using Android.Widget;
+using Xamarin.CommunityToolkit.Android.Effects;
 using Xamarin.CommunityToolkit.Effects;
 using Xamarin.Forms;
 using Xamarin.Forms.Platform.Android;
@@ -9,7 +10,7 @@ using Xamarin.Forms.Platform.Android;
 [assembly: ResolutionGroupName(nameof(Xamarin.CommunityToolkit))]
 [assembly: ExportEffect(typeof(EntryFormattedTextEffect), nameof(EntryFormattedTextEffect))]
 
-namespace Xamarin.CommunityToolkit.Effects
+namespace Xamarin.CommunityToolkit.Android.Effects
 {
 	/// <summary>
 	/// The FormattedText <see cref="PlatformEffect"/> specific
--- a/XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
+++ b/XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
@@ -1,0 +1,109 @@
+﻿using System.ComponentModel;
+using Android.Text;
+using Android.Text.Style;
+using Android.Widget;
+using Xamarin.CommunityToolkit.Effects;
+using Xamarin.Forms;
+using Xamarin.Forms.Platform.Android;
+
+[assembly: ResolutionGroupName(nameof(Xamarin.CommunityToolkit))]
+[assembly: ExportEffect(typeof(EntryFormattedTextEffect), nameof(EntryFormattedTextEffect))]
+
+namespace Xamarin.CommunityToolkit.Effects
+{
+	/// <summary>
+	/// The FormattedText <see cref="PlatformEffect"/> specific
+	/// to <see cref="Entry"/> controls.
+	/// </summary>
+	public class EntryFormattedTextEffect : PlatformEffect
+	{
+		/// <summary>
+		/// Gets the native <see cref="FormsEditText"/> control.
+		/// </summary>
+		public new FormsEditText Control => (FormsEditText)base.Control;
+
+		/// <summary>
+		/// Gets the Xamarin.Forms <see cref="Entry"/> element.
+		/// </summary>
+		public new Entry Element => (Entry)base.Element;
+
+		/// <inheritdoc />
+		protected override void OnAttached()
+		{
+			UpdateFormattedText();
+		}
+
+		/// <inheritdoc />
+		protected override void OnDetached()
+		{
+		}
+
+		/// <inheritdoc />
+		protected override void OnElementPropertyChanged(PropertyChangedEventArgs args)
+		{
+			base.OnElementPropertyChanged(args);
+
+			if (args.PropertyName == FormattedTextEffect.FormattedTextProperty.PropertyName)
+			{
+				UpdateFormattedText();
+			}
+		}
+
+		/// <summary>
+		/// Updates the native <see cref="FormsEditText"/> with the
+		/// Text or FormattedText. If the FormattedText is null the
+		/// Control defaults to using Text property. Otherwise it uses
+		/// the custom Spans in FormattedText.
+		/// </summary>
+		void UpdateFormattedText()
+		{
+			var selectionStart = Control.SelectionStart;
+			var selectionEnd = Control.SelectionEnd;
+			var formattedText = FormattedTextEffect.GetFormattedText(Element);
+
+			if (formattedText == null)
+			{
+				Control.Text = Element.Text;
+			}
+			else
+			{
+				var text = new SpannableStringBuilder();
+				var position = 0;
+				for (var index = 0; index < formattedText.Spans.Count; index++)
+				{
+					var currentSpan = formattedText.Spans[index];
+					text.Append(currentSpan.Text);
+
+					var nativeSpan = CreateNativeSpan(currentSpan);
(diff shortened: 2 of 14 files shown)


ahoefling:
This is being submitted early to get feedback on how the shared code API looks and the Android API works. As of right now I understand the following things are missing
- Only supports Changing Text Color

XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
```diff
@@ -0,0 +1,162 @@
...
+			if (HasSpan())
+			{
+				var startPosition = spannable.GetSpanStart(findSpans[0]);
+				var endPosition = spannable.GetSpanEnd(findSpans[0]);
+
+				spannable.SetSpan(findSpans[0], startPosition, endPosition, SpanTypes.ExclusiveExclusive);
+			}
+
+			bool HasSpan() => findSpans != null & findSpans.Length > 0;
+		}
```
ahoefling, line 77: The rules here are kind of complicated and it still needs a lot of work to get right. Currently this algorithm works really good if you are inserting text in the middle of a `<Span>` or at the end of a `<Span>`. If your spans collide as in there is no spacing between them odd behaviors happen.

I am open to suggestions on this, but I think this is a great start. I am highlighting this part of the code as this is a key item to get right as in the specification we said we wanted to retain the formatting if the user inserts or removes characters from a span

ahoefling: I was looking at adding the different span types after submitting this and I believe this part needs to be re-written. My original approach here used the following rules
1. Attempt to get the closest first span
2. Apply that span to the spannable

This solution worked okay for the happy path of just TextColor, but once I added all the other properties I don't think it will work well and will become very complicated. 

I am thinking of creating a new instance level property on this class that will store all the spans and their native objects. Once the text property has updated I can then easily retrieve all spans related to the segment and invoke `spannable.SetSpan()` for each of those spans. I think this new proposed solution will be easier for us to manage and won't have as many problems as the original event handler only solution

XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
```diff
@@ -0,0 +1,162 @@
...
+
+		/// <inheritdoc />
+		protected override void OnElementPropertyChanged(PropertyChangedEventArgs args)
+		{
+			base.OnElementPropertyChanged(args);
+
+			if (args.PropertyName == FormattedTextEffect.FormattedTextProperty.PropertyName)
+			{
+				UpdateFormattedText();
```
pictos, line 53: When we have just one call, we don't use brackets... I still working to make our analyzer better, looks like editorconfig is being ignored by VS.

pictos (commented):
So far so good. I'm still trying to see how it works with VM Bindings. Like this will be trigged every time the user types something, right? And since we are using strings this can allocate a lot of strings, right? I never used before on Xamarin world but maybe this can be a good place to try `Span<string>`

ahoefling:
This is exactly the feedback I was looking for. We really should be using Span<string> for this.

XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
ahoefling: My mistake, I thought stylecop told me otherwise.

I'll update this, trying to make sure I'm 100% compliant with the coding standard

pictos:
@ahoefling Also, if we can make an incredible performance feature, can we expand this to ʻEditor`? I can imagine a markdown (or something similar) editor as we have on Whatsapp or Discord using this feature.

ahoefling:
That is why I am naming the effect `FormattedTextEffect` and naming the platform implementation `EntryFormattedTextEffect`.

Once we have this merged, I want to do something very similar for the `<Editor>` control. I know the Android implementation will be trivial at that point since it uses the same native control. Not sure about other platforms

XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
pictos: Don't worry, the stylecop doesn't have a definition for this case (I didn't find so far), this one should be `editorconfig` responsibility, but he has not been working well.

ahoefling: I'm going to see if I can find the stylecop setting for this. I've seen it in other projects I've worked on. I'll DM you about it if I find it

jfversluis:
Adding do not merge until you finish the implementation

ahoefling:
@jfversluis should we be doing merge commits or rebases to keep this branch up to date? I have been doing rebases and was planning to do a force-update the next time I update this PR

XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
```diff
@@ -0,0 +1,162 @@
...
+			}
+		}
+
+		void OnTextChanged(object sender, global::Android.Text.TextChangedEventArgs args)
+		{
+			var spannable = (SpannableStringBuilder)args.Text;
```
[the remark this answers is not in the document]
ahoefling: This makes sense, thanks for the suggestion. I'll update this on the next code push

jfversluis:
Yeah My bad. The CI had been stuck over the weekend and wanted to give it a nudge and pushed the GitHub update branch button here. Ideally just rebase, but we'll squash when merging anyway so it doesn't all matter that much :)

ahoefling:
No problem, I wanted to check as my local repo is a few commits ahead. I have been tinkering all weekend on the `OnTextChanged` algorithm and how we can keep things synchronized with the shared objects. Not an easy problem to solve.

When I make my next update I'll be sure it is all up to date and force push

XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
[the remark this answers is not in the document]
ahoefling: This has been updated in the latest code push

XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
ahoefling: This algorithm has been completely re-written to use shared code as much as possible. Let's open a new discussion thread

XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
ahoefling: This has been fixed in the latest code push

XamarinCommunityToolkit.UnitTests/Xamarin.CommunityToolkit.UnitTests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;net461</TargetFrameworks>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
ahoefling, line 4: This must have been part of a bad rebase commit. This needs to be fixed

XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
```diff
@@ -0,0 +1,258 @@
...
+				item.StartIndex = position;
+				position += item.Span.Text.Length;
+			}
+
+			foreach (var nativeSpan in segment.NativeSpans)
+			{
+				spannable.RemoveSpan(nativeSpan);
+				spannable.SetSpan(nativeSpan, segment.StartIndex, segment.StartIndex + segment.Span.Text.Length, SpanTypes.ExclusiveExclusive);
+			}
+		}
```
ahoefling, line 104: This algorithm has been re-written from the original PR with a focus on managing the state via shared code and only applying the spans to native objects when needed. I am hoping we can extract the rules in this algorithm to use a shared object across the different platform implementations. If we can do that it means the business rules will be consistent and much easier to maintain then N-number of implementations.

The rules are documented in the linked issue

XamarinCommunityToolkit.UnitTests/Xamarin.CommunityToolkit.UnitTests.csproj
ahoefling: fixed in latest code push

ahoefling:
I have updated the PR to include a new algorithm for managing the Span objects in a non-bindable property scenario. The new algorithm uses a shared code approach that I hope can be extracted into a shared class that could be used in an iOS and UWP implementation.

@pictos and anyone else interested, can you please take another look and provide any feedback. Once we go through another round of review on the non-bindable property approach, I plan to document how the bindable property works. The bindable property technique with this is much easier, which is why I am saving it for the end 😁

XamarinCommunityToolkit/Effects/EntryFormattedTextEffect.android.cs
```diff
@@ -0,0 +1,258 @@
...
+		protected override void OnElementPropertyChanged(PropertyChangedEventArgs args)
+		{
+			base.OnElementPropertyChanged(args);
+
+			if (args.PropertyName == FormattedTextEffect.FormattedTextProperty.PropertyName)
+				UpdateFormattedText();
+		}
+
+		void OnTextChanged(object sender, global::Android.Text.TextChangedEventArgs args)
+		{
```
pictos, line 61: The code looks very code to me! I found just one place that we can improve perf a little bit. Accessing prop values directly is bad, we can use a local variable instead. So in this case `args.Start` is used more than once, so would be good to do something like `var argStart = args.Start` and use this `argStart` in the code. Since this method will be called a lot, this can have a good impact.
Yesterday night I tried to set up a benchmark for this feature but without success. I'll try again in the next days and share it with you.
And awesome work!

ahoefling: Thank you 🙏 for taking a look at this. Anything we can do to improve the overall performance of this is important to get right in this algorithm.

jsuarezruiz:
@ahoefling can you fix these merge conflicts? After that, I'll take a look at this PR!

brminnick:
Closing PR as it has been abandoned.

I anybody is interested in taking over this PR we are more than happy to re-open it!

PR #382, opened 2020-10-06
Implemented the BindablePropertyBuilder (#365)

### LINKED ISSUE #365: [Enhancement] Fluent Builder Pattern For Bindable Properties
## Summary

I'm personally not a fan of the `BindableProperty.Create` method with its many optional parameters - it is VERY difficult to figure out which parameters you need, which ones you have omitted, etc. I think using a fluent builder pattern would make it easier to create a new `BindableProperty`.

## API Changes

~New class: `BindablePropertyParameters`:~ Edit: removed

~And~ `BindablePropertyBuilder`:

```csharp
public class BindablePropertyBuilder
{
    private Type declaringType;
    private BindingMode defaultBindingMode;
    private object defaultValue;
    private string propertyName;
    private Type returnType;
    
    private BindableProperty.BindingPropertyChangedDelegate bindingPropertyChangedDelegate;
    private BindableProperty.BindingPropertyChangingDelegate bindingPropertyChangingDelegate;
    private BindableProperty.CoerceValueDelegate coerceValueDelegate;
    private BindableProperty.CreateDefaultValueDelegate createDefaultValueDelegate;
    private BindableProperty.ValidateValueDelegate validateValueDelegate;

    public BindablePropertyBuilder();
    public BindablePropertyBuilder SetReturnType(Type returnType);
    public BindablePropertyBuilder SetPropertyName(string propertyName);
    public BindablePropertyBuilder SetDefaultValue(object defaultValue);
    public BindablePropertyBuilder SetDefaultBindingMode(BindingMode bindingMode);
    public BindablePropertyBuilder SetDeclaringType(Type declaringType);

    public BindablePropertyBuilder SetValidateValueDelegate(BindableProperty.ValidateValueDelegate validateValueDelegate);
    public BindablePropertyBuilder SetPropertyChangedDelegate(BindableProperty.BindingPropertyChangedDelegate propertyChangedDelegate);
    public BindablePropertyBuilder SetPropertyChangingDelegate(BindableProperty.BindingPropertyChangingDelegate propertyChangingDelegate);
    public BindablePropertyBuilder SetCoerceValueDelegate(BindableProperty.CoerceValueDelegate coerceValueDelegate);
    public BindablePropertyBuilder SetCreateDefaultValueDelegate(BindableProperty.CreateDefaultValueDelegate createDefaultValueDelegate);

    public BindableProperty Build();
}
```

## Intended Use Case
Simply put, instead of using the current `Create` method

```csharp
public static readonly BindableProperty MyBindableProperty = BindableProperty.Create(nameof(MyProperty), typeof(string), typeof(MyControl), string.Empty, BindingMode.TwoWay, OnMyPropertyChanged)
```

developers can use a fluent builder syntax

```csharp
public static readonly BindableProperty MyBindableProperty = new BindablePropertyBuilder()
    .SetPropertyName(nameof(MyProperty))
    .SetDeclaringType(typeof(MyControl))
    .SetReturnType(typeof(string))
    .SetDefaultValue(string.Empty)
    .SetDefaultBindingMode(BindingMode.TwoWay)
    .SetPropertyChangedDelegate(OnMyPropertyChanged)
    .Build();
```

## Who Will Do The Work?

- [x] I am willing to take this on myself (I have already implemented this on a personal project)

aguevara716:
### Description of Change ###

Added support for a fluent builder syntax when creating new instances of `BindableProperty`.

### Bugs Fixed ###

- Fixes #365 

### API Changes ###

Added: 
 
- `public class BindablePropertyBuilder`
- `internal class BindablePropertyParameters`

### Behavioral Changes ###

None - this is an optional change

### PR Checklist ###


XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.cs
```diff
@@ -0,0 +1,161 @@
...
+		/// <summary>
+		/// (Required) Sets the type of the value that is bound to the <see cref="BindableProperty"/>
+		/// </summary>
+		/// <param name="returnType">The type of the value that is bound to the <see cref="BindableProperty"/></param>
+		/// <returns>The <see cref="BindablePropertyBuilder"/> to support the fluent syntax</returns>
+		public BindablePropertyBuilder SetReturnType(Type returnType)
+		{
+			bpParameters.ReturnType = returnType;
+			return this;
+		}
```
AndreiMisiukevich, line 21: How about adding

```csharp
public BindablePropertyBuilder SetReturnType<TReturn>()
	=> SetReturnType(typeof(TReturn));
```

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.cs
```diff
@@ -0,0 +1,161 @@
...
+				throw new ArgumentNullException(nameof(bpParameters.PropertyName), $"{nameof(bpParameters.PropertyName)} is required");
+			if (bpParameters.ReturnType == default)
+				throw new ArgumentNullException(nameof(bpParameters.ReturnType), $"{nameof(bpParameters.ReturnType)} is required");
+			if (bpParameters.DeclaringType == default)
+				throw new ArgumentNullException(nameof(bpParameters.DeclaringType), $"{nameof(bpParameters.DeclaringType)} is required");
+
+			if (bpParameters.DefaultValue != null && bpParameters.ReturnType != bpParameters.DefaultValue.GetType())
+				throw new ArgumentException($"The return type is {bpParameters.ReturnType.Name}, but the default value is of type {bpParameters.DefaultValue.GetType().Name}");
+
+			var bp = BindableProperty.Create(propertyName: bpParameters.PropertyName,
```
AndreiMisiukevich, line 148: ```return BindableProperty.Create(...``` ?

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.cs
```diff
@@ -0,0 +1,161 @@
...
+		/// </exception>
+		public BindableProperty Build()
+		{
+			if (string.IsNullOrEmpty(bpParameters.PropertyName))
+				throw new ArgumentNullException(nameof(bpParameters.PropertyName), $"{nameof(bpParameters.PropertyName)} is required");
+			if (bpParameters.ReturnType == default)
+				throw new ArgumentNullException(nameof(bpParameters.ReturnType), $"{nameof(bpParameters.ReturnType)} is required");
+			if (bpParameters.DeclaringType == default)
+				throw new ArgumentNullException(nameof(bpParameters.DeclaringType), $"{nameof(bpParameters.DeclaringType)} is required");
+
```
AndreiMisiukevich, line 144: should we remove this space?

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.cs
```diff
@@ -0,0 +1,161 @@
...
+		/// and <see cref="BindableProperty.ReturnType"/> don't match.
+		/// </exception>
+		public BindableProperty Build()
+		{
+			if (string.IsNullOrEmpty(bpParameters.PropertyName))
+				throw new ArgumentNullException(nameof(bpParameters.PropertyName), $"{nameof(bpParameters.PropertyName)} is required");
+			if (bpParameters.ReturnType == default)
+				throw new ArgumentNullException(nameof(bpParameters.ReturnType), $"{nameof(bpParameters.ReturnType)} is required");
+			if (bpParameters.DeclaringType == default)
+				throw new ArgumentNullException(nameof(bpParameters.DeclaringType), $"{nameof(bpParameters.DeclaringType)} is required");
```
AndreiMisiukevich, line 143: can we extract it into the separate method?

```
void CheckParameter<TParameter> (TParameter parameter, Predicate<TParameter> predicate, string parameterName) {
    if (!predicate.Invoke(parameter))
        throw new ArgumentNullException(parameterName, $"{parameterName} is required";
}
```

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.cs
```diff
@@ -0,0 +1,161 @@
...
+
+			if (bpParameters.DefaultValue != null && bpParameters.ReturnType != bpParameters.DefaultValue.GetType())
```
AndreiMisiukevich, line 145: just ```bpParameters.ReturnType != bpParameters.DefaultValue?.GetType()```

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.cs
```diff
@@ -0,0 +1,161 @@
+﻿using System;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Helpers
+{
+	public class BindablePropertyBuilder
```
AndreiMisiukevich, line 6: Should we make it "sealed"?

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.cs
aguevara716: It's a habit of mine - when I'm debugging, it's easier to see what I'm returning. I can change it if you all really want me to, it's just my preference to toss it into a variable and then return the variable.

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.cs
```diff
@@ -0,0 +1,177 @@
...
+			bpParameters.DeclaringType = declaringType;
+			return this;
+		}
+
+		/// <summary>
+		/// (Required) Sets the type (usually a control) that's declaring the <see cref="BindableProperty"/>
+		/// </summary>
+		/// <typeparam name="TReturn">The type (usually a control) that's declaring the <see cref="BindableProperty"/></typeparam>
+		/// <returns>The <see cref="BindablePropertyBuilder"/> to support the fluent syntax</returns>
+		public BindablePropertyBuilder SetDeclaringType<TReturn>() => SetDeclaringType(typeof(TReturn));
```
AndreiMisiukevich, line 79:
```suggestion
		public BindablePropertyBuilder SetDeclaringType<TDeclaring>() => SetDeclaringType(typeof(TDeclaring));
```

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.cs
```diff
@@ -0,0 +1,177 @@
...
+		/// If the type of the <see cref="BindableProperty.DefaultValue"/>
+		/// and <see cref="BindableProperty.ReturnType"/> don't match.
+		/// </exception>
+		public BindableProperty Build()
+		{
+			CheckParameter(!string.IsNullOrEmpty(bpParameters.PropertyName), nameof(bpParameters.PropertyName));
+			CheckParameter(bpParameters.ReturnType != default, nameof(bpParameters.ReturnType));
+			CheckParameter(bpParameters.DeclaringType != default, nameof(bpParameters.DeclaringType));
+			if (bpParameters.ReturnType != bpParameters.DefaultValue?.GetType())
+				throw new ArgumentException($"The return type is {bpParameters.ReturnType.Name}, but the default value is of type {bpParameters.DefaultValue.GetType().Name}");
```
AndreiMisiukevich, line 156: This check is wrong.

First of all ```bpParameters.DefaultValue.GetType()``` can cause null reference exception before raising argument exception.

But DefaultValue can be null, it means that it's impossible to get its type.
Or even if somebody decides to use DefaultValueDelegate instead.

I see two possible approaches for handling this case.

1) Just removing these two lines
2) Ori if we want to check the type before creating, we will have to make BindablePropertyBuilder generic type

```new BindablePropertyBuilder<TDeclaring, TReturn>()```

File: XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj [hunk 9]
```diff
@@ -67,6 +67,12 @@
     <Compile Include="**/*.shared.*.cs" />
     <None Include="..\LICENSE" PackagePath="" Pack="true" />
     <None Include="..\assets\XamarinCommunityToolkit_128x128.png" PackagePath="icon.png" Pack="true" />
+    <Compile Include="Helpers\BindablePropertyParameters.cs">
+      <SubType></SubType>
+    </Compile>
+    <Compile Include="Helpers\BindablePropertyBuilder.cs">
+      <SubType></SubType>
+    </Compile>
```
AndreiMisiukevich, line 75: Please fix file names and discard this file (This file should be preserved non-changed)

File names should be ```BindablePropertyParameters.shared.cs``` and ```BindablePropertyBuilder.shared.cs```

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.cs
AndreiMisiukevich: up to you then 👍

AndreiMisiukevich (changes requested):
Please address my comments 👍

pictos:
@aguevara716, I want to suggest that you implement an UnitTests for this feature since it's a _new thing_ and - probably - we will change our internal BPs to use this new pattern.

Also, personally, I didn't like this idea `new BindablePropertyBuilder()`, can we have a static method create it, like the old `BindableProperty.Create`? My point here is that we have other types of BPs and we should support all(?)

Types:
- `Create`
- `CreateReadOnly`
- `CreateAttached`
- `CreateAttachedReadOnly`

aguevara716:
@pictos would you kindly share an example of how you'd envision using this?

pictos:
@aguevara716, is the `this` related to the `new BindablePropertyBuilder()`?

aguevara716:
@pictos yes, sorry, I was referring to replacing the constructor with a static method

pictos:
@aguevara716 , to me the builder should be a singleton. I don't _feel_ well creating a new Builder object per BP. 

```csharp
// Implementation

class Builder
{
    static Lazy<Builder> lazyBuilder = new Lazy<Builder>(() => new Builder());
    public static Builder Instace => Builder.Value;

    Builder()
    {
    }
}

// usage

class MyUsage
{
    public BindableProperty MyProp = Builder.Instance.SetName().SetType();
}
```

In this way, we can easily add the overlay to other BPs types

Types:
- `Create`
- `CreateReadOnly`
- `CreateAttached`
- `CreateAttachedReadOnly`

aguevara716:
@pictos understood - I had something similar in mind, but I didn't want to make too many assumptions.

My only concern about creating a singleton is that it would be "dirty" between usages (in other words, it would retain the `BindablePropertyParameters` from the previous use). Unless the final `Build()` method resets these parameters.

Alternatively, I was thinking of the following approach:

```csharp
public class BindablePropertyBuilder
{
    readonly BindalbePropertyParameters bpParameters;

    BindablePropertyBuilder() => bpParameters = new BindablePropertyParameters();

    public static BindablePropertyBuilder InitializeWithPropertyName(string propertyName)
    {
        var bpb = new BindablePropertyBuilder();
        bpb.PropertyName = propertyName;
        return bpb;
    }

    // TODO add the rest of the Set*() methods and the Build() method
}
```

pictos:
Sounds good. @aguevara716 I think that we can move this discussion to the issue xD. I'll continue there

XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj
```diff
@@ -158,4 +152,13 @@
...
   </ItemGroup>
+
+  <ItemGroup>
+    <Compile Update="Helpers\BindablePropertyBuilder.shared.cs">
+      <SubType></SubType>
+    </Compile>
+    <Compile Update="Helpers\BindablePropertyParameters.shared.cs">
+      <SubType></SubType>
+    </Compile>
+  </ItemGroup>
```
AndreiMisiukevich, line 163: These lines can be safely removed

ahoefling (commented):
Thank you for the PR, I do like the idea of adding fluent API builders to XCT. I think it will serve a great purpose. I have some questions about the implementation that I noted in my review.

In this PR I only see an implementation for `BindablePropert.Create`. Are you planning to include all the options for BindableProperty in this PR or others? 

My biggest concern is how do you envision maintaining several Bindable Property builders? Since this is only 1 type of Bindable Property builder it would be useful to see how the others would be implemented.
- Separate builder class
- Parameterized API to specify Bindable Property type

Have you ran this through `DotNetBenchmarking` to see if there are any performance concerns we should be aware of?

aguevara716:
@ahoefling 

I was actually considering adding the other `Create` methods to this PR, but realized it's out of scope for the issue that was created. I'm planning to create a new issue where I add support for the other `Create*` methods.

Basically, my initial design would be to create an abstract `BindablePropertyBuilderBase` that defines all of the `Set*` methods and an abstract `Build` method, and have 1 derived class corresponding to each of the `Create` methods.

* `BindablePropertyBuilder`
* `BindablePropertyKeyBuilder`
* `AttachedBindablePropertyBuilder`
* `AttachedBindablePropertyKeyBuilder`

This is just an initial design, though.

ahoefling:
I think it is best to get your specification documented and an API in a GH issue prior to coding. It will give everyone an easy concise way to understand your thoughts before implementation. 

Maybe it makes the most sense to edit the issue with detailed API structure and example use cases

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.shared.cs
```diff
@@ -0,0 +1,180 @@
...
+		/// </exception>
+		/// <exception cref="ArgumentException">
+		/// If the type of the <see cref="BindableProperty.DefaultValue"/>
+		/// and <see cref="BindableProperty.ReturnType"/> don't match.
+		/// </exception>
+		public BindableProperty Build()
+		{
+			CheckParameter(!string.IsNullOrEmpty(bpParameters.PropertyName), nameof(bpParameters.PropertyName));
+			CheckParameter(bpParameters.ReturnType != default, nameof(bpParameters.ReturnType));
+			CheckParameter(bpParameters.DeclaringType != default, nameof(bpParameters.DeclaringType));
```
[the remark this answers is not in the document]
AndreiMisiukevich:
```csharp
if (string.IsNullOrEmpty(bpParameters.PropertyName)
    throw new ArgumentNullException(parameterName, $"{parameterName} is required");

if (bpParameters.ReturnType == null)
    throw new ArgumentNullException(parameterName, $"{parameterName} is required");
```

Do you think this approach is better?

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.shared.cs
```diff
@@ -0,0 +1,180 @@
...
+			if (bpParameters.DefaultValue != null && bpParameters.ReturnType != bpParameters.DefaultValue.GetType())
```
AndreiMisiukevich, line 155: This check is wrong
1) DefaultValue can be null (for reference types)
2) DefaultValueCreator can be used instead of DefaultValue (so default value will be null as well)

ahoefling: I missed this in my review and agree, I actually just built an attached bindable property the other day where it is common to be null

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.shared.cs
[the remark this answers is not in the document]
ahoefling: For an OSS project like this I think it is. The code clearly explains what it is doing where the method name `CheckParameter()` doesn't really explain what is happening. 

When I first glanced at this file I didn't realize the `CheckParameter()` method was throwing an exception and it was hiding logic. It may just be how I read the code file, but if I didn't notice it right away, I'm sure others would have the same problem. I think what you suggested will make this easier to maintain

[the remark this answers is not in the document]
aguevara716: I agree with @ahoefling - I greatly prefer the more verbose and explicit syntax of test-and-throw instead of tucking it away in a method.

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.shared.cs
```diff
@@ -0,0 +1,180 @@
...
+	{
+		BindablePropertyParameters bpParameters = new BindablePropertyParameters();
+
+		/// <summary>
+		/// Creates a new instance of <see cref="BindablePropertyBuilder"/> while setting name of the
+		/// property that is bound to the <see cref="BindableProperty"/>
+		/// </summary>
+		/// <param name="propertyName">The name of the property that is bound to the <see cref="BindableProperty"/></param>
+		/// <returns>The <see cref="BindablePropertyBuilder"/> to support the fluent syntax</returns>
+		public static BindablePropertyBuilder InitializeWithPropertyName(string propertyName)
```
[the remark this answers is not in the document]
aguevara716: I agree, I was just trying this out to make the syntax a little friendlier, but introducing a new way of working with something is not always better or intuitive.

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.shared.cs
```diff
@@ -0,0 +1,180 @@
...
+			var bp = BindableProperty.Create(propertyName: bpParameters.PropertyName,
+											 returnType: bpParameters.ReturnType,
+											 declaringType: bpParameters.DeclaringType,
+											 defaultValue: bpParameters.DefaultValue,
+											 defaultBindingMode: bpParameters.DefaultBindingMode,
+											 validateValue: bpParameters.ValidateValueDelegate,
+											 propertyChanged: bpParameters.BindingPropertyChangedDelegate,
+											 propertyChanging: bpParameters.BindingPropertyChangingDelegate,
+											 coerceValue: bpParameters.CoerceValueDelegate,
+											 defaultValueCreator: bpParameters.CreateDefaultValueDelegate);
```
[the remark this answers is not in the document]
aguevara716: I do want to support creating the other types of `Create` methods, but I feel that is currently out of scope for this task. Once/if this gets merged, I will create a new task to support the rest of the `Create` methods.

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.shared.cs
AndreiMisiukevich: Please remove this check :)

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.shared.cs
```diff
@@ -0,0 +1,192 @@
...
+
+		/// <summary>
+		/// Resets the <see cref="BindablePropertyParameters"/> to its default values and
+		/// returns the <see cref="BindablePropertyBuilder"/> singleton
+		/// </summary>
+		/// <returns>The <see cref="BindablePropertyBuilder"/> singleton</returns>
+		public static BindablePropertyBuilder CreateBuilder()
+		{
+			var bpbSingleton = lazyBindablePropertyBuilder.Value;
+			bpbSingleton.bpParameters.ResetParameters();
```
AndreiMisiukevich, line 22: So we share a single builder among all bindable properties?
Wow, not sure it's a good idea due to threads race condition

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.shared.cs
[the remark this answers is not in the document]
ahoefling: Can you update the title of the PR to document this and mark the description to be `Related to #365`. As this PR is constructed it will automatically close the issue and if this is just 1 step for that issue I don't believe this PR should close that issue

XamarinCommunityToolkit.UnitTests/Helpers/BindablePropertyBuilder_Tests.cs
```diff
@@ -13,7 +13,8 @@ public void InitializeWithPropertyNameShouldSetBindablePropertyPropertyName()
 		{
 			var propertyName = "adsf";
 
-			var bindableProperty = BindablePropertyBuilder.InitializeWithPropertyName(propertyName)
+			var bindableProperty = BindablePropertyBuilder.CreateBuilder()
```
ahoefling, line 16: This is much better and very clear to the developer what they are doing. Thanks for making this change!

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.shared.cs
ahoefling: I don't think this is thread safe, if an app is loading 100s of Bindable Properties at once that exist in different controls I could see multiple threads being generated. Another scenario would be an app aggressively loading all pages of a tabbed dashboard at once.

Adding a simple lock will not solve this concurrency problem

## Additional Questions
- Whey can't we just instantiate a new `BindablePropertyParameters` each time a `CreateBuilder` is invoked? When I `CreateBuilder` I expect to have a fresh clean object.
- Why do we need the `BindablePropertyParameters`? I would prefer if we flattened the structure and just had instance properties in the builder class itself. If there is a good reason for this I can be convinced otherwise

ahoefling (changes requested):
I think this is looking much better. I left some notes and I am in agreement with @AndreiMisiukevich

XamarinCommunityToolkit/Helpers/BindablePropertyBuilder.shared.cs
roubachof: I think we should go back to the main issue #365 to settle this debate on static vs instance and performance.
Till we agree on the best solution, @aguevara716 won't have to reimplement the builder after each comment we make :)

aguevara716: Thanks, @roubachof - I've actually been tempted to close this PR and issue and creating a new issue just so we can start the discussion from scratch.

That being said, I am actually learning a lot from everyone's comments! This is my first time contributing to an open source project, and the first time contributing to something that would be used by many people. I'm used to getting things to work for my use case and moving on, but this is definitely helping me to think more about how other developers would interact with my code.

AndreiMisiukevich:
Thank you for your interest in this project! I appreciate your effort in implementing this PR.
But unfortunately, this PR won't be merged.

If you wish to have more clarity when creating bindable properties, I propose using named parameters like

```csharp
		public static readonly BindableProperty NameProperty = BindableProperty.CreateAttached(
			propertyName: nameof(Name),
			returnType: typeof(bool),
			declaringType: typeof(TouchEffect),
			defaultValue: true,
			propertyChanged: OnNamePropertyChanged);
```

roubachof:
Mmmh. There were a lot of discussion about this PR on the matching issue and @aguevara716 kindly fixed several times its implementation to follow up with our discussions. I don't known when the decision was taken to close it but I guess he deserves a better explanation than just "thank you goodbye".

AndreiMisiukevich:
@roubachof I didn't want to look impolite.
Sorry @aguevara716 if I offended you.

@roubachof I proposed a different approach for creating bindable properties, as you can see above (using named parameter).
If you think it's not sufficient, then we can continue the conversation. I am totally OK

brminnick:
Closing PR as it has been abandoned.

If anybody is interested in taking over this PR we are more than happy to re-open it!

PR #392, opened 2020-10-09
Remove and dispose observer

jessejiang0214:
### Description of Change ###

Need to call Dispose function after removing the observer, also we need to catch any exceptions as there's no way from Xamarin.Forms level to catch the exception when Element changed. 

GetPosition will be called when the avPlayerViewController is null. I got a few crashes from the backlog.

### Bugs Fixed ###

- Fixes # https://github.com/xamarin/XamarinCommunityToolkit/issues/335

### API Changes ###

N/A

### Behavioral Changes ###

N/A
### PR Checklist ###


XamarinCommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
```diff
@@ -191,10 +195,17 @@ TimeSpan Position
...
+					if (avPlayerViewController.Player.CurrentTime.IsInvalid)
+						return TimeSpan.Zero;
 
-				return TimeSpan.FromSeconds(avPlayerViewController.Player.CurrentTime.Seconds);
+					return TimeSpan.FromSeconds(avPlayerViewController.Player.CurrentTime.Seconds);
+				}
+				catch
+				{
+					return TimeSpan.Zero;
+				}
```
AndreiMisiukevich, line 208: Just this.
```csharp
    if (_avPlayerViewController.Player?.CurrentTime.IsInvalid ?? true)
        return TimeSpan.Zero;

    return TimeSpan.FromSeconds(_avPlayerViewController.Player.CurrentTime.Seconds);
```
We should try to find ways for avoiding null-refs rather than suppressing all exceptions.

XamarinCommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
```diff
@@ -134,7 +134,11 @@ protected void RemoveStatusObserver()
 				try
 				{
 					avPlayerViewController?.Player?.CurrentItem?.RemoveObserver(statusObserver, "status");
+					statusObserver.Dispose();
 				}
+				catch
+                {
+                }
```
AndreiMisiukevich, line 141: wrong indentation

XamarinCommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
```diff
@@ -373,8 +384,34 @@ protected override void OnElementChanged(ElementChangedEventArgs<MediaElement> e
...
+				}
+
+				if (rateObserver != null)
+				{
+					try
+					{
+						avPlayerViewController?.Player?.RemoveObserver(rateObserver, "rate");
+						rateObserver.Dispose();
+					}
+					catch
```
AndreiMisiukevich, line 408: What kind of exception do we want t cath here and in other places? Can we specify the type?
Also, I am not fancy of leaving **catch** blocks empty. Can we add a comment at least? "Suppress exception due to ...{reason}"

AndreiMisiukevich:
Hmm, I see that in the original file (In XF) there was an overridden method **Dispose** but now it doesn't exist.

Should we just add this and perform disposing of observers there?

CC @jsuarezruiz

AndreiMisiukevich:
@jessejiang0214 Can I ask you to add a test case inside the "Test Cases" tab for this crash?
I'd like to reproduce this crash myself.
Perhaps moving some lines to "Dispose" method will help us

AndreiMisiukevich:
I moved my PR from XF to XCT repo https://github.com/xamarin/XamarinCommunityToolkit/pull/394

It is not related to your changes, but perhaps some conflicts will appear)

jfversluis:
Just checked, this fix is still needed.

@jessejiang0214 let me know if you are able to finish it over the next couple of days, else I will take over :)

jessejiang0214:
Hi @jfversluis , I was busy on handle customer's issue this week. Maybe I can find some time on next week.

jfversluis:
No problem at all, I totally understand :)

XamarinCommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
jessejiang0214: Hi @AndreiMisiukevich, I just find we cannot catch any exceptions from there, so I remove all of them

jessejiang0214:
@AndreiMisiukevich , I cannot add a test case in UnitTest for this. The way to reproduce this is when you go to the player page, then put the App in the background then bring the App to front ground it will crash.

The crash report said it takes more than 10 seconds to go to the background. 

And you can see my fix I add background mode to the .plist file as the video have to continue playing when App to go the background.

XamarinCommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
```diff
@@ -133,16 +133,9 @@ protected string ResolveMsAppDataUri(Uri uri)
...
-				{
-					avPlayerViewController?.Player?.CurrentItem?.RemoveObserver(statusObserver, "status");
-				}
-				finally
-				{
-					statusObserver = null;
-				}
-			}
+				avPlayerViewController?.Player?.CurrentItem?.RemoveObserver(statusObserver, "status");
+			statusObserver?.Dispose();
```
AndreiMisiukevich, line 137:
```suggestion
			avPlayerViewController?.Player?.CurrentItem?.RemoveObserver(statusObserver, "status");
			statusObserver?.Dispose();
```

AndreiMisiukevich (approved):

XamarinCommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
```diff
@@ -133,16 +133,9 @@ protected string ResolveMsAppDataUri(Uri uri)
...
-				try
-				{
-					avPlayerViewController?.Player?.CurrentItem?.RemoveObserver(statusObserver, "status");
-				}
-				finally
-				{
-					statusObserver = null;
-				}
-			}
+			avPlayerViewController?.Player?.CurrentItem?.RemoveObserver(statusObserver, "status");
```
AndreiMisiukevich, line 136:
```suggestion
				avPlayerViewController?.Player?.CurrentItem?.RemoveObserver(statusObserver, "status");
```

AndreiMisiukevich (approved):

PR #431, opened 2020-10-11
Don't overwrite margin when detaching safe area effect if the effect was not applied

### LINKED ISSUE #323: [Bug] Safe area effect restoring issue

AndreiMisiukevich:
### Description of Change ###
Refactored iOS SafeAreaEffectRouter to not override margin if it was not applied.

### Bugs Fixed ###
- Fixes #323 

### API Changes ###
None

### Behavioral Changes ###
None

- [X] Has tests 
- [X] Has samples


XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs
```diff
@@ -14,34 +14,39 @@ public class SafeAreaEffectRouter : PlatformEffect
...
-
-			if (!UIDevice.CurrentDevice.CheckSystemVersion(11, 0) || !UIApplication.SharedApplication.Windows.Any())
+			if (!IsEligibleToConsumeEffect)
 				return;
 
 			var insets = UIApplication.SharedApplication.Windows[0].SafeAreaInsets;
-
-			initialMargin = element.Margin;
-
-			if (insets.Top <= 0)
```
AndreiMisiukevich, line 29: @sthewissen I am pretty sure this check is not needed (e.g. if some device has only left/right insets or only bottom etc.)

XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs
```diff
@@ -14,34 +14,39 @@ public class SafeAreaEffectRouter : PlatformEffect
...
+			Element.Margin = new Thickness(
+				initialMargin.Left + CalculateInsets(insets.Left, safeArea.Left),
+				initialMargin.Top + CalculateInsets(insets.Top, safeArea.Top),
+				initialMargin.Right + CalculateInsets(insets.Right, safeArea.Right),
+				initialMargin.Bottom + CalculateInsets(insets.Bottom, safeArea.Bottom));
 		}
 
 		protected override void OnDetached()
 		{
-			if (Element is Layout element)
```
jfversluis, line 43: Technically this check has changed now since we are not checking anymore if the Element is a Layout? Do I see that correctly?

jfversluis (approved):
Ran this with old and new version, functionality seems to work the same, just one minor question. If that is just something dumb on my part feel free to merge

XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.ios.cs
AndreiMisiukevich: Yes, previously we checked if Element is a subclass of **View** on attaching and for some reason checked that the Element is a subclass of **Layout** on detaching. I guess it was a mistake

PR #437, opened 2020-10-12
Saving Localization Manager from Linker

pictos:
### Description of Change ###

I changed the Linker behavior in sample projects to the minimal, that way we can see when the linker removes bits.

### Bugs Fixed ###

@AndreiMisiukevich found this issue when he ran the sample project using linker at the minimum config.

### API Changes ###

```csharp
public class TranslateExtension{
   public BindingBase ProvideValue(IServiceProvider serviceProvider)
		{
#if !NETSTANDARD1_0

			if (DateTime.Now.Ticks < 0)
				_ = LocalizationResourceManager.Current[Text];

// rest of the code
}
```

### Behavioral Changes ###

Now the labels in `WelcomePage` appear.

### PR Checklist ###

### DIFF
--- a/XamarinCommunityToolkit/Extensions/TranslateExtension.shared.cs
+++ b/XamarinCommunityToolkit/Extensions/TranslateExtension.shared.cs
@@ -17,14 +17,18 @@ namespace Xamarin.CommunityToolkit.Extensions
 		public BindingBase ProvideValue(IServiceProvider serviceProvider)
 		{
 #if !NETSTANDARD1_0
-            var binding = new Binding
-            {
-                Mode = BindingMode.OneWay,
-                Path = $"[{Text}]",
-                Source = LocalizationResourceManager.Current,
-                StringFormat = StringFormat
-            };
-            return binding;
+
+			if (DateTime.Now.Ticks < 0)
+				_ = LocalizationResourceManager.Current[Text];
+
+			var binding = new Binding
+			{
+				Mode = BindingMode.OneWay,
+				Path = $"[{Text}]",
+				Source = LocalizationResourceManager.Current,
+				StringFormat = StringFormat
+			};
+			return binding;
 #else
 			throw new NotSupportedException("Translate XAML MarkupExtension is not supported on .NET Standard 1.0");
 #endif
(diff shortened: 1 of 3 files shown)


XamarinCommunityToolkit/Extensions/TranslateExtension.shared.cs
```diff
@@ -17,14 +17,18 @@ public class TranslateExtension : IMarkupExtension<BindingBase>
...
-            {
-                Mode = BindingMode.OneWay,
-                Path = $"[{Text}]",
-                Source = LocalizationResourceManager.Current,
-                StringFormat = StringFormat
-            };
-            return binding;
+
+			if (DateTime.Now.Ticks < 0)
+				_ = LocalizationResourceManager.Current[Text];
```
pictos, line 22: We can't use the `Preserve` attribute, because that way the linker never removes the class, and we want to remove it if it's not used. So for this, I called this property `string this [string text]` inside `LocalizationResourceManager` directly and now the linker see that we use it. But this code is never executed and the compiler will perform some improvements in this area.
I learned this from [here](https://github.com/xamarin/Essentials/pull/1209#discussion_r405051562)

XamarinCommunityToolkitSample.iOS/Xamarin.CommunityToolkit.Sample.iOS.csproj
```diff
@@ -25,7 +25,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <MtouchArch>x86_64</MtouchArch>
-    <MtouchLink>None</MtouchLink>
+    <MtouchLink>SdkOnly</MtouchLink>
```
AndreiMisiukevich, line 28: Do we want to set Linker to SdkOnly by default for Debug?

XamarinCommunityToolkitSample.Android/Xamarin.CommunityToolkit.Sample.Android.csproj
```diff
@@ -31,7 +31,7 @@
     <DefineConstants>DEBUG;</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
-    <AndroidLinkMode>None</AndroidLinkMode>
+    <AndroidLinkMode>SdkOnly</AndroidLinkMode>
```
AndreiMisiukevich, line 34: and here

XamarinCommunityToolkitSample.iOS/Xamarin.CommunityToolkit.Sample.iOS.csproj
pictos: I think it's a good idea. That way we can reflect on how the lib will behave in release builds and the person that collaborates with will see any issue sooner.

AndreiMisiukevich (approved):

Issue #442, opened 2020-10-12
[Bug] AvatarView: Label is not centralized 


VladislavAntonyuk:
### Steps to Reproduce

1. Create avatar view with text

### Expected Behavior
Text is vertically centrilized

### Actual Behavior
text is in the top

### Reproduction imagery

![image](https://user-images.githubusercontent.com/33021114/95790599-60264980-0ce8-11eb-8bb8-475d05fe9340.png)

VladislavAntonyuk:
I am going to fix it

VladislavAntonyuk:
expected:
![image](https://user-images.githubusercontent.com/33021114/95790914-f6f30600-0ce8-11eb-9c22-bf6a79a13234.png)

pictos:
Closing this issue since #443 is merged

PR #443, opened 2020-10-12
Fix AvatarView #441, #442

### LINKED ISSUE #441: [Bug] AvatarView: Label is visible if Source is not null
### Steps to Reproduce

1. Create avatar view with transparent image

### Expected Behavior
Label is not visible

### Actual Behavior
Label is visible

### Reproduction imagery
image

VladislavAntonyuk:
### Bugs Fixed ###

- Fixes #441, #442 

### Behavioral Changes ###

Text is centralized and is visible if image source is hidden
image

### PR Checklist ###


XamarinCommunityToolkit/Views/AvatarView.shared.cs
```diff
@@ -191,10 +191,11 @@ void OnValuePropertyChanged()
 			Image.BatchCommit();
 
 			Label.BatchBegin();
+			Label.IsVisible = !Image.IsVisible;
 			var text = Text?.Trim() ?? string.Empty;
 			Label.Text = string.IsNullOrWhiteSpace(text)
 				? emptyText
-				: text?.Trim();
+				: text.Trim();
```
VladislavAntonyuk, line 198: we checked for null on line 196

XamarinCommunityToolkit/Views/AvatarView.shared.cs
```diff
@@ -132,7 +132,7 @@ public FontAttributes FontAttributes
 		protected override void OnControlInitialized(Frame control)
 		{
 			IsClippedToBounds = true;
-			MainLayout.Children.Add(Label, new Rectangle(0, 0, 1, 1), AbsoluteLayoutFlags.All);
+			MainLayout.Children.Add(Label, new Rectangle(0.5, 0.5, -1, -1), AbsoluteLayoutFlags.PositionProportional);
```
VladislavAntonyuk, line 135: #442 Centralize label

XamarinCommunityToolkit/Views/AvatarView.shared.cs
```diff
@@ -191,10 +191,11 @@ void OnValuePropertyChanged()
 			Image.BatchCommit();
 
 			Label.BatchBegin();
+			Label.IsVisible = !Image.IsVisible;
```
VladislavAntonyuk, line 194: #441 Hide label if image is visible

AndreiMisiukevich: I didn't hide the Label when the Image is visible because I wanted to make text visible while the image is loading.
Now if the image has a source, the text will be empty while the source is loading (even if the URI/Path is broken)

XamarinCommunityToolkit/Views/AvatarView.shared.cs
AndreiMisiukevich: Weird

```
HorizontalTextAlignment = TextAlignment.Center,
VerticalTextAlignment = TextAlignment.Center,
```

Seems these guys don't work as expected

XamarinCommunityToolkit/Views/AvatarView.shared.cs
VladislavAntonyuk: we should find a way to subscribe on "OnImageLoaded" and hide label in this case. it doesn't look good now:
image

also adding Aspect property would be nice. what do you think?

AndreiMisiukevich: I totally agree with both points 👍

AndreiMisiukevich: What will happen if we just set the same background color to Image as we set to the entire control?
Maybe it will fix the initial issue?

VladislavAntonyuk: Good idea, I will try it!

AndreiMisiukevich: Ah, no, I think it won't work.. :( we still showing the image above the text.
So, yeah. Need to get the info about successful image load somehow

XamarinCommunityToolkit/Views/AvatarView.shared.cs
VladislavAntonyuk: Why did you choose AbsoluteLayout as the main layout? are there any performance preferences over the grid?
```xml
<Grid>
        <Label HorizontalOptions="CenterAndExpand" VerticalOptions="CenterAndExpand" />
        <Image
            HorizontalOptions="CenterAndExpand"
            VerticalOptions="CenterAndExpand" />
    </Grid>
```

AndreiMisiukevich: Yes, as I know AbsoluteLayot has the best performance in comparison with all other layouts in XF

XamarinCommunityToolkit/Views/AvatarView.shared.cs
AndreiMisiukevich: Here is the custom image source handler

https://github.com/xamarin/XamarinCommunityToolkit/tree/main/XamarinCommunityToolkit/Views/GravatarImageSource

But probably that's too complex.

VladislavAntonyuk: As far as I read Image has property IsLoading which should cover over scenarios

AndreiMisiukevich: Yes, but it didn't provide info about loading status (success/failure)

XamarinCommunityToolkit/Views/AvatarView.shared.cs
```diff
@@ -179,6 +189,13 @@ void OnSizePropertyChanged()
 			BatchCommit();
 		}
 
+		void OnAspectPropertyChanged()
+		{
+			Image.BatchBegin();
```
AndreiMisiukevich, line 194: I think it can be removed since we update a single property :)

XamarinCommunityToolkit/Views/AvatarView.shared.cs
```diff
@@ -31,6 +31,8 @@ public class AvatarView : BaseTemplatedView<Frame>
 			RGB(255, 255, 255)
 		};
 
+		public static readonly BindableProperty AspectProperty = BindableProperty.Create(nameof(Aspect), typeof(Aspect), typeof(AvatarView), Forms.Aspect.AspectFill, propertyChanged: OnAspectPropertyChanged);
```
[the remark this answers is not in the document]
VladislavAntonyuk: if we use OnValuePropertyChanged won't it reset Source each time and tries to download image?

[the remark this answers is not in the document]
AndreiMisiukevich: I think propertyChanged won't be triggered in such cases because the old value and new value are literally the same

https://github.com/xamarin/Xamarin.Forms/blob/5.0.0/Xamarin.Forms.Core/BindableObject.cs#L426

VladislavAntonyuk:
I make it work as expected:
image

VladislavAntonyuk:
image

XamarinCommunityToolkit/Views/AvatarView.shared.cs
```diff
@@ -179,22 +189,29 @@ void OnSizePropertyChanged()
...
 		{
 			if (Control == null)
 				return;
 
 			Image.BatchBegin();
-			var source = Source;
-			Image.IsVisible = source != null;
-			Image.Source = source;
+			Image.IsVisible = false;
+			var imageBytes = await GetImageBytesAsync(Source);
```
AndreiMisiukevich, line 199: Well, 

1) OnValuePropertyChanged is called multiple times, and multiple times we run downloading task
2) We lost data caching

XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="MSBuild.Sdk.Extras/2.0.54">
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;netstandard2.0;netstandard2.1;Xamarin.iOS10;MonoAndroid90;MonoAndroid10.0;Xamarin.TVOS10;Xamarin.WatchOS10;Xamarin.Mac20;tizen40</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;Xamarin.iOS10;MonoAndroid90;MonoAndroid10.0;Xamarin.TVOS10;Xamarin.WatchOS10;Xamarin.Mac20;tizen40</TargetFrameworks>
```
AndreiMisiukevich, line 4: we don't want to drop netstandard1.0 support

XamarinCommunityToolkit/Views/AvatarView.shared.cs
VladislavAntonyuk: As for the first item it is easy fix.
How to solve the issue with caching? I found that FFImageLoading has all the features we need, but it is very complex and I don’t want to add a new dependency

XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj
VladislavAntonyuk: I added the code to check if image source is valid, but it has APIs which were added in net standard 2.0. I can revert this changes but the issue with Label visibility will remain for net standard 1.0

VladislavAntonyuk: Also why do we use net standard 1.0? In the compatibility table I see that all minimum required OS (android, iOS, Windows) work with net standard 2.0

AndreiMisiukevich: I guess because some of the users can still use PCL

XamarinCommunityToolkit/Views/AvatarView.shared.cs
VladislavAntonyuk: Probably I have found a solution which should satisfy our requirements. And it should work with net standard 1.0

VladislavAntonyuk:
@AndreiMisiukevich I implemented it for WPF. Please check if this solution is applicable, so I continue to implement it for other platforms.

XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj
```diff
@@ -151,6 +151,7 @@
     <Compile Include="**\*.wpf.cs" />
     <Compile Include="**\*.wpf.*.cs" />
     <Compile Remove="**\Snackbar.shared.cs" />
+    <Compile Remove="**\AvatarView\AvatarView.shared.cs" />
```
AndreiMisiukevich, line 154: i think this line and a line above should not be there, right?

AndreiMisiukevich:
> @AndreiMisiukevich I implemented it for WPF. Please check if this solution is applicable, so I continue to implement it for other platforms.

Hm, interesting case. I think it can work, but just a few comments

1) Do we need to create own handler for every image? Or we can resolve singleton from anywhere?
2) ?

```csharp
async Task<bool> IsImageSourceValid(ImageSource source)
{
	var handler = GetHandler(source);
	if (handler == null)
		return false;
		
	var imageSource = await handler.LoadImageAsync(source);
	return imageSource != null;
}

IImageSourceHandler GetHandler(ImageSource source)
{
	if (source is UriImageSource)
		return new UriImageSourceHandler();

	if (source is StreamImageSource)
		return new StreamImageSourceHandler();

	if (source is FileImageSource fileSource && File.Exists(f.File))
		return new FileImageSourceHandler()

	return null;
}
```

VladislavAntonyuk:
> > @AndreiMisiukevich I implemented it for WPF. Please check if this solution is applicable, so I continue to implement it for other platforms.
> 
> Hm, interesting case. I think it can work, but just a few comments
> 
> 1. Do we need to create own handler for every image? Or we can resolve singleton from anywhere?
> 2. ?
> 
> ```cs
> async Task<bool> IsImageSourceValid(ImageSource source)
> {
> 	var handler = GetHandler(source);
> 	var imageSource = handler != null
> 		? await handler.LoadImageAsync(source)
> 		: null;
> 	return imageSource != null;
> }
> 
> IImageSourceHandler GetHandler(ImageSource source)
> {
> 	if (source is UriImageSource)
> 		return new UriImageSourceHandler();
> 
> 	if (source is StreamImageSource)
> 		return new StreamImageSourceHandler();
> 
> 	if (source is FileImageSource fileSource && File.Exists(f.File))
> 		return new FileImageSourceHandler()
> 
> 	return null;
> }
> ```

The issue is that net standard doesn't have IImageSourceHandler, so we can not use it in shared.

AndreiMisiukevich:
> > > @AndreiMisiukevich I implemented it for WPF. Please check if this solution is applicable, so I continue to implement it for other platforms.
> > 
> > 
> > Hm, interesting case. I think it can work, but just a few comments
> > 
> > 1. Do we need to create own handler for every image? Or we can resolve singleton from anywhere?
> > 2. ?
> > 
> > ```cs
> > async Task<bool> IsImageSourceValid(ImageSource source)
> > {
> > 	var handler = GetHandler(source);
> > 	var imageSource = handler != null
> > 		? await handler.LoadImageAsync(source)
> > 		: null;
> > 	return imageSource != null;
> > }
> > 
> > IImageSourceHandler GetHandler(ImageSource source)
> > {
> > 	if (source is UriImageSource)
> > 		return new UriImageSourceHandler();
> > 
> > 	if (source is StreamImageSource)
> > 		return new StreamImageSourceHandler();
> > 
> > 	if (source is FileImageSource fileSource && File.Exists(f.File))
> > 		return new FileImageSourceHandler()
> > 
> > 	return null;
> > }
> > ```
> 
> The issue is that net standard doesn't have IImageSourceHandler, so we can not use it in shared.

```#if !NETSTANDARD10``` ? And make something else for netstandard?

**Comment by VladislavAntonyuk** [comment]
> > > > @AndreiMisiukevich I implemented it for WPF. Please check if this solution is applicable, so I continue to implement it for other platforms.
> > > 
> > > 
> > > Hm, interesting case. I think it can work, but just a few comments
> > > 
> > > 1. Do we need to create own handler for every image? Or we can resolve singleton from anywhere?
> > > 2. ?
> > > 
> > > ```cs
> > > async Task<bool> IsImageSourceValid(ImageSource source)
> > > {
> > > 	var handler = GetHandler(source);
> > > 	var imageSource = handler != null
> > > 		? await handler.LoadImageAsync(source)
> > > 		: null;
> > > 	return imageSource != null;
> > > }
> > > 
> > > IImageSourceHandler GetHandler(ImageSource source)
> > > {
> > > 	if (source is UriImageSource)
> > > 		return new UriImageSourceHandler();
> > > 
> > > 	if (source is StreamImageSource)
> > > 		return new StreamImageSourceHandler();
> > > 
> > > 	if (source is FileImageSource fileSource && File.Exists(f.File))
> > > 		return new FileImageSourceHandler()
> > > 
> > > 	return null;
> > > }
> > > ```
> > 
> > 
> > The issue is that net standard doesn't have IImageSourceHandler, so we can not use it in shared.
> 
> `#if !NETSTANDARD10` ? And make something else for netstandard?

different platforms have different LoadImageAsync range of parameters

File: XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj (hunk 8 above)
Comment (VladislavAntonyuk) [reply XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj]: fixed in SnackBarOptions PR

File: XamarinCommunityToolkit/Views/AvatarView/AvatarView.shared.cs [hunk 9]
```diff
@@ -1,11 +1,19 @@
...
 using static System.Math;
 
 namespace Xamarin.CommunityToolkit.UI.Views
 {
 	public class AvatarView : BaseTemplatedView<Frame>
 	{
 		const string emptyText = "X";
+		static readonly IImageSourceValidator imageSourceValidator;
+
+		static AvatarView() => imageSourceValidator = new ImageSourceValidator();
```
VladislavAntonyuk, line 14: should we allow users to implement there own IImageSourceValidator like we do for IColorScheme?

XamarinCommunityToolkit/Views/AvatarView/AvatarView.shared.cs
```diff
@@ -160,22 +176,32 @@ void OnSizePropertyChanged()
 			BatchCommit();
 		}
 
-		void OnValuePropertyChanged()
+		async void OnValuePropertyChanged(bool force)
```
AndreiMisiukevich, line 179: can we please choose a more clear name?

Maybe "shouldUpdateSource" ?

XamarinCommunityToolkit/Views/AvatarView/ImageSourceValidator.android.ios.macos.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Xamarin.CommunityToolkit.UI.Views
+{
+	class ImageSourceValidator : IImageSourceValidator
+	{
+		public async Task<bool> IsImageSourceValid(ImageSource source)
+		{
+			IImageSourceHandler handler;
+
+			switch (source)
```
AndreiMisiukevich, line 21: Don't you like the idea of extracting the logic of getting an image handle into a separate method?)

```csharp
IImageSourceHandler GetHandler(ImageSource source)
{
	if (source is UriImageSource)
		return new UriImageSourceHandler();

	if (source is StreamImageSource)
		return new StreamImageSourceHandler();

	if (source is FileImageSource fileSource && File.Exists(f.File))
		return new FileImageSourceHandler()

	return null;
}
```

XamarinCommunityToolkit/Views/AvatarView/ImageSourceValidator.uwp.wpf.gtk.tizen.cs
```diff
@@ -0,0 +1,54 @@
...
+using Xamarin.Forms.Platform.UWP;
+#elif NET471
+using Xamarin.Forms.Platform.GTK.Renderers;
+#elif TIZEN
+using Xamarin.Forms.Platform.Tizen;
+#else
+using Xamarin.Forms.Platform.WPF;
+#endif
+
+namespace Xamarin.CommunityToolkit.UI.Views
```
AndreiMisiukevich, line 14: We probably can try to put all platforms inside the single file

```csharp
namespace Xamarin.CommunityToolkit.UI.Views
{
    #if UWP || NET471 || TIZEN || WPF
    using ImageLoaderSourceHandler = UriImageSourceHandler;
    #endif 

    #if NET471 || MONOANDROID || __IOS__ || __MAC__
    using StreamImageSourceHandler = StreamImagesourceHandler;
    #endif
```

XamarinCommunityToolkit/Views/AvatarView/ImageSourceValidator.uwp.wpf.gtk.tizen.cs
```diff
@@ -0,0 +1,54 @@
...
+				default:
+					return false;
+			}
+
+#if TIZEN
+			var imageSource = await handler.LoadImageAsync(null, source) ? true : (bool?)null;
+#else
+			var imageSource = await handler.LoadImageAsync(source);
+#endif
+			return imageSource != null;
```
AndreiMisiukevich, line 51:
```suggestion
			return await handler.LoadImageAsync(null, source);
#else
			var imageSource = await handler.LoadImageAsync(source);
			return imageSource != null;
#endif
```

XamarinCommunityToolkit/Views/AvatarView/IImageSourceValidator.shared.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Threading.Tasks;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.UI.Views
+{
+	public interface IImageSourceValidator
+	{
+		Task<bool> IsImageSourceValid(ImageSource source);
```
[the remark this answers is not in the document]
VladislavAntonyuk: I agree with Async, but from the name "IsImageSourceValidAsync" it is clear that the result is boolean

XamarinCommunityToolkit/Views/AvatarView/ImageSourceValidator.android.ios.macos.cs
VladislavAntonyuk: I like the idea putting it into the single platform file. probably I misunderstood you. I thought you suggested to return IImageSourceHandler as a result of IsImageSourceValidAsync(...)

XamarinCommunityToolkit/Views/AvatarView/IImageSourceValidator.shared.cs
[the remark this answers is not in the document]
AndreiMisiukevich: Usually "is" is used as a suffix for properties and fields. But maybe I am wrong :)

XamarinCommunityToolkit/Views/AvatarView/ImageSourceValidator.android.ios.macos.cs
AndreiMisiukevich: Oh, this suggestion is only about code style :)

Replace 

```
IImageSourceHandler handler;
switch (source) 
{
....
}
```

by

```
var handler = GetHandler(source);

IImageSourceHandler GetHandler(ImageSource source)
{
	if (source is UriImageSource)
		return new UriImageSourceHandler();

	if (source is StreamImageSource)
		return new StreamImageSourceHandler();

	if (source is FileImageSource fileSource && File.Exists(f.File))
		return new FileImageSourceHandler()

	return null;
}
```

XamarinCommunityToolkit/Views/AvatarView/AvatarView.shared.cs
AndreiMisiukevich: I think no, at least for now

XamarinCommunityToolkit/Views/AvatarView/AvatarView.shared.cs
AndreiMisiukevich: For parameter "force", I mean

XamarinCommunityToolkit/Views/AvatarView/IImageSourceValidator.shared.cs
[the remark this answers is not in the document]
VladislavAntonyuk: I follow the example like this:
```
public static bool IsNullOrEmpty(string value) {
    return (value == null || value.Length == 0);
}
```

XamarinCommunityToolkit/Views/AvatarView/AvatarView.shared.cs
```diff
@@ -160,22 +176,32 @@ void OnSizePropertyChanged()
...
-			Image.Source = source;
+			if (force)
+			{
+				Image.IsVisible = false;
+				var isValid = await imageSourceValidator.IsImageSourceValid(Source);
+				if (isValid)
+				{
+					Image.IsVisible = true;
+					Image.Source = Source;
+				}
```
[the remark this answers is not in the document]
VladislavAntonyuk: we shouldn't load image if image source is invalid

XamarinCommunityToolkit/Views/AvatarView/AvatarView.shared.cs
AndreiMisiukevich, line 14:
```suggestion
		const string emptyText = "X";
		
		static readonly IImageSourceValidator imageSourceValidator = new ImageSourceValidator();
```

XamarinCommunityToolkit/Views/AvatarView/AvatarView.shared.cs
```diff
@@ -160,22 +175,31 @@ void OnSizePropertyChanged()
...
-			Image.IsVisible = source != null;
-			Image.Source = source;
+			if (shouldUpdateSource)
+			{
+				var isImageSourceValid = await imageSourceValidator.IsImageSourceValidAsync(Source);
+				Image.IsVisible = isImageSourceValid;
+				if (isImageSourceValid)
+				{
+					Image.Source = Source;
+				}
```
AndreiMisiukevich, line 191:
```suggestion
				Image.IsVisible = await imageSourceValidator.IsImageSourceValidAsync(Source);
				Image.Source = Source;
```

XamarinCommunityToolkit/Views/AvatarView/ImageSourceValidator.shared.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Threading.Tasks;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.UI.Views
+{
+#if NETSTANDARD || __TVOS__ || __WATCHOS__
+	class ImageSourceValidator : IImageSourceValidator
+	{
+		// we should still try to load image
```
AndreiMisiukevich, line 9:
```suggestion
		// We should still try to load an image
```

AndreiMisiukevich (approved):

AndreiMisiukevich:
Perfect work!
I believe that's fine that we try to "pre-load" an image for checking if it's a valid image or not (Caching will save us from loading the same image several times)

@jfversluis @pictos please take a look at this implementation as well

XamarinCommunityToolkit/Views/AvatarView/ImageSourceValidator.android.ios.macos.uwp.wpf.gtk.tizen.cs
```diff
@@ -0,0 +1,62 @@
...
+#endif
+
+	class ImageSourceValidator : IImageSourceValidator
+	{
+		public async Task<bool> IsImageSourceValidAsync(ImageSource source)
+		{
+			var handler = GetHandler(source);
+
+#if TIZEN
+			return await handler.LoadImageAsync(null, source);
```
pictos, line 38:
```suggestion
			return await handler.LoadImageAsync(null, source).ConfiguraAwait(false);
```

XamarinCommunityToolkit/Views/AvatarView/ImageSourceValidator.android.ios.macos.uwp.wpf.gtk.tizen.cs
```diff
@@ -0,0 +1,62 @@
...
+using Xamarin.Forms.Platform.GTK.Renderers;
+#elif TIZEN
+using Xamarin.Forms.Platform.Tizen;
+#else
+using Xamarin.Forms.Platform.WPF;
+#endif
+
+namespace Xamarin.CommunityToolkit.UI.Views
+{
+#if MONOANDROID || __IOS__ || __MACOS__
```
pictos, line 23: any reason for these `using`s be here (inside the namespace)?

pictos (changes requested):
This fix is very welcome! Tks to implement on all platforms!

I'm facing a `NullReferenceException` when I try to run on Android.
image

I can't test on iOS right now, but when I do I let you know my thoughts (:

XamarinCommunityToolkit/Views/AvatarView/ImageSourceValidator.android.ios.macos.uwp.wpf.gtk.tizen.cs
AndreiMisiukevich: Otherwise, we will have to specify the full namespace path

XamarinCommunityToolkit/Views/AvatarView/ImageSourceValidator.android.ios.macos.uwp.wpf.gtk.tizen.cs
```diff
@@ -0,0 +1,62 @@
...
+#if MONOANDROID || __IOS__ || __MACOS__ || NET471
+	using StreamImageSourceHandler = StreamImagesourceHandler;
+#endif
+
+	class ImageSourceValidator : IImageSourceValidator
+	{
+		public async Task<bool> IsImageSourceValidAsync(ImageSource source)
+		{
+			var handler = GetHandler(source);
+
```
AndreiMisiukevich, line 36: need to insert a check

```csharp
if (handler == null)
    return false;
```

pictos: That solves for me

XamarinCommunityToolkit/Views/AvatarView/AvatarView.shared.cs
```diff
@@ -1,11 +1,18 @@
...
+using System.Threading.Tasks;
+using Xamarin.Forms;
 using static System.Math;
 
 namespace Xamarin.CommunityToolkit.UI.Views
 {
 	public class AvatarView : BaseTemplatedView<Frame>
 	{
 		const string emptyText = "X";
+
```
pictos, line 12: We have two tabs here. Can you please remove it
```suggestion

```

XamarinCommunityToolkit/Views/AvatarView/ImageSourceValidator.android.ios.macos.uwp.wpf.gtk.tizen.cs
VladislavAntonyuk: if we return False the image won't displayed if ImageSource is GravatarImageSource. Is it expected?

pictos: @VladislavAntonyuk I tried here, and the image displayed after loading. So maybe it's null when we load the page in first time. At least for android

AndreiMisiukevich: Hm, you are right. Then let's return True

VladislavAntonyuk: @pictos AvatarView line 185. if we return false the image IsVisible should be false.

pictos: so... Why return false works?

XamarinCommunityToolkit/Views/AvatarView/ImageSourceValidator.android.ios.macos.uwp.wpf.gtk.tizen.cs
```diff
@@ -0,0 +1,64 @@
...
+		public async Task<bool> IsImageSourceValidAsync(ImageSource source)
+		{
+			var handler = GetHandler(source);
+			if (handler == null)
+    			return true;
+
+#if TIZEN
+			return await handler.LoadImageAsync(null, source).ConfigureAwait(false);
+#elif MONOANDROID
+			var imageSource = await handler.LoadImageAsync(source, null);
```
pictos, line 42: Tested here and works fine.
```suggestion
			var imageSource = await handler.LoadImageAsync(source, null).ConfigureAwait(false);
```

XamarinCommunityToolkit/Views/AvatarView/ImageSourceValidator.android.ios.macos.uwp.wpf.gtk.tizen.cs
```diff
@@ -0,0 +1,64 @@
...
+			return imageSource != null;
+#else
+			var imageSource = await handler.LoadImageAsync(source);
```
pictos, line 45:
```suggestion
			var imageSource = await handler.LoadImageAsync(source).ConfigureAwait(false);
```

XamarinCommunityToolkit/Views/AvatarView/ImageSourceValidator.shared.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Threading.Tasks;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.UI.Views
+{
+#if NETSTANDARD || __TVOS__ || __WATCHOS__
+	class ImageSourceValidator : IImageSourceValidator
+	{
+		// We should still try to load an image
```
pictos, line 9: Do we need this comment here?

pictos (approved):
Everything works well here, I added a couple of suggestions, but that doesn't invalidate your amazing work!

Thanks!

XamarinCommunityToolkit/Views/AvatarView/ImageSourceValidator.android.ios.macos.uwp.wpf.gtk.tizen.cs
VladislavAntonyuk: @AndreiMisiukevich we should return false. Otherwise Label is not visible, because Label.IsVisible=!Image.IsVisible

VladislavAntonyuk: I added more ImageSources (GravatarImageSource, FontImageSource)

AndreiMisiukevich (approved):

pictos (approved):

AndreiMisiukevich:
XD

PR #444, opened 2020-10-13
Fix unit tests

jfversluis:
### Description of Change ###

Removes the net461 target and an API in the mock that is not available in the installed Forms version in order to restore the unit test project to build and run successfully

### DIFF
--- a/XamarinCommunityToolkit.UnitTests/Xamarin.CommunityToolkit.UnitTests.csproj
+++ b/XamarinCommunityToolkit.UnitTests/Xamarin.CommunityToolkit.UnitTests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;net461</TargetFrameworks>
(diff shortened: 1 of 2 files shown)


XamarinCommunityToolkit.UnitTests/Xamarin.CommunityToolkit.UnitTests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;net461</TargetFrameworks>
```
jfversluis, line 4: @VladislavAntonyuk is there any reason you added this as part of your WPF and GTK PR?

jfversluis:
Maybe this was just something that is not working locally. It seems to build on the build server... 😕

XamarinCommunityToolkit.UnitTests/Xamarin.CommunityToolkit.UnitTests.csproj
VladislavAntonyuk: Snackbar unit tests are failing with netcoreapp3.1 because it tries to run Snackbar for WPF

PR #449, opened 2020-10-14
Fix #286 Prevent extra iOS camera inits

### LINKED ISSUE #286: [Bug] [CameraView] [iOS] When binding CameraOptions the Camera gets initialized twice
### Description
If you bind the Property CameraOptions, the CameraViewRenderer initializes the Camera twice.
If the initial value is CameraOptions.Front it will first initialize the back camera (because of CameraOptions.Default) and then the front camera.
If the initial value is CameraOptions.Back it will first initialize the back camera (because of CameraOptions.Default) and then again the back camera.

### Steps to Reproduce
1. Create CameraView
2. Bind property CameraOptions to a view model

### Expected Behavior
Only the camera that is bound with CameraOptions will be initialized.

### Actual Behavior
The back camera will always be initialized because of CameraOptions.Default, following by the camera referenced by CameraOptions.

### Basic Information

- Platform Target Frameworks: 
  - iOS:  13.6.1

bmacombe:
### Description of Change ###

Caches the last camera position initialed and only reinitialize if the position changes.
Removes un-need init from FormsCamera ctor.
If camera permission must be asked for, after a successful granting of permission it retries camera init.

I tested in the sample app and now I only get 1 camera init to the default (back) using the sample.  If in the camera sample page xaml I set the camera option to "Front", I only get 1 camera init to the front camera.

As long as the camera option is set before the renderer is created, it should only result in 1 init.

Testing Procedure
* Set a breakpoint in FormsCameraView.RetrieveCameraDevice 
* Make sure it is only called once at startup. (previous behavior was called twice)
* Delete sample app from a device, redeploy and then make sure the camera preview shows up after asking for camera permissions. (previous behavior was a blank screen).

### Bugs Fixed ###

- Fixes #286 

### API Changes ###
None

### Behavioral Changes ###
None to user

### PR Checklist ###

### DIFF
--- a/XamarinCommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs
+++ b/XamarinCommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs
@@ -52,7 +52,7 @@ namespace Xamarin.CommunityToolkit.UI.Views
 			};
 
 			mainView.Layer.AddSublayer(previewLayer);
-			RetrieveCameraDevice(CameraOptions.Default);
+			//RetrieveCameraDevice(CameraOptions.Default);
 
 			Add(mainView);
 
--- a/XamarinCommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs
+++ b/XamarinCommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs
@@ -21,6 +21,7 @@ namespace Xamarin.CommunityToolkit.UI.Views
 		AVCaptureMovieFileOutput videoOutput;
 		AVCaptureConnection captureConnection;
 		AVCaptureDevice device;
+		AVCaptureDevicePosition? lastPosition;
 		bool isBusy;
 		bool isAvailable;
 		CameraFlashMode flashMode;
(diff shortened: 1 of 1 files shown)


XamarinCommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs
```diff
@@ -51,7 +52,7 @@ public FormsCameraView()
 			};
 
 			mainView.Layer.AddSublayer(previewLayer);
-			RetrieveCameraDevice(CameraOptions.Default);
+			//RetrieveCameraDevice(CameraOptions.Default);
```
jfversluis, line 55: Should we just remove this so we won't wonder why it was there in the first place? :D

bmacombe: yes, I just noticed that!

bmacombe:
Just noticed an issue when doing a fresh start of the app and the camera view not initializing, moved back to draft until I sort that out.

bmacombe:
Might need to add a check to init the camera after permission to use the camera is asked of the user.

bmacombe:
Added some retry logic to init the camera after the user has given permission for the app to access the camera.  Previously it would not correctly retry after permission was given.  So the camera control wasn't initialized until the CameraOption was changed.

bmacombe:
@jfversluis @pictos This PR should be ready for final review.

Thanks!

jfversluis (approved):
Seems to work fine. There is however something funny about the sample app. It will never show the button text to stop recording when making a video. Functionally it does work. Not sure it that has to do with this change though

Issue #275, opened 2020-09-08
[Spec] StateLayout


sthewissen:
# StateLayout

Displaying items when your app is in a specific state is a common pattern throughout any mobile app. For instance, creating loading views to overlay on an entire screen or maybe a subsection of your screen needs an individual loader. When there's no data to display we create empty state views or when something goes wrong we need to build an error state view. By implementing the `StateLayout`'s attached properties you can turn any layout element like a `Grid` or `StackLayout` into an individual state-aware element.

This is pretty much me wanting to incorporate StateSquid 🦑  in here :)

# API

## StateView
```
public class StateView : ContentView
{
    public State StateKey { get; set; }
    public string CustomStateKey { get; set; }
    public int RepeatCount  { get; set; }
    public StateView RepeatTemplate { get; set; }

    ...
}
```

## State (Enum)
```
public enum State
{
    None,
    Loading,
    Saving,
    Success,
    Error,
    Empty,
    Custom
}
```
## StateLayout
```
public static class StateLayout
{
    public IList<StateView> StateViews { get; set; }
    public State CurrentState { get; set; }
    public string CurrentCustomStateKey { get; set; }
    public bool AnimateStateChanges { get; set; }
}
```

## StateLayoutController
```
public class StateLayoutController
{
    void SwitchToContent(bool animate) { }
    void SwitchToTemplate(string customState, bool animate)  { }
    void SwitchToTemplate(State state, string customState, bool animate) { }
    bool HasTemplateForState(State state, string customState) { }

    ...
}
```

# Scenarios

## Basic scenario

Each layout that you make state-aware using the `StateLayout` attached properties contains a collection of `StateView` objects. These can be used as templates for the different states supported by `StateLayout`. Whenever the `CurrentState` is set to a value that matches the `State` property of one of the `StateViews` its contents will be displayed instead of the main content.

```
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms" 
   xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml" 
   xmlns:state="clr-namespace:Xamarin.Forms.StateSquid;assembly=Xamarin.Forms.StateSquid" 
   x:Class="SampleApp.MainPage">

   <Grid state:StateLayout.CurrentState="{Binding CurrentState}">
      <state:StateLayout.StateViews>
         <state:StateView StateKey="Loading">
            <Grid BackgroundColor="White">
[...]

PR #450, opened 2020-10-16
Added the StateLayout

sthewissen:
### Description of Change ###
This PR adds `StateLayout` and all of its helper classes. This is currently up as a separate NuGet called "StateSquid" here:
https://www.nuget.org/packages/Xamarin.Forms.StateSquid 🦑

### Bugs Fixed ###
- Implements #275

### API Changes ###
Added all the classes as already described in #275. The only thing I haven't put in here is the `SkeletonView` that was mentioned there. I don't want to take that in to be honest, as I think it should be up to the user to create their own visualization. I'm sure we could come up with a better skeleton loading solution anyway.

### PR Checklist ###


sthewissen:
I hope to be able to add some documentation for this and can port some of the existing samples from the repo. Not sure if I'll come around to writing some adequate tests due to time constraints, so if anyone is up for that feel free to add them.

jfversluis:
Builds 🎉

XamarinCommunityToolkit/Converters/StateToBooleanConverter.shared.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Globalization;
+using Xamarin.Forms;
+using Xamarin.CommunityToolkit.UI.Views;
+
+namespace Xamarin.CommunityToolkit.Converters
+{
+	public class StateToBooleanConverter : IValueConverter
```
AndreiMisiukevich, line 8: Should it be inherited from ValueConverterExtension as well as other converters?
+ Add a StateToCompare property

XamarinCommunityToolkit/Converters/StateToBooleanConverter.shared.cs
```diff
@@ -0,0 +1,16 @@
...
+	{
+		public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+			=> value is State state && parameter is State stateToCompare && state == stateToCompare;
```
AndreiMisiukevich, line 11: 1) if value is not State, then raise exception
2) try to get State from parameter (if no luck then go ahead with StateToCompare property)

I believe we implemented a similar approach :0

XamarinCommunityToolkit/Converters/StateToBooleanConverter.shared.cs
```diff
@@ -0,0 +1,16 @@
...
+
+		public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+			=> State.None;
```
AndreiMisiukevich, line 14: Perhaps we should raise an exception here (?)

XamarinCommunityToolkit/Views/StateLayout/State.shared.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+	public enum State
+	{
+		None,
+		Loading,
+		Saving,
+		Success,
+		Error,
+		Empty,
+		Custom
```
AndreiMisiukevich, line 11: I wonder, is there any sense to keep enum if we cannot describe all possible cases?

Maybe we can refactor it to static class (similar to how XF Device class operates with RuntimePlatform)
```csharp
public static class LayoutState {
    public const string None = nameof(None);
    public const string Loading = nameof(Loading);
    /// etc.
}
```

XamarinCommunityToolkit/Views/StateLayout/State.shared.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Xamarin.CommunityToolkit.UI.Views
+{
+	public enum State
```
AndreiMisiukevich, line 3: Hmm, LayoutState?

State is too general, I believe

XamarinCommunityToolkit/Views/StateLayout/StateLayout.shared.cs
```diff
@@ -0,0 +1,88 @@
+﻿using System;
+using System.Collections.Generic;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.UI.Views
+{
+	public static class StateLayout
+	{
+		internal static readonly BindablePropertyKey StateViewsPropertyKey
+			= BindableProperty.CreateReadOnly("StateViews", typeof(IList<StateView>), typeof(Layout<View>), default(IList<StateView>),
```
AndreiMisiukevich, line 10: IList<StateView> -> IEnumerable<StateView>

(?) Does it make sense?

XamarinCommunityToolkit/Views/StateLayout/StateLayout.shared.cs
```diff
@@ -0,0 +1,88 @@
...
+				defaultValueCreator: bindable =>
```
AndreiMisiukevich, line 11: Is there any sense to use defaultValueCreator here?
I guess we can just put ```new List<StateView>()``` as a default value
Or ```Enumerable.Empty<StateView>()``` in case we switch to IEnumerable interface

XamarinCommunityToolkit/Views/StateLayout/StateLayout.shared.cs
```diff
@@ -0,0 +1,88 @@
...
+		{
+			// Swap out the current children for the Loading Template.
+			if (oldValue != newValue && newValue != State.None && newValue != State.Custom)
+			{
+				GetLayoutController(bindable).SwitchToTemplate(newValue, null, GetAnimateStateChanges(bindable));
+			}
+			else if (oldValue != newValue && newValue == State.None)
+			{
+				GetLayoutController(bindable).SwitchToContent(GetAnimateStateChanges(bindable));
+			}
```
AndreiMisiukevich, line 70:
```csharp
			if (oldValue == newValue)
				return;
				
			// Swap out the current children for the Loading Template.
			switch (newValue) {
				case State.Custom:
					break;
				case State.None:
					GetLayoutController(bindable).SwitchToContent(GetAnimateStateChanges(bindable));
					break;
				default:
					GetLayoutController(bindable).SwitchToTemplate(newValue, null, GetAnimateStateChanges(bindable));
					break;
			}
```

XamarinCommunityToolkit/Views/StateLayout/StateLayout.shared.cs
```diff
@@ -0,0 +1,88 @@
...
+		}
+
+		static void OnCurrentCustomStateKeyChanged(BindableObject bindable, string oldValue, string newValue)
+		{
+			var state = GetCurrentState(bindable);
+
+			// Swap out the current children for the Loading Template.
+			if (oldValue != newValue && state == State.Custom)
```
AndreiMisiukevich, line 78: Same here.

```if (oldValue == newValue) return;```

File: XamarinCommunityToolkit/Views/StateLayout/StateLayoutController.shared.cs [hunk 10]
```diff
@@ -0,0 +1,177 @@
...
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.UI.Views
+{
+	public class StateLayoutController
+	{
+		readonly WeakReference<Layout<View>> layoutWeakReference;
+		bool layoutIsGrid = false;
+		IList<View> originalContent;
+		State previousState = State.None;
```
AndreiMisiukevich, line 14:
```suggestion
		bool layoutIsGrid;
		State previousState;
		IList<View> originalContent;
```

XamarinCommunityToolkit/Views/StateLayout/StateLayoutController.shared.cs
```diff
@@ -0,0 +1,177 @@
...
+
+		async Task ChildrenFadeTo(Layout<View> layout, bool animate, bool isHide)
+		{
+			if (animate && layout?.Children?.Count > 0)
+			{
+				var tasks = new List<Task<bool>>();
+				foreach (var a in layout.Children)
+					tasks.Add(a.FadeTo(isHide ? 0 : 1, isHide ? 100u : 500u));
+
+				await Task.WhenAll(tasks);
```
AndreiMisiukevich, line 173:
```csharp
var opacity = 1;
var time = 500u;
if (isHide)
{
	opacity = 0;
	time = 	100u;
}
await Task.WhenAll(layout.Children.Select(a => a.FadeTo(opacity, time)));
```

XamarinCommunityToolkit/Views/StateLayout/StateView.shared.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Xamarin.CommunityToolkit.UI.Views
+{
+	public class StateView : ContentView
+	{
+		public static readonly BindableProperty StateKeyProperty
+			= BindableProperty.Create(nameof(StateKey), typeof(State), typeof(StateView), default(State));
+
+		public static readonly BindableProperty CustomStateKeyProperty
+			= BindableProperty.Create(nameof(CustomStateKey), typeof(string), typeof(StateView), default(string));
```
AndreiMisiukevich, line 12: default(string) can be omitted

XamarinCommunityToolkit/Views/StateLayout/StateView.shared.cs
```diff
@@ -0,0 +1,44 @@
...
+
+		public static readonly BindableProperty RepeatCountProperty
+			= BindableProperty.Create(nameof(RepeatCount), typeof(int), typeof(StateView), 1);
+
+		public static readonly BindableProperty RepeatTemplateProperty
+			= BindableProperty.Create(nameof(RepeatTemplate), typeof(DataTemplate), typeof(StateView), null);
```
AndreiMisiukevich, line 18: null can be omitted

XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj
```diff
@@ -23,8 +23,12 @@
     <RepositoryUrl>https://github.com/xamarin/XamarinCommunityToolkit</RepositoryUrl>
     <PackageReleaseNotes>
       
+
```
AndreiMisiukevich, line 26: Maybe let's remove this section at all? it always tries to insert new empty lines

XamarinCommunityToolkitSample.Android/Xamarin.CommunityToolkit.Sample.Android.csproj
```diff
@@ -31,7 +31,6 @@
     <DefineConstants>DEBUG;</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
-    <AndroidLinkMode>SdkOnly</AndroidLinkMode>
```
AndreiMisiukevich, line 34: Probably we don't want to drop this line

XamarinCommunityToolkitSample/Resx/AppResources.resx
```diff
@@ -491,43 +491,49 @@ If the user manually assigns a row or column on a view, it will be honored.</val
...
+    <data name="ToggleSafeAreaEffect" xml:space="preserve">
     <value>Toggle SafeAreaEffect</value>
   </data>
+    <data name="StateLayoutTitle" xml:space="preserve">
+    <value>StateLayout</value>
+  </data>
+    <data name="StateLayoutDescription" xml:space="preserve">
+    <value>A collection of attached properties that let you specify one or more state views for any of your existing layouts.</value>
+  </data>
 </root>
```
AndreiMisiukevich, line 539: Hmm, this file updated indentation...

XamarinCommunityToolkit/Views/StateLayout/State.shared.cs
sthewissen: Thing is, that would mean we have `StateLayout` and `LayoutState`. That confuses me even more 😅

AndreiMisiukevich: Yeah, that's fair. But if we will add several new controls which have "State" property, it will confuse not less :)

XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj
sthewissen: We can do that, definitely. Once we actually have release notes on a docs portal, we can add them back in.

XamarinCommunityToolkit/Views/StateLayout/State.shared.cs
sthewissen: Agree. Perhaps I shall think a bit longer on an alternative :)

XamarinCommunityToolkitSample/Resx/AppResources.resx
sthewissen: Yeah, seems my VSMac save messes it up. Tried to fix.

XamarinCommunityToolkit/Views/StateLayout/State.shared.cs
jfversluis: Are we still going to change this name?

sthewissen: Open to suggestions?

AndreiMisiukevich:
Hmmm, something went wrong with rebase, I guess

sthewissen:
> Hmmm, something went wrong with rebase, I guess

Working on it :)

XamarinCommunityToolkit/Views/StateLayout/State.shared.cs
sthewissen: Been trying this but so far it's been a bunch of misery :P The custom state is explicitly added for scenarios not covered by the default states. Also making it a string introduces some strange scenarios where the user **has to** explicitly set a value for it to work. We can fix that in the component obviously, but it feels a bit strange.

sthewissen:
I rebased once more. Unless we can come up with a better name I'd say leave `State` as is for now. `ViewState` is a possible alternative, but that can also be a bit confusing. Regarding making the `State` constant strings, I'm personally not a fan as it would require quite rewriting and logic adjustments. I personally don't have the time for that and would make migration for existing users more cumbersome (I think).

AndreiMisiukevich:
> I rebased once more. Unless we can come up with a better name I'd say leave `State` as is for now. `ViewState` is a possible alternative, but that can also be a bit confusing. Regarding making the `State` constant strings, I'm personally not a fan as it would require quite rewriting and logic adjustments. I personally don't have the time for that and would make migration for existing users more cumbersome (I think). @jfversluis @AndreiMisiukevich

That's up to you.
But I do like "LayoutState" name. Similar to "ExpandState" in Expander. Or StateLayoutState.

sthewissen:
@AndreiMisiukevich I renamed to `LayoutState`, let's see how it goes :)

AndreiMisiukevich (approved):

PR #451, opened 2020-10-16
Max length reached behavior

### LINKED ISSUE #388: [Enhancement] Implement MaxLengthReachedBehavior
## Summary

I would like to have a behavior that executes a certain command (or invokes a certain event handler) when the max length of an entry field has been reached.

## API Changes

Create a new MaxLengthReachedBehavior with the following properties:

- Command: of type ICommand. This will get executed when the MaxLength of the Text property of the attached Entry field has been reached.
- MaxLengthReachedEvent: of type eventhandler. This will get invoked when the MaxLength of the Text property of the attached Entry field has been reached.
- AutoDismissKeyboard: of type boolean. Default value is true. When the commands get executes, the keyboard will be dismissed (or not).

## Intended Use Case
If your user needs to type in a certain code to verify 'something' which always contains x number of characters, I would like to automatically execute a certain command without the user having to click a button to continue.

If your user has to type in his date of birth and we use the nice MaskBehavior (which is also available in the XCT!), we could invoke a certain event handler which could set the focus of the next entry field on our form, so our user can quickly continue filling out the entire form without having to click on the next entry field manually to start typing.

## Who Will Do The Work?

jBijsterboschNL:
### Description of Change ###

Implemented MaxLengthReachedBehavior. When the MaxLength on an Entry field has been reached, any configured EventHandler gets invoked and a Command gets Executed.

### Bugs Fixed ###

- Fixes #388

### API Changes ###

Added: 

```csharp
public class MaxLengthReachedBehavior : BaseBehavior<InputView>
{
	public static readonly BindableProperty CommandProperty
		= BindableProperty.Create(nameof(Command), typeof(ICommand), typeof(UserStoppedTypingBehavior));

	public static readonly BindableProperty ShouldDismissKeyboardAutomaticallyProperty
		= BindableProperty.Create(nameof(ShouldDismissKeyboardAutomatically), typeof(bool), typeof(UserStoppedTypingBehavior), false);

	public event EventHandler<MaxLengthReachedEventArgs> MaxLengthReached;

	public ICommand Command
	{
		get => (ICommand)GetValue(CommandProperty);
		set => SetValue(CommandProperty, value);
	}

	public bool ShouldDismissKeyboardAutomatically
	{
		get => (bool)GetValue(ShouldDismissKeyboardAutomaticallyProperty);
		set => SetValue(ShouldDismissKeyboardAutomaticallyProperty, value);
	}

	...
}
``` 

### Behavioral Changes ###

None

### PR Checklist ###


jBijsterboschNL:
I've noticed that the unit test `UserStoppedTypingBehavior_Tests.ShouldDismissKeyboardWhenTimeThresholdHasExpired` fails due to a `Device.BeginInvokeOnMainThread()` call being made inside the UserStoppedTypingBehavior. @AndreiMisiukevich could you have a look at this on what to do with this one?

jBijsterboschNL:
Also I've added a `MaxLengthReachedEventArgs` class inside the *.Behaviors namespace. Not sure if this is in the right namespace? And I would also like to discuss if I should even implement an EventHandler on a Behavior or not. The reason for this is that I could add an event handler that will set focus to another input field on the screen by implementing this MaxLengtReached event on the page itself, instead of from a Command execution from the ViewModel. Maybe someone has a better suggestion for this?

XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj
```diff
@@ -214,4 +214,8 @@
   <ItemGroup>
     <Folder Include="Views\Snackbar\Helpers\WPF\" />
   </ItemGroup>
+
```
VladislavAntonyuk, line 217: probably it should be removed

XamarinCommunityToolkitSample/ViewModels/Behaviors/MaxLengthReachedBehaviorViewModel.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Windows.Input;
+using Xamarin.CommunityToolkit.Sample.Resx;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Sample.ViewModels.Behaviors
+{
+	public class MaxLengthReachedBehaviorViewModel : BaseViewModel
+	{
+		#region Properties
```
VladislavAntonyuk, line 10: do we use regions in code?

XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.ComponentModel;
+using System.Windows.Input;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Behaviors
+{
+    public class MaxLengthReachedBehavior : BaseBehavior<InputView>
+	{
+		public static readonly BindableProperty CommandProperty
+			= BindableProperty.Create(nameof(Command), typeof(ICommand), typeof(UserStoppedTypingBehavior));
```
AndreiMisiukevich, line 11: typeof(MaxLengthReachedBehavior)

XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using System.ComponentModel;
+using System.Windows.Input;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Behaviors
+{
+    public class MaxLengthReachedBehavior : BaseBehavior<InputView>
+	{
```
AndreiMisiukevich, line 9: wrong indentation here

XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs
```diff
@@ -0,0 +1,51 @@
...
+		public static readonly BindableProperty CommandProperty
+			= BindableProperty.Create(nameof(Command), typeof(ICommand), typeof(UserStoppedTypingBehavior));
+
+		public static readonly BindableProperty ShouldDismissKeyboardAutomaticallyProperty
+			= BindableProperty.Create(nameof(ShouldDismissKeyboardAutomatically), typeof(bool), typeof(UserStoppedTypingBehavior), false);
```
AndreiMisiukevich, line 14: typeof(MaxLengthReachedBehavior)

XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs
```diff
@@ -0,0 +1,51 @@
...
+
+		public event EventHandler<MaxLengthReachedEventArgs> MaxLengthReached;
```
AndreiMisiukevich, line 16: Should we use a weak event manager here?

XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs
```diff
@@ -0,0 +1,51 @@
...
+		protected override void OnViewPropertyChanged(object sender, PropertyChangedEventArgs e)
+		{
+			base.OnViewPropertyChanged(sender, e);
+			if (e.PropertyName == InputView.TextProperty.PropertyName)
+				OnTextPropertyChanged();
+		}
+
+		void OnTextPropertyChanged()
+		{
+			if (View.Text == null || View.Text.Length < View.MaxLength)
```
AndreiMisiukevich, line 39:
```
if (!(View.Text?.Length >= View.MaxLength))
    return;
```

XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs
```diff
@@ -0,0 +1,51 @@
...
+				return;
+
+			if (ShouldDismissKeyboardAutomatically)
+				View.Unfocus();
+
```
AndreiMisiukevich, line 44: I wonder should we trim Text if the Lenght is greater than MaxLength? (Keep in mind that I can copy-paste long string)

XamarinCommunityToolkit/Behaviors/MaxLengthReachedEventArgs.shared.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Xamarin.CommunityToolkit.Behaviors
+{
+	public class MaxLengthReachedEventArgs : EventArgs
+	{
+		public string Text { get; }
+
+		public MaxLengthReachedEventArgs(string text)
+		{
+			Text = text;
```
AndreiMisiukevich, line 11: => Text = text;

XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj
AndreiMisiukevich: Vlad is right. Discard this file please

XamarinCommunityToolkitSample/Pages/Behaviors/MaxLengthReachedBehaviorPage.xaml.cs
```diff
@@ -0,0 +1,19 @@
...
+	public partial class MaxLengthReachedBehaviorPage : BasePage
+	{
+		public MaxLengthReachedBehaviorPage()
+		{
+			InitializeComponent();
+		}
+
+		void MaxLengthReachedBehavior_MaxLengthReached(object sender, MaxLengthReachedEventArgs e)
+		{
+			nextEntry.Focus();
```
AndreiMisiukevich, line 16: => nextEntry.Focus();

XamarinCommunityToolkitSample/Pages/Behaviors/MaxLengthReachedBehaviorPage.xaml.cs
```diff
@@ -0,0 +1,19 @@
...
+using Xamarin.Forms.Xaml;
+
+namespace Xamarin.CommunityToolkit.Sample.Pages.Behaviors
+{
+	[XamlCompilation(XamlCompilationOptions.Compile)]
+	public partial class MaxLengthReachedBehaviorPage : BasePage
+	{
+		public MaxLengthReachedBehaviorPage()
+		{
+			InitializeComponent();
```
AndreiMisiukevich, line 11: => InitializeComponent();

XamarinCommunityToolkitSample/ViewModels/Behaviors/BehaviorsGalleryViewModel.cs
```diff
@@ -49,7 +49,11 @@ public class BehaviorsGalleryViewModel : BaseViewModel
 			new SectionModel(
 				typeof(ImpliedOrderGridBehaviorPage),
 				nameof(ImpliedOrderGridBehavior),
-				AppResources.AutoGridLayoutBehaviorDescription)
+				AppResources.AutoGridLayoutBehaviorDescription),
+			new SectionModel(
+				typeof(MaxLengthReachedBehaviorPage),
+				nameof(MaxLengthReachedBehavior),
+				AppResources.MaxLengthReachedBehaviorShortDescription)
```
AndreiMisiukevich, line 56: AppResources.MaxLengthReachedBehaviorShortDescription),

(please add a comma to the end of the line)

XamarinCommunityToolkitSample/ViewModels/Behaviors/MaxLengthReachedBehaviorViewModel.cs
AndreiMisiukevich: They are not forbidden, I believe.
And this is a sample code, so... I do not insist on removing them. But I don't see any reason to keep it here (class is so slim)

XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using System.ComponentModel;
+using System.Windows.Input;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Behaviors
+{
+    public class MaxLengthReachedBehavior : BaseBehavior<InputView>
```
pictos, line 8: Can we change this name? Because we can have something similar to label as well. Something like, when the text in the labels reaches out the limit we add "..." at the end of the string. You can see this feature [here](https://github.com/kphillpotts/ArtAuction/blob/master/src/ArtAuction/ArtAuction/CustomControl/LabelLength.xaml.cs), that time I did a custom control, but I think that it can be converted into a `behavior`.

pictos (changes requested):
Hey, @jBijsterboschNL good to see another PR from you! For now, I just have this small observation, let me know what do you think about it.

XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs
jBijsterboschNL: Hi @pictos , I like your suggestion! But the 'behavior' of your suggestion is something totally different then this one and I don't think we should combine this into 1 behavior. Because that would become confusing in my opinion that the behavior would behave differently when set on an Entry than it would behave on a Label. But your suggestion could lead to something like a `MaxLengthFormatBehavior`? And I would love to implement this, if you would like me to implement this as well (in a different PR)? Now I think of it, wouldn't a ValueConverter be a better option for your scenario? What do you guys think @jfversluis, @AndreiMisiukevich, @VladislavAntonyuk?

XamarinCommunityToolkit/Xamarin.CommunityToolkit.csproj
jBijsterboschNL: Every time I add a class to this project, it gets added like this and I have to manually set the Build Action to c# compiler. Do you guys have any idea why that is? As this time I forgot to double check the project file. Anyway, I've removed this 👍

XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs
jBijsterboschNL: Visually this is not a problem, as the Entry control only shows the value up to the MaxLength. I've added some use cases and added a Substring() to the MaxLengthBehavior :) thanks for this suggestion!

XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs
jBijsterboschNL: Hi @AndreiMisiukevich, could you elaborate a bit more on this? What the WeakEventManager exactly does and how to implement this here?

XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs
pictos: @jBijsterboschNL I'm not suggesting you implement on this PR, I'm just pointing out that this name can lead to some confusion in our users, since at least to me, it can apply in every text control that has a Text prop(editor, entry, label, button, etc) and isn't true, since this implementation is just for `InputView`.

> And I would love to implement this, if you would like me to implement this as well (in a different PR)?

Go for it! I'll love to see it!

> Now I think of it, wouldn't a ValueConverter be a better option for your scenario? What do you guys think @jfversluis, @AndreiMisiukevich, @VladislavAntonyuk?

Also, we can create an issue later on to discuss the specification.

XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs
pictos: @jBijsterboschNL [Long story](https://www.codeproject.com/Articles/738109/The-NET-Weak-Event-Pattern-in-Csharp) short with the WeakEventManager the GC can collect the `EventHandler` reference and we don't need to worry about memory leak.

The XCT has an implementation for it, you see the implementation [here](https://github.com/xamarin/XamarinCommunityToolkit/blob/main/XamarinCommunityToolkit/Helpers/WeakEventManager.shared.cs) and a sample of use [here](https://github.com/xamarin/XamarinCommunityToolkit/blob/main/XamarinCommunityToolkit/Views/Expander/Expander.shared.cs#L16)

XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs
AndreiMisiukevich: I think we can leave as is for now

pictos: @AndreiMisiukevich are you saying about the name? If it takes into consideration that a name change is a breaking change

AndreiMisiukevich: I mean this behavior triggers a command when maximum length is reached (also it doesn't allow to add more characters)

Why is this name not good?

```MaximumLengthReachedBehavior``` ?

File: XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs (hunk 6 above)
Comment (jBijsterboschNL) [reply XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs]: Thanks @pictos. I've implemented the WeakEventManager on the MaxLengthReached EventHandler

File: XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs (hunk 13 above)
Comment (jBijsterboschNL) [reply XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs]: Also all other 'read-only' controls with the MaxLength property cannot 'reach' the max length by any user input. That would clarify why this Behavior is not available for the Button and Label controls (and also explains the name for this behavior, at least for me). Also for those 'read-only' controls the actual behavior would be different (not invoking an EventHandler or executing a Command). @pictos what name would you propose then?

Comment (pictos) [reply XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs]: @jBijsterboschNL I'm terrible with names, I was hoping that you came with a great name xD. Maybe `InputMaxLengthReachedBehavior` or `InputMaxLengthBehavior`, I think that having an `Input` prefix in the name will be more clear the usage of this behavior.
But I'm ok with keeping this original name if this makes sense to you.

**Review by AndreiMisiukevich — APPROVED** [review]

**Comment by AndreiMisiukevich** [comment]
@pictos could you please take a look at his PR again? 👍

File: XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs (hunk 13 above)
Comment (AndreiMisiukevich) [reply XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs]: Hey guys, please reach an agreement regarding the final name

**Review by pictos — CHANGES_REQUESTED** [review]
@jBijsterboschNL this is more a question than a request change. When I get the MaxLengt limit and go back to the entry, and type again the keyboard isn't dismissed. Is that expected? Please see the gif below to see how I tested.

ezgif com-gif-maker

File: XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs [hunk 14]
```diff
@@ -0,0 +1,60 @@
...
+		protected override void OnViewPropertyChanged(object sender, PropertyChangedEventArgs e)
+		{
+			base.OnViewPropertyChanged(sender, e);
+			if (e.PropertyName == InputView.TextProperty.PropertyName)
+				OnTextPropertyChanged();
+		}
+
+		void OnTextPropertyChanged()
+		{
+			if (!(View.Text?.Length >= View.MaxLength))
```
pictos, line 46: Should we verify if the value is less than zero and if it's then throw an Exception?

XamarinCommunityToolkit/Behaviors/MaxLengthReachedBehavior.shared.cs
```diff
@@ -0,0 +1,60 @@
...
+				return;
+
+			if (ShouldDismissKeyboardAutomatically)
+				View.Unfocus();
+
+			var newTextValue = View.Text.Substring(0, View.MaxLength);
+
+			maxLengthReachedEventManager.RaiseEvent(this, new MaxLengthReachedEventArgs(newTextValue), nameof(MaxLengthReached));
+
```
pictos, line 55: You have 3 tabs here, can you please remove it?
```suggestion

```

jfversluis (approved):

jfversluis:
I think this has been discussed enough for now, let's merge it and see if there is anything to do after.

PR #455, opened 2020-10-17
Fix SafeAreaEffect

### LINKED ISSUE #393: [Bug] Crash: SIGABRT: Default constructor not found for type Xamarin.CommunityToolkit.iOS.Effects.SafeAreaEffectRouter
### Steps to Reproduce

1. Add an XCT safe area effect, say: `<StackLayout xcteffects:SafeAreaEffect.SafeArea="True">`
2. Run on an ACTUAL device
3. It crashes with "SIGABRT: Default constructor not found for type Xamarin.CommunityToolkit.iOS.Effects.SafeAreaEffectRouter"

### Expected Behavior

It shouldn't crash.

### Actual Behavior

It crashes.

### Basic Information

- Version with issue: All XCT versions
- Nuget Packages: All Xamarin.Forms versions

### Workaround
Don't use XCT.

Guys, this is the *third* blocker (https://github.com/xamarin/XamarinCommunityToolkit/issues/367, https://github.com/xamarin/XamarinCommunityToolkit/issues/368) when running on or building for an actual device.
I don't want to sound impolite, but has anyone actually run anything built with XCT on an *actual device*? I am wondering because XCT is touted as the upgrade path for certain features in XF 5, and this is starting to make me a bit uneasy.

Thanks.

### LINKED ISSUE #454: [Bug] SafeAreaEffect sample is broking and Linker
### Description
Here I'll report two bugs.

1. When you run the SafeAreaEffect sample and configure the `Common Language Runtime Exceptions` inside the `Exception Settings` window, to catch all exceptions you see the error below.

image

This will be fixed using the code-behind and opens an opportunity if someone wants to create a converter that uses a `boolean` to construct the `SafeArea` struct.

2. Linker is removing the `SafeAreaEffectRouter.ios` implementation since this class is constructed using reflection linker remove it from the final `.dll` (Related to the #379)

### Expected Behavior
Everything works like a charm.
### Actual Behavior
SafeAreaEffect doesn't work.

*I'll fix it*

pictos:
### Description of Change ###

Now the linker keeps the SafeAreaEffect implementation into the `.dll` and the sample page doesn't throw an exception.
### Bugs Fixed ###

- Fixes #454 
- Fixes #393 

### API Changes ###

Added a static call to the SafeArea constructor.

### Behavioral Changes ###

Nothing breaks and everyone is happy.

### PR Checklist ###

### DIFF
--- a/XamarinCommunityToolkitSample/Pages/Effects/SafeAreaEffectPage.xaml
+++ b/XamarinCommunityToolkitSample/Pages/Effects/SafeAreaEffectPage.xaml
@@ -9,10 +9,9 @@
     BackgroundColor="{StaticResource NormalButtonBackgroundColor}">
     <ContentPage.Content>
         <StackLayout
+            x:Name="stack"
             Padding="10,10"
-            effects:SafeAreaEffect.SafeArea="{Binding Path=IsToggled}"
             BackgroundColor="{StaticResource AppBackgroundColor}"
-            BindingContext="{x:Reference Name=ActivationToggle}"
             HorizontalOptions="Fill"
             Spacing="10"
             VerticalOptions="Fill">
@@ -20,7 +19,10 @@
             <StackLayout Orientation="Horizontal" Spacing="10">
                 <Label Text="{rsx:Translate ToggleSafeAreaEffect}" TextColor="{StaticResource NormalLabelTextColor}" />
 
-                <Switch x:Name="ActivationToggle" IsToggled="False" />
+                <Switch
+                    x:Name="ActivationToggle"
+                    IsToggled="False"
+                    Toggled="ActivationToggle_Toggled" />
             </StackLayout>
         </StackLayout>
     </ContentPage.Content>
--- a/XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.shared.cs
+++ b/XamarinCommunityToolkit/Effects/SafeAreaEffectRouter.shared.cs
@@ -1,4 +1,5 @@
-﻿using Xamarin.Forms;
+﻿using System;
+using Xamarin.Forms;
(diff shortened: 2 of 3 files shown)


XamarinCommunityToolkitSample/Pages/Effects/SafeAreaEffectPage.xaml
```diff
@@ -9,18 +9,20 @@
     BackgroundColor="{StaticResource NormalButtonBackgroundColor}">
     <ContentPage.Content>
         <StackLayout
+            x:Name="stack"
             Padding="10,10"
-            effects:SafeAreaEffect.SafeArea="{Binding Path=IsToggled}"
             BackgroundColor="{StaticResource AppBackgroundColor}"
-            BindingContext="{x:Reference Name=ActivationToggle}"
```
AndreiMisiukevich, line 15: Why was this changed?)

pictos: because this was throwing an exception. So I changed it to use the codebehind

image

jfversluis: That's strange. It worked for me before and I guess we want it to work through binding as well, right? But maybe have a look at that separately

jfversluis (approved):

XamarinCommunityToolkitSample/Pages/Effects/SafeAreaEffectPage.xaml
pictos: @jfversluis I think that we can create a converter or something like that

PR #470, opened 2020-10-20
[iOS] Fix NullReferenceException when MediaElement methods are called before setting Source

### LINKED ISSUE #414: [Bug] MediaElement throws NullReferenceException paper cut
### Description

`MediaElement` throws a `NullReferenceException` when calling the `Play`, `Pause`, and `Stop` methods when a `Source` hasn't been specified. Obviously this can be coded around by only calling those methods dependent upon the `CurrentState` property. So this is a paper cut rather than a bug.

Perhaps the UWP approach should be adopted, which is that "Any calls to Play, Pause, and Stop methods that occur before the MediaOpened event is raised are ignored."

daniel-shuy:
### Description of Change ###

This PR updates the iOS `MediaElementRenderer` to ignore `Play()`/`Pause()`/`Stop()` events if [AVPlayerItem](https://developer.apple.com/documentation/avfoundation/avplayeritem) is not yet set (using [constructor](https://developer.apple.com/documentation/avfoundation/avplayer/1387104-init) or [AVPlayer.replaceCurrentItem(AVPlayerItem?)](https://developer.apple.com/documentation/avfoundation/avplayer/1390806-replacecurrentitem), which is called when the Source is updated).

### Bugs Fixed ###

- Fixes #414

### API Changes ###

None

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
``` -->

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/XamarinCommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
+++ b/XamarinCommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
@@ -307,6 +307,12 @@ namespace Xamarin.CommunityToolkit.UI.Views
 
 		void MediaElementStateRequested(object sender, StateRequested e)
 		{
+			// if no Source is set, do nothing
+			if (avPlayerViewController.Player?.CurrentItem == null)
+            {
+				return;
+            }
+			
 			switch (e.State)
 			{
 				case MediaElementState.Playing:
--- a/XamarinCommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
+++ b/XamarinCommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
@@ -309,9 +309,7 @@ namespace Xamarin.CommunityToolkit.UI.Views
 		{
 			// if no Source is set, do nothing
 			if (avPlayerViewController.Player?.CurrentItem == null)
-            {
 				return;
-            }
 			
 			switch (e.State)
 			{

### CODE REVIEW

**Comment by dnfadmin** [comment]
[CLA assistant check](https://cla.dotnetfoundation.org/xamarin/XamarinCommunityToolkit?pullRequest=470) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/xamarin/XamarinCommunityToolkit?pullRequest=470) before we can accept your contribution.<br/><br/>:x: daniel-shuy [sign now](https://cla.dotnetfoundation.org/xamarin/XamarinCommunityToolkit?pullRequest=470)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/xamarin/XamarinCommunityToolkit?pullRequest=470) it.</sub>

File: XamarinCommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs [hunk 1]
```diff
@@ -307,6 +307,12 @@ protected virtual void Play()
 
 		void MediaElementStateRequested(object sender, StateRequested e)
 		{
+			// if no Source is set, do nothing
+			if (avPlayerViewController.Player?.CurrentItem == null)
+            {
+				return;
+            }
```
AndreiMisiukevich, line 314:
```suggestion
			if (avPlayerViewController.Player?.CurrentItem == null)
				return;
```

jfversluis:
@daniel-shuy thank you for this! Have you actually tried to reproduce this? 
I tried to reproduce to verify this fix, but I can't seem to be able to, so maybe it was already fixed somehow? Or am I overlooking something?

I took the current main branch, removed the `Source` from the `MediaElement` and added a button that calls the `Play` method on the `MediaElement` but nothing happens with the current code as well.

daniel-shuy:
Sorry, I made the fix based on what was reported in #414 without testing it myself, but looking closer at it I realize that the bug is actually in `Xamarin.Forms`' `MediaElement`, and it seems like the `XamarinCommunityToolkit`'s `MediaElement` doesn't have this same issue.

PR #487, opened 2020-10-22
Removes internationalization through the app

### LINKED ISSUE #464: [Improvement] Multi-language improvements
### Description
Remove multi-language for each and every page, except for the sample page to show how it works.

The reason for this being that with each feature or change we onboard, we also have to update all the different translations. Every contributor is expected to provide a sample already (preferably) but adding the need to translate them as well adds additional unnecessary work that might deter people. I would personally much rather show off the translation extension on its own little isolated sample page than throughout the entire sample app.

Happy to discuss this though if people feel otherwise.

jfversluis:
### Description of Change ###

Removes translations throughout the whole app and just shows one example on one page so we don't have the burden of translating everything.

I've kept the Settings page intact which can serve as a demo for the localization for now. We might want to consider moving that somewhere else to make it more clear it's a demo.

### Bugs Fixed ###

- Fixes #464


jfversluis:
@AndreiMisiukevich please prioritize and merge if you think it's OK since this will affect a lot of files and therefore also other people/PRs

jfversluis:
The build actually succeeded. Not sure why that's not reported here, so ignore the failing build. Although you probably won't have the ability to still merge then... Just approve, ping me and I'll push it through 😅

jfversluis:
Too late, I'm using magic powers now :P

AndreiMisiukevich:
> Too late, I'm using magic powers now :P

image

Issue #483, opened 2020-10-22
[Enhancement] Bring Back The Old Effects!


Axemasta:
## Summary

This repository used to contain a nice variety of effects that i've tended to use in various places. The old NuGet package is still available and I do still use it however it would be nice to carry those effects over to this revitalised version of the lib. 

Below is a list of some of the available effects
![List of old effects](https://user-images.githubusercontent.com/33064621/96876980-fdb01480-1470-11eb-89d1-859188b5ac63.png)

Now I think some of these effects are probably outdated however I would assume not all of them are useless. I specifically want to use `EntryRemoveBorder` and `EntryRemoveLine` in many places as it's very necessary for styling.

Was there a reason that these effects were removed from this lib? a cursory search of the issues doesn't reveal much and I haven't looked at the commit history.

Since the old NuGet package is pointing to this repository ([link](https://www.nuget.org/packages/Xamarin.Toolkit.Effects)), I think the effects should either be added back or it should be made clear they are no longer available in this lib.

## Who Will Do The Work?

- [X] I am willing to take this on myself
- [ ] Just putting this out there for someone to pick up

AndreiMisiukevich:
Don't you mind to refactor the code a little bit?
At least names :)

jfversluis:
I'm not opposed to bring them back, but I would want to check which are actually still needed. Maybe some of them are now supported by Forms out of the box? If you can update or comment here which ones you actually want to bring back, what they actually do and give us some more details, that would be cool. Thanks!

Axemasta:
Cool, I couldnt find any docs for these old effects so I've had a look at them and wildy assumed their functionality. here is a table of the old effects, what they do and if they are supported by Forms [OTB](https://www.google.com/search?q=what+does+out+fo+the+box+mean&oq=what+does+out+fo+the+box+mean&aqs=chrome..69i57j0i13i457j0i13j0i10i13i30j0i13i30l4.3194j1j4&sourceid=chrome&ie=UTF-8)

| Effect                    | Purpose                                                      | Supported Out The Box?          |
| ------------------------- | ------------------------------------------------------------ | ------------------------------- |
| EntryCapitalizeKeyboard   | Capitalize all inputs on an `Entry`                          | [ ]                             |
| EntryClear                | Adds clear button to `Entry`                                 | [X] - `ClearButtonVisibility`   |
| EntryDisableAutoCorrect   | Disables autocorrect for `Entry`                             | [ ]                             |
[...]

PR #492, opened 2020-10-23
Added Old Effects

Axemasta:
### Description of Change ###

Yesterday I raised #483, I wanted to use the `EntryNoBorderEffect` and `EntryNoUnderlineEffect` effects in my app. This meant installing an extremely old version of this library (built in 2018) OR taking the effect code out of that repository and putting it into my app. I believe these effects are both incredibly useful and they are not supported by the `Xamarin.Forms` platform out of the box.

On my issue I did some research into which old effects were useful and which have become obsolete (where forms supports them out of the box). The following effects were identified as being able to be included in this library:

- `EntryNoBorderEffect`
- `EntryNoUnderlineEffect`
- `EntrySelectAllTextEffect`

I have added the following effects in this PR:

- `RemoveBorderEffect`
- `SelectAllTextEffect`

#### `RemoveBorderEffect`

this is an effect for `Entry` which removes the border (iOS) and underline (Android) from an entry. The code for this effect has been completely taken from the old repository, no changes have been made to it at all.

**iOS**
RemoveBorderEffect on iOS

**Android**
image

#### `SelectAllTextEffect`

This is an effect for `Entry` and `Editor` which selects all text when the control is focussed. The code for this effect has been modified on both platforms from the original effect. On Android I have just cleaned up the functionality into a single method which the `PlatformEffect` methods will both call with a `bool`. On iOS the implementation is a bit more tricky and quite frankly needs some further consideration. Apply this effect to a `UITextField` is incredibly easy and I had refactored the code to be a little cleaner and more maintainable. The method `void ApplyEffect(bool apply)` would get the `UITextField` and then attach / detach a handler:
```csharp
if (apply)
{
    textField.EditingDidBegin += OnEditingDidBegin;
}
else
{
    textField.EditingDidBegin -= OnEditingDidBegin;
}
```
This would mean that the `OnEditingDidBegin` method would not need to be declared through a lambda twice.

However when I added `Editor` to the effect I complicated the implementation and its a little spaghetti in its current state. I call the following method to see if the control is supported:

```csharp
bool ControlIsSupported(Type controlType)
{
    if (controlType == null)
        throw new ArgumentException("Type was null", nameof(controlType));

    switch (controlType.Name)
    {
        case "UITextField":
        case "FormsUITextView":
            return true;
        default:
            return false;
    }
}
```
Then if my control is supported I apply the effect to the specific control type. `Entry` renders to a `UITextView` on iOS however the `Control.GetType().Name` call returns`FormsUITextView` adding to the confusion in the code. The `UITextView` does not appear to have an `EditingDidBegin` event like `UITextField`, I used the Forms element to subscribe to the `Focussed` event where I then have to get the Forms and Native object, check if the `Editor` is focussed and then select its text.

I'm not happy with the code in `SelectAllTextEffect.ios.cs` so wanted to put this PR up and get some feedback on how to better implement a renderer / effect in this way. It functions as expected so at least thats the case.

**iOS**
ios select all effect gif

**Android**
android select all effect gif

### API Changes ###

- [ ] Has tests (if omitted, state reason in description) - No tests, not sure how to test effects!


jfversluis:
Sorry for the trouble, but I did some work on the sample app and that seems to have cause some conflicts. I removed most of the resource strings etc. so it shouldn't be too hard to fix :)

Could you fix it? Thank!

XamarinCommunityToolkit/Effects/RemoveBorder/RemoveBorderEffect.android.cs
```diff
@@ -0,0 +1,28 @@
...
+		protected override void OnAttached()
+		{
+			var shape = new ShapeDrawable(new RectShape());
+			shape.Paint.Color = global::Android.Graphics.Color.Transparent;
+			shape.Paint.StrokeWidth = 0;
+			shape.Paint.SetStyle(Paint.Style.Stroke);
+			Control.Background = shape;
+		}
+
+		protected override void OnDetached()
```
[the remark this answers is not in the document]
Axemasta: Its an abstract method from the base class, `CS0534 'RemoveBorderEffect' does not implement inherited abstractmember 'Effect.OnDetached()'`

I took this from the old control, I don't think theyre handling readding the underline

XamarinCommunityToolkit/Effects/EffectIds.shared.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using Xamarin.Forms;
+
+[assembly: ResolutionGroupName("Xamarin.CommunityToolkit.Effects")]
+
+namespace Xamarin.CommunityToolkit.Effects
+{
+	public class EffectIds
```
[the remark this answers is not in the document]
Axemasta: I really don't like having magic strings around an application and prefer to put them in a common file that is filled with constants. We can do it either way but this approach leads slightly less room for error because if a developer were to add a new effect it is a lot more clear how they need to format the name of the new effect.

XamarinCommunityToolkit/Effects/RemoveBorder/RemoveBorderEffect.android.cs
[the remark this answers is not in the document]
AndreiMisiukevich: Well, then shouldn't we clear the background once the effect is unattached?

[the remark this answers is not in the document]
Axemasta: I've added:
`Control.Background = null;`
Although Im not sure if I should capture the original background (in `OnAttached`) and then set it back here. What are your thoughts?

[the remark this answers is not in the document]
AndreiMisiukevich: Hm.. good question. probably capturing the original background seems a more proper solution... 🤔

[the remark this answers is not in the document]
Axemasta: I've updated it to capture the original value. The GC should collect the old value, is it better to just call dispose on `originalBackground` after setting it on the control?

XamarinCommunityToolkit/Effects/RemoveBorder/RemoveBorderEffect.android.cs
```diff
@@ -0,0 +1,35 @@
...
+using Xamarin.CommunityToolkit.Effects;
+using Xamarin.Forms;
+using Xamarin.Forms.Platform.Android;
+using Effects = Xamarin.CommunityToolkit.Android.Effects;
+
+[assembly: ExportEffect(typeof(Effects.RemoveBorderEffect), nameof(RemoveBorderEffect))]
+
+namespace Xamarin.CommunityToolkit.Android.Effects
+{
+	[Preserve(AllMembers = true)]
```
pictos, line 14: Can you please remove this attribute?

XamarinCommunityToolkit/Effects/RemoveBorder/RemoveBorderEffect.ios.cs
```diff
@@ -0,0 +1,34 @@
...
+using Xamarin.CommunityToolkit.Effects;
+using Xamarin.Forms;
+using Xamarin.Forms.Platform.iOS;
+using Effects = Xamarin.CommunityToolkit.iOS.Effects;
+
+[assembly: ExportEffect(typeof(Effects.RemoveBorderEffect), nameof(RemoveBorderEffect))]
+
+namespace Xamarin.CommunityToolkit.iOS.Effects
+{
+	[Preserve(AllMembers = true)]
```
pictos, line 12: Can you please remove this attribute?

XamarinCommunityToolkit/Effects/SafeArea/SafeAreaEffectRouter.ios.cs
```diff
@@ -1,15 +1,16 @@
...
 using Xamarin.Forms;
 using Xamarin.Forms.Platform.iOS;
 using Effects = Xamarin.CommunityToolkit.iOS.Effects;
 
-[assembly: ResolutionGroupName(nameof(Xamarin.CommunityToolkit))]
 [assembly: ExportEffect(typeof(Effects.SafeAreaEffectRouter), nameof(SafeAreaEffectRouter))]
 
 namespace Xamarin.CommunityToolkit.iOS.Effects
 {
+	[Preserve(AllMembers =true)]
```
pictos, line 13: Can you please remove this attribute?

XamarinCommunityToolkit/Effects/SelectAllText/SelectAllTextEffect.android.cs
```diff
@@ -0,0 +1,23 @@
...
+using Xamarin.CommunityToolkit.Effects;
+using Xamarin.Forms;
+using Xamarin.Forms.Platform.Android;
+using Effects = Xamarin.CommunityToolkit.Android.Effects;
+
+[assembly: ExportEffect(typeof(Effects.SelectAllTextEffect), nameof(SelectAllTextEffect))]
+
+namespace Xamarin.CommunityToolkit.Android.Effects
+{
+	[Preserve(AllMembers = true)]
```
pictos, line 12: Can you please remove this attribute?

XamarinCommunityToolkit/Effects/SelectAllText/SelectAllTextEffect.ios.cs
```diff
@@ -0,0 +1,111 @@
...
+using Xamarin.CommunityToolkit.Effects;
+using Xamarin.Forms;
+using Xamarin.Forms.Platform.iOS;
+using Effects = Xamarin.CommunityToolkit.iOS.Effects;
+
+[assembly: ExportEffect(typeof(Effects.SelectAllTextEffect), nameof(SelectAllTextEffect))]
+
+namespace Xamarin.CommunityToolkit.iOS.Effects
+{
+    [Preserve(AllMembers = true)]
```
pictos, line 14: Can you please remove this attribute?

pictos (changes requested):
@Axemasta tks for your contribution, I don't have time to do a full review right now. Since I saw that you added the `Preserve` attribute I need to request these changes.
The reason that I'm asking you to remove the attribute is that we want this lib to be LinkerSafe, which means if the dev doesn't use a feature the linker will remove it. If you have this kind of Preserve-sh thing the linker will not remove this reference. If you are facing some issue with the linker please see [this PR](https://github.com/xamarin/XamarinCommunityToolkit/pull/455) and follows this pattern, if you have any question fell free to ping me and will answer ASAP (:

XamarinCommunityToolkit/Effects/SelectAllText/SelectAllTextEffect.ios.cs
Axemasta: done

XamarinCommunityToolkit/Effects/SelectAllText/SelectAllTextEffect.android.cs
Axemasta: done

XamarinCommunityToolkit/Effects/SafeArea/SafeAreaEffectRouter.ios.cs
Axemasta: done

XamarinCommunityToolkit/Effects/RemoveBorder/RemoveBorderEffect.ios.cs
Axemasta: done

XamarinCommunityToolkit/Effects/RemoveBorder/RemoveBorderEffect.android.cs
Axemasta: done

Axemasta:
> @Axemasta tks for your contribution, I don't have time to do a full review right now. Since I saw that you added the `Preserve` attribute I need to request these changes.
> The reason that I'm asking you to remove the attribute is that we want this lib to be LinkerSafe, which means if the dev doesn't use a feature the linker will remove it. If you have this kind of Preserve-sh thing the linker will not remove this reference. If you are facing some issue with the linker please see [this PR](https://github.com/xamarin/XamarinCommunityToolkit/pull/455) and follows this pattern, if you have any question fell free to ping me and will answer ASAP (:

Thanks for the feedback, I've updated the code accordingly. What had confused me about the existing effect was the effect & effect router class, I hadn't thought that the conditional `#if __IOS__` was the part of the code that was referencing the effect & protecting the code from the linker. Thanks for the tip, I'll remember that one in the future 😊

I'm also inclined to update the documentation once these are merged because it would be very useful thing to point people to before they contribute! 😁

jfversluis:
> I'm also inclined to update the documentation once these are merged because it would be very useful thing to point people to before they contribute! 😁

Absolutely, sorry that there was some confusion. We're still trying to find our way ourselves here, so everything is not perfect (yet? 🤔 ). And we probably have already developed a blind spot for the things that come natural to us, but are not obvious for newcomers. Please feel free to either update any documentation or open an issue describing what you ran into so we can fix it :)

In any case; thank you for your time and efforts on this!

XamarinCommunityToolkit/Effects/EffectIds.shared.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using Xamarin.Forms;
+
+[assembly: ResolutionGroupName("Xamarin.CommunityToolkit.Effects")]
+
+namespace Xamarin.CommunityToolkit.Effects
+{
+	public sealed class EffectIds
```
AndreiMisiukevich, line 8: Should we make it internal?
I don't think we should expose it for XCT users

Axemasta: `internal`? even better! I'll update it now

Axemasta: Interestingly the stylecop shouts at me here

Stylecop shouting at me

The suggested code change is:
`sealed class EffectIds`

Could there be any issues with this not being explicitly declared as `internal`? Without an accessibility modifier it defaults to `private` AFAIK. According to the [Access Modifiers documentation](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/access-modifiers):
> internal: The type or member can be accessed by any code in the same assembly, but not from another assembly.
> private: The type or member can be accessed only by code in the same class or struct.

So `internal` is the correct modifier to use here. Do I make the change with the warning, do I suppress the warning? What are your thoughts?

AndreiMisiukevich: For classes **internal** is default :) so, yes, just leave "sealed class"

Axemasta: done

AndreiMisiukevich (dismissed):
LGMT if everybody OK with having effects id in the separate file

XamarinCommunityToolkit/Effects/EffectIds.shared.cs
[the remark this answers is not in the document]
pictos: @Axemasta has a good point here, to make everything in a single file and make it easier to maintain. But otherwise isn't too hard to define a name of the effect and isn't something that we will change a lot, isn't like a `key` that can change on new releases.

@AndreiMisiukevich, said that I don't have a strong opinion on that. Sorry to be that person xD

XamarinCommunityToolkit/Effects/EffectIds.shared.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using Xamarin.Forms;
+
+[assembly: ResolutionGroupName("Xamarin.CommunityToolkit.Effects")]
+
+namespace Xamarin.CommunityToolkit.Effects
+{
+	sealed class EffectIds
+	{
+        static string effectResolutionGroupName = $"{nameof(Xamarin)}.{nameof(CommunityToolkit)}.{nameof(Effects)}";
```
pictos, line 10:
```suggestion
        static string EffectResolutionGroupName = $"{nameof(Xamarin)}.{nameof(CommunityToolkit)}.{nameof(Effects)}";
```

pictos: @jfversluis will we use this approach to do docs? Or will we set up the mdocs with a separate xml file?

If the answer from @jfversluis is a `yes`, can you add a `summary` for this?

XamarinCommunityToolkit/Effects/RemoveBorder/RemoveBorderEffect.android.cs
```diff
@@ -0,0 +1,34 @@
...
+using Xamarin.Forms.Platform.Android;
+using Effects = Xamarin.CommunityToolkit.Android.Effects;
+
+[assembly: ExportEffect(typeof(Effects.RemoveBorderEffect), nameof(RemoveBorderEffect))]
+
+namespace Xamarin.CommunityToolkit.Android.Effects
+{
+	public class RemoveBorderEffect : PlatformEffect
+	{
+		Drawable originalBackground;
```
pictos, line 16: I saw that on iOS you use the `?`, makes sense to use here as well?
```suggestion
		Drawable? originalBackground;
```

XamarinCommunityToolkit/Effects/RemoveBorder/RemoveBorderEffect.android.cs
```diff
@@ -0,0 +1,34 @@
...
+			shape.Paint.Color = global::Android.Graphics.Color.Transparent;
+			shape.Paint.StrokeWidth = 0;
+			shape.Paint.SetStyle(Paint.Style.Stroke);
+			Control.Background = shape;
+		}
+
+		protected override void OnDetached()
+		{
+			Control.Background = originalBackground;
+		}
```
pictos, line 32:
```suggestion
		protected override void OnDetached() => Control.Background = originalBackground;
```

XamarinCommunityToolkit/Effects/SafeArea/SafeAreaEffectRouter.ios.cs
```diff
@@ -1,11 +1,11 @@
 ﻿using System.Linq;
+using Foundation;
```
pictos, line 2: we don't use the `Foundation` in this class, can you remove it, please?
```suggestion

```

XamarinCommunityToolkit/Effects/SelectAllText/SelectAllTextEffect.ios.cs
```diff
@@ -0,0 +1,110 @@
...
+[assembly: ExportEffect(typeof(Effects.SelectAllTextEffect), nameof(SelectAllTextEffect))]
+
+namespace Xamarin.CommunityToolkit.iOS.Effects
+{
+    public class SelectAllTextEffect : PlatformEffect
+	{
+        protected override void OnAttached()
+        {
+            ApplyEffect(true);
+        }
```
pictos, line 19:
```suggestion
        protected override void OnAttached() => ApplyEffect(true);
```

XamarinCommunityToolkit/Effects/SelectAllText/SelectAllTextEffect.ios.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        protected override void OnDetached()
+        {
+            ApplyEffect(false);
+        }
```
pictos, line 24:
```suggestion
        protected override void OnDetached() => ApplyEffect(false);
```

XamarinCommunityToolkit/Effects/SelectAllText/SelectAllTextEffect.ios.cs
```diff
@@ -0,0 +1,110 @@
...
+            if (controlType == typeof(UITextField))
+			{
+                var textField = Control as UITextField;
+                if (textField == null)
+                    return;
+
+                if (apply)
+                {
+                    textField.EditingDidBegin += OnEditingDidBegin;
+                }
```
pictos, line 42:
```suggestion
                    textField.EditingDidBegin += OnEditingDidBegin;
```

XamarinCommunityToolkit/Effects/SelectAllText/SelectAllTextEffect.ios.cs
```diff
@@ -0,0 +1,110 @@
...
+                else
+                {
+                    textField.EditingDidBegin -= OnEditingDidBegin;
+                }
```
pictos, line 46:
```suggestion
         textField.EditingDidBegin -= OnEditingDidBegin;
```

XamarinCommunityToolkit/Effects/SelectAllText/SelectAllTextEffect.ios.cs
```diff
@@ -0,0 +1,110 @@
...
+                throw new ArgumentException("Type was null", nameof(controlType));
+
+            switch (controlType.Name)
+			{
+                case "UITextField":
+                case "FormsUITextView":
+	                return true;
+                default:
+	                return false;
+            }
```
pictos, line 107: can you change this to use the switch expression?

pictos (changes requested):
@Axemasta Tks for your hard work on this PR. Most of my suggestions are about code style, if you have time to fix the others that I didn't ask will be awesome, but if you don't have no problem, this doesn't invalidate your awesome PR!

I just ran here, on Android, and I found something that can be a bug, when I use the effect my device doesn't show the _helper tab_ that allows me to copy, paste, cut, etc. To me is a bug, because the user selects the text to perform one of these actions. please see the gif below.
ezgif com-gif-maker (2)

XamarinCommunityToolkit/Effects/RemoveBorder/RemoveBorderEffect.android.cs
Axemasta: In the iOS renderer the old style `UITextBorderStyle` is an enum so needs to be explicitly marked as nullable. Marking the `Drawable` as nullable is not necessary as it already can be `null` 😊

XamarinCommunityToolkit/Effects/SelectAllText/SelectAllTextEffect.ios.cs
Axemasta: I've not come across a switch expression before. I just had a read of the [docs](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression) and is this what you want to see?

```csharp
return controlType.Name switch
{
    "UITextField" => true,
    "FormsUITextView" => true,
    _ => false
};
```

Looks a bit wierd to me 🤪 but i'm happy to add it!

XamarinCommunityToolkit/Effects/EffectIds.shared.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using Xamarin.Forms;
+
+[assembly: ResolutionGroupName("Xamarin.CommunityToolkit.Effects")]
+
+namespace Xamarin.CommunityToolkit.Effects
+{
+	sealed class EffectIds
+	{
+        static string EffectResolutionGroupName = $"{nameof(Xamarin)}.{nameof(CommunityToolkit)}.{nameof(Effects)}";
```
Axemasta, line 10: I should not have commited this change, the field is private so should be camelCased, this also wan't changed in the rest of the class so I broke the build 🤪

Axemasta:
> I just ran here, on Android, and I found something that can be a bug, when I use the effect my device doesn't show the _helper tab_ that allows me to copy, paste, cut, etc. To me is a bug, because the user selects the text to perform one of these actions. please see the gif below.
> 
> If this the expected behavior, just let me know.

@pictos this is an interesting one, I can understand why this could be perceived as a bug. On android when you select text it will immediately popup a context menu. The same is true on iOS when you manually hold on some text, a context menu will show. Continuing our example on iOS if text is selected for you (like tapping the url on safari) the text is highlighted but the user has to tap the highlighted text to see the context menu. On iOS this can be done with a short tap. On Android when text is auto highlighted by the application, the context menu can be shown with a  long tap gesture. A short tap will select the position within the text. This is quite annoying but I'd say the context menu not showing is expected functionality. What comes next is at the least an annoying quirk and at worst a bug. I had a cursory play with showing a context menu but couldn't get anything meaningful to work.

Below is a gif of this behaviour on both platforms.
Example of highlighting text on both platforms

XamarinCommunityToolkit/Effects/RemoveBorder/RemoveBorderEffect.android.cs
pictos: Got ya, I thought that you were using the new nullable feature (:

Axemasta:
@pictos I've refactored `SelectAllTextEffect.iOS.cs` to use a switch expression. Can't say I'd come across them before but they are really cool and I can already think of a few places I can use them in a few projects, thanks for the suggestion!

pictos (approved):
@Axemasta Thank you for your hard work! Well done!

pictos:
@AndreiMisiukevich can you re-review it, when you have time?

AndreiMisiukevich (approved):
LGTM

AndreiMisiukevich:
> @AndreiMisiukevich can you re-review it, when you have time?

done

Issue #477, opened 2020-10-21
[Enhancement] Allow configuring snackbar in one place


Tommigun1980:
ACT.UserDialogs's toasts allows one to set up toast defaults in one place.

I think XCT's snackbar should separate the config from actually showing the notifications, as currently one has to send in all the styling, durations etc every time a snackbar is shown.

I'd like to set up defaults once (theming, fonts, durations etc.), and in my code just define the message and actions when I show them. One has now to make a wrapper class or utility methods for XCT's snackbar that always sends in the theming, or duplicate a ton of code.

As a pointer, this is roughly how I am setting up ACT.UserDialog's toasts:

```c#
        private void SetupToastDefaults()
        {
            ToastConfig.DefaultPosition = ToastPosition.Top;
            ToastConfig.DefaultDuration = TimeSpan.FromSeconds(4.5);
            ToastConfig.DefaultBackgroundColor = (Color)this.Resources["SomeColor"];
            ToastConfig.DefaultMessageTextColor = (Color)this.Resources["SomeOtherColor"];
            ToastConfig.DefaultActionTextColor = (Color)this.Resources["YetAnotherColor"];
        }
```

There is no equivalent for XCT's snackbar -- I'd have to send in all of those every time a notification is shown.

pictos:
I'll ping you first and ask later

@VladislavAntonyuk any thoughts?

Do you mind if I mention you in the `snackbar` issues?

VladislavAntonyuk:
It can be implemented. I can do it after finishing with DrawingView PR

Tommigun1980:
Thanks guys, I appreciate it.

VladislavAntonyuk:
in progress

AndreiMisiukevich:
Fixed in https://github.com/xamarin/XamarinCommunityToolkit/pull/508

Tommigun1980:
Thanking you here as well @VladislavAntonyuk, absolutely fantastic job once again! You truly are a wizard.

Issue #474, opened 2020-10-21
[Bug] SnackBarOptions.Actions has getter only


Tommigun1980:
Makes it cumbersome to init these as one has to use the constructor and provide all the parameters just to set some actions.
Actions should have a setter also.

VladislavAntonyuk:
We like immutability. Use ctor

Tommigun1980:
> We like immutability. Use ctor

Eh. I won't. It's so much clearer to chain construction such as
```c#
new A()
{
   1, 3 4,
   new B()
   {
      1, 2, 3
   }
}
```

etc, than to send in everything via constructors. Why would an entire API disregard a common pattern just because someone doesn't like it? Also, every other property has a setter than Actions, why is it special?

Tommigun1980:
> We like immutability. Use ctor

Well that's awfully smug. I won't. It's so much clearer to chain construction such as

```c#
new Something()
{
   A = 1,
   B = new SomethingElse()
   {
      D = 1,
      E = 2,
      F = new List<int>()
      {
         1, 2, 3
      }
   }
}
```

etc, than to send in everything via constructors. Why would an entire API disregard a common pattern just because someone doesn't like it? Also, every other property has a setter than Actions, why is it special?

jfversluis:
I don't know who "we" is in this case, but as far as I am concerned everything can be discussed :) Not sure if this is a bug though. Technically it works as intended atm

Does adding a setter make for any side-effects as far as we can tell right now?

Tommigun1980:
> I don't know who "we" is in this case, but as far as I am concerned everything can be discussed :) Not sure if this is a bug though. Technically it works as intended atm
> 
> Does adding a setter make for any side-effects as far as we can tell right now?

Well you are right, maybe it's a feature request then. I thought it was a bug because other properties in the same class has setters except Actions, so it looked like a bug to me.

Tommigun1980:
Any chance this could get corrected at some point? Construction of SnackBar is quite ugly atm as some objects (MessageOptions) the constructor requires has setters only, so you are forced to combine all kinds of different initialisation styles. Instead of being able to chain everything as in the OP, one has to chain SOME of the objects and ultimately send them in to the constructor, as its Actions property does not have a setter. Everything else has setters and some don't even have constructor arguments so it really is a special case.

[...]

PR #508, opened 2020-10-28
#474 Add setters to the snackbar options, #477 Snackbar default settings

VladislavAntonyuk:
### Description of Change ###

Add setters to the snackbar options
Add Snackbar default settings

### Bugs Fixed ###
#474
#477 

### API Changes ###

Added: 
 
- static default properties for all options

Changed:

 - `ActionOptions` => `ToastOptions`
 - removed ctors for SnackBarOptions and ToastOptions. they were rather complex and required all options.

### PR Checklist ###


VladislavAntonyuk:
@jfversluis, @AndreiMisiukevich could you please review? it has some breaking changes, it would be nice to fix them before release of stable version.

AndreiMisiukevich (dismissed):
Looks good
Only one question. Is it important to rename "ActionOptions => ToastOptions"?
If others are OK with such renaming, I am OK too

VladislavAntonyuk:
> Looks good
> Only one question. Is it important to rename "ActionOptions => ToastOptions"?
> If others are OK with such renaming, I am OK too

We use it with Toast. so ToastOptions is what user expect to use. ActionOptions is something really abstract. What is action for you? I think we should not associated it with Toast, SnackBar etc.

pictos:
I will review this until tomorrow.

XamarinCommunityToolkit/Views/SnackBar/Options/ToastOptions.shared.cs
```diff
@@ -0,0 +1,48 @@
...
+		/// Is Right to left
+		/// </summary>
+		public bool IsRtl { get; set; } = DefaultIsRtl;
+
+		public static bool DefaultIsRtl { get; set; } = false;
+
+		/// <summary>
+		/// The duration for the SnackBar.
+		/// </summary>
+		public int Duration { get; set; } = DefaultDuration;
```
pictos, line 37: Can we add the unit here, in the name of the property, or/and in the `summary` part?

XamarinCommunityToolkit/Views/SnackBar/Options/ToastOptions.shared.cs
```diff
@@ -0,0 +1,48 @@
...
+
+		public static int DefaultDuration { get; set; } = 3000;
```
pictos, line 39: Can we add the unit here? (ms, us, seconds)

pictos (changes requested):
@VladislavAntonyuk nice work, as always! I have just some considerations about names, to make it more clear to our end users. Please let me know what do you think.

VladislavAntonyuk:
> @VladislavAntonyuk nice work, as always! I have just some considerations about names, to make it more clear to our end users. Please let me know what do you think.

Thank you! I found that XF has Duration property of MediaElement. To be consistent with it I think we can leave it Duration. But what do you think if I change the type to TimeSpan?

pictos:
@VladislavAntonyuk I like the idea of TimeSpan.

pictos (approved):

AndreiMisiukevich (approved):

Tommigun1980:
Thank you so much once again @VladislavAntonyuk! Looks fantastic.

PR #551, opened 2020-11-12
[Validation] MultiValidationBehavior with TextValidationBehavior : Regex init fix

### LINKED ISSUE #550: [Bug] TextValidationBehavior does not work in MultiValidationBehavior
### Description

Using CharactersValidationBehavior inside a MultiValidationBehavior does not work properly : the validation is always invalid.
After some search, its due to TextValidationBehavior, the initialisation of the `regex` is not made properly in that case.

### Steps to Reproduce

1. Put an Entry
2. Add MultiValidationBehavior on it
3. Add 1 or more CharactersValidationBehavior

### Expected Behavior
The validation it's made correctly

### Actual Behavior
Always invalid

### Basic Information

- Version with issue: main
- Last known good version: Never
- IDE: Version 16.8.0

### Workaround
No workaround founded (maybe put a regex on the CharactersValidationBehavior)

Seuleuzeuh:
### Description of Change ###

The method `OnRegexPropertyChanged()` was called in `OnAttachedTo` method but when a TextValidationBehavior (or one of his child) is used in MultiValidationBehavior this method was never called.
So `regex` property is never initialized and validation failed.

<s>I've moved the call of OnRegexPropertyChanged in the ctor of the TextValidationBehavior.</s>
Adding new methods on BaseBehavior OnAttached & OnDeattached, called on the OnAttachedTo & OnDetachingFrom.
I've added a sample for that case (Password validation sample using multiple CharactersValidationBehavior).

Screenshot_1605194158

### Bugs Fixed ###

- Fixes #550 

### API Changes ###

Added:
- `OnAttached` method on BaseBehavior.
- `OnDeattached` method on BaseBehavior.

Methods called in ``OnAttachedTo` & `OnDetachingFrom` of the BaseBehavior class.
In the MultiValidationBehavior methods called when ValidationBehavior is Added or Remove of the children.

### PR Checklist ###

### DIFF
--- a/XamarinCommunityToolkit/Behaviors/BaseBehavior.shared.cs
+++ b/XamarinCommunityToolkit/Behaviors/BaseBehavior.shared.cs
@@ -37,12 +37,6 @@ namespace Xamarin.CommunityToolkit.Behaviors
 				};
 				SetBinding(BindingContextProperty, defaultBindingContextBinding);
 			}
-
-			OnAttached();
-		}
-
-		internal virtual void OnAttached()
-		{
 		}
 
 		protected override void OnDetachingFrom(TView bindable)
@@ -57,12 +51,6 @@ namespace Xamarin.CommunityToolkit.Behaviors
 
 			bindable.PropertyChanged -= OnViewPropertyChanged;
 			View = null;
-
-			OnDeattached();
-		}
-
-		internal virtual void OnDeattached()
-		{
 		}
 
 		protected bool IsBound(BindableProperty property, BindingBase defaultBinding = null)
--- a/XamarinCommunityToolkit/Behaviors/BaseBehavior.shared.cs
+++ b/XamarinCommunityToolkit/Behaviors/BaseBehavior.shared.cs
@@ -37,6 +37,12 @@ namespace Xamarin.CommunityToolkit.Behaviors
 				};
 				SetBinding(BindingContextProperty, defaultBindingContextBinding);
 			}
+
+			OnAttached();
+		}
+
+		internal virtual void OnAttached()
+		{
 		}
 
 		protected override void OnDetachingFrom(TView bindable)
@@ -51,6 +57,12 @@ namespace Xamarin.CommunityToolkit.Behaviors
 
 			bindable.PropertyChanged -= OnViewPropertyChanged;
 			View = null;
+
+			OnDeattached();
+		}
+
+		internal virtual void OnDeattached()
+		{
 		}
 
 		protected bool IsBound(BindableProperty property, BindingBase defaultBinding = null)
--- a/XamarinCommunityToolkit/Behaviors/Validators/MultiValidationBehavior.shared.cs
+++ b/XamarinCommunityToolkit/Behaviors/Validators/MultiValidationBehavior.shared.cs
@@ -68,17 +68,13 @@ namespace Xamarin.CommunityToolkit.Behaviors
 						Path = BindingContextProperty.PropertyName,
 						Source = this
 					});
-					child.OnAttached();
 				}
 			}
 
 			if (e.OldItems != null)
 			{
 				foreach (var child in e.OldItems.OfType<ValidationBehavior>())
-				{
 					child.RemoveBinding(BindingContextProperty);
-					child.OnDeattached();
-				}
 			}
 		}
 	}
--- a/XamarinCommunityToolkit/Behaviors/Validators/MultiValidationBehavior.shared.cs
+++ b/XamarinCommunityToolkit/Behaviors/Validators/MultiValidationBehavior.shared.cs
@@ -68,13 +68,17 @@ namespace Xamarin.CommunityToolkit.Behaviors
 						Path = BindingContextProperty.PropertyName,
 						Source = this
 					});
+					child.OnAttached();
 				}
 			}
 
 			if (e.OldItems != null)
 			{
 				foreach (var child in e.OldItems.OfType<ValidationBehavior>())
+				{
 					child.RemoveBinding(BindingContextProperty);
+					child.OnDeattached();
+				}
 			}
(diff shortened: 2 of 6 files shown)


Seuleuzeuh:
Adding simple tests for the MultiValidationBehavior for this particular use case (CharactersValidationBehavior inside a MultiValidationBehavior).

jfversluis:
Loving it, thank you so much!

AndreiMisiukevich (commented):
Hello and thank you very much for your effort in making this lib a little better :)

I have a question. Don't you think it would make sense to apply this fix in MultiValidationBehavior

When a child is added we would call ```child.OnAttached()```
When a child is removed we would call ```child.OnDeattached()```

for achieving that, we would add a couple of new **internal virtual** methods to BaseValidationBehavior ```OnAttached``` and ```OnDeattached```

And we would use those methods in all places instead of **protected** ones

CC @jfversluis @pictos

I am fine with mergin it as is now as "hotfix" but I propose considering my idea

Seuleuzeuh:
That was my first idea at first, using the base interface IAttachedObject but it's Forms Internal so I've managed to moved the call. I've not thinking of call the base Beau of XCT instead, it's a great idea.
If you are ok with it, I can make the change.

AndreiMisiukevich:
> That was my idea at first, using the base interface IAttachedObject but it's Forms Internal so I've managed to moved the call. I've not thinking of call the base Behavior of XCT instead, it's a great idea.
> If you are ok with it, I can make the change.

if you wish, it would be great :)

Seuleuzeuh:
@AndreiMisiukevich It's done. Take a look and tell me if you thought the implementation like that.

XamarinCommunityToolkit/Behaviors/BaseBehavior.shared.cs
```diff
@@ -37,6 +37,12 @@ protected override void OnAttachedTo(TView bindable)
 				};
 				SetBinding(BindingContextProperty, defaultBindingContextBinding);
 			}
+
+			OnAttached();
+		}
+
+		internal virtual void OnAttached()
+		{
 		}
```
AndreiMisiukevich, line 46: This method must contain all code that the currently **protected** method contains.
The protected method will be pretty slim

```csharp
protected override void OnAttachedTo(TView bindable)
{
	base.OnAttachedTo(bindable);
	OnAttached(bindable)
}
```

XamarinCommunityToolkit/Behaviors/BaseBehavior.shared.cs
```diff
@@ -51,6 +57,12 @@ protected override void OnDetachingFrom(TView bindable)
 
 			bindable.PropertyChanged -= OnViewPropertyChanged;
 			View = null;
+
+			OnDeattached();
+		}
+
+		internal virtual void OnDeattached()
+		{
 		}
```
AndreiMisiukevich, line 66: The same for this method

XamarinCommunityToolkit/Behaviors/Validators/MultiValidationBehavior.shared.cs
```diff
@@ -68,13 +68,17 @@ void OnChildrenCollectionChanged(object sender, NotifyCollectionChangedEventArgs
 						Path = BindingContextProperty.PropertyName,
 						Source = this
 					});
+					child.OnAttached();
```
AndreiMisiukevich, line 71: Yeah, exactly :)

XamarinCommunityToolkit/Behaviors/Validators/MultiValidationBehavior.shared.cs
```diff
@@ -68,13 +68,17 @@ void OnChildrenCollectionChanged(object sender, NotifyCollectionChangedEventArgs
...
 				}
 			}
 
 			if (e.OldItems != null)
 			{
 				foreach (var child in e.OldItems.OfType<ValidationBehavior>())
+				{
 					child.RemoveBinding(BindingContextProperty);
+					child.OnDeattached();
```
AndreiMisiukevich, line 80: 👍

XamarinCommunityToolkit/Behaviors/Validators/TextValidationBehavior.shared.cs
```diff
@@ -53,16 +53,20 @@ public RegexOptions RegexOptions
...
 		{
+			base.OnAttached();
 			OnRegexPropertyChanged();
-			base.OnAttachedTo(bindable);
 		}
 
+		protected virtual string DefaultRegexPattern => string.Empty;
+
+		protected virtual RegexOptions DefaultRegexOptions => RegexOptions.None;
+
```
AndreiMisiukevich, line 69: This file should contain only one change

```-``` ```protected override void OnAttachedTo(View bindable)```
```+``` ```internal override void OnAttached(View bindable)```

**Review by AndreiMisiukevich — CHANGES_REQUESTED** [review]
Actually, I think that a new internal method should appear inside ValidationBehavior instead of BaseBehavior.

Can we try to overwrite the old protected method with a new internal?

```csharp
internal new virtual void OnAttachedTo(TView bindable)
{
	base.OnAttachedTo(bindable);
	// Content
}
```

AndreiMisiukevich (changes requested):
Well, I should apologize for the bad bits of advice.
I investigated code a little more and come up with the decision that your initial solution is the best fit for us.

Changes only in TextValidationBehavior
1) Remove OnAttachedTo overridden method
2) Add a constructor with OnRegexPropertyChanged invocation

Please ignore my other comments

Seuleuzeuh:
@AndreiMisiukevich Reverted changes to the initial PR code.

AndreiMisiukevich (approved):

PR #566, opened 2020-11-18
TouchEffect (Press & Hover visual handling) + LongPress

### LINKED ISSUE #200: [Enhancement] TouchEffect
## Summary

Import https://github.com/AndreiMisiukevich/TouchEffect

Hey folks (especially @Char0394) do you think I should import this one into the ToolKit as well?

## Intended Use Case
Responsive touches

### LINKED ISSUE #500: Long tap gesture behaviour

### LINKED ISSUE #561: [Enhancement] Down and Up gesture
## Summary

2 new gestures -
- Down (triggered when the user touches the screen)
- Up (triggered when the user is no longer touching the screen)

In one of the controls I built for a personal project, I've built a similar thing that may help the implementation -

- Android - https://github.com/IeuanWalker/Xamarin.Forms.FlexButton/blob/master/Flex.Android/CustomRenderers/GestureFrameRenderer.cs
- iOS - https://github.com/IeuanWalker/Xamarin.Forms.FlexButton/blob/master/Flex.iOS/CustomRenderers/GestureFrameRenderer.cs

Could link this up with the work on #500 

## Intended Use Case
- Would allow me to remove system specific code in some projects, i.e. one linked above.
- Could be used for user interaction i.e. user could hold down on an object to trigger an animation i.e. 
99miles_playback_komoot_dribbble_4

## Who Will Do The Work?

- [x] Just putting this out there for someone to pick up

AndreiMisiukevich:
### Description of Change ###
Moving TouchEffect from https://github.com/AndreiMisiukevich/TouchEffect to XCT 

Nov-25-2020 16-32-34

### Bugs Fixed ###
- Fixes #200 
- Fixes #500 
- Fixes #561 

### API Changes ###
https://github.com/AndreiMisiukevich/TouchEffect/blob/master/README.md

### Behavioral Changes ###
None

### PR Checklist ###


XamarinCommunityToolkit/Effects/Touch/GestureManager.shared.cs
```diff
@@ -0,0 +1,548 @@
...
+			if (duration <= 0)
+			{
+				control.BackgroundColor = color;
+				return;
+			}
+
+			var animationCompletionSource = new TaskCompletionSource<bool>();
+			new Animation
+            {
+				{ 0, 1,  new Animation(v => control.BackgroundColor = new Color(v, control.BackgroundColor.G, control.BackgroundColor.B, control.BackgroundColor.A), control.BackgroundColor.R, color.R) },
```
pictos, line 314: to make it better, I think that we could move the `control.BackgroundColor` or each `control.BackgroundColor.X` into a local variable. What do you think?

XamarinCommunityToolkit/Effects/Touch/GestureManager.shared.cs
```diff
@@ -0,0 +1,548 @@
...
+				return;
+
+			var opacity = normalOpacity;
+
+			if (touchState == TouchState.Pressed)
+				opacity = pressedOpacity;
+			else if (hoverState == HoverState.Hovered && sender.Control.IsSet(TouchEffect.HoveredOpacityProperty))
+				opacity = hoveredOpacity;
+
+			await sender.Control.FadeTo(opacity, (uint)Abs(duration), easing);
```
pictos, line 340:
```suggestion
			await sender.Control.FadeTo(opacity, (uint)Abs(duration), easing).ConfigureAwait(false);
```

pictos: I'm wondering if makes sense to warning the dev that `duration` is negative (in some edge case), something like.

```csharp

			if (duration < 0)
				Xamarin.Forms.Internals.Log.Warning("GestureManager", $"Attention! {nameof(duration)} is negative.");
```

XamarinCommunityToolkit/Effects/Touch/GestureManager.shared.cs
```diff
@@ -0,0 +1,548 @@
...
+			{
+				if (double.IsNaN(v))
+					return;
+
+				control.Scale = v;
+			}, control.Scale, scale, 16, (uint)Abs(duration), easing, (v, b) => tcs.SetResult(b));
+			await tcs.Task;
+		}
+
+		async Task SetTranslationAsync(TouchEffect sender, TouchState touchState, HoverState hoverState, int duration, Easing easing)
```
pictos, line 373: we don't need to have `async` here, we can just return the Task.

XamarinCommunityToolkit/Effects/Touch/GestureManager.shared.cs
```diff
@@ -0,0 +1,548 @@
...
+
+			if (touchState == TouchState.Pressed)
+				opacity = pressedOpacity;
+			else if (hoverState == HoverState.Hovered && sender.Control.IsSet(TouchEffect.HoveredOpacityProperty))
+				opacity = hoveredOpacity;
+
+			await sender.Control.FadeTo(opacity, (uint)Abs(duration), easing);
+		}
+
+		async Task SetScaleAsync(TouchEffect sender, TouchState touchState, HoverState hoverState, int duration, Easing easing)
```
pictos, line 343: we don't need to have `async` here, we can just return the Task.

XamarinCommunityToolkit/Effects/Touch/PlatformTouchEffect.android.cs
```diff
@@ -0,0 +1,340 @@
...
+
+			if (effect?.IsDisabled ?? true)
+                return;
+
+			if (IsAccessibilityMode)
+				return;
+
+			switch (e.Event.ActionMasked)
+			{
+				case MotionEventActions.Down:
```
pictos, line 162: What do you think about moving the large `cases` into local functions? To me will be easy to read.

XamarinCommunityToolkit/Effects/Touch/PlatformTouchEffect.android.cs
```diff
@@ -0,0 +1,340 @@
...
+					var diffX = Math.Abs(e.Event.GetX() - startX) / View.Context.Resources.DisplayMetrics.Density;
+					var diffY = Math.Abs(e.Event.GetY() - startY) / View.Context.Resources.DisplayMetrics.Density;
+					var maxDiff = Math.Max(diffX, diffY);
+					var disallowTouchThreshold = effect.DisallowTouchThreshold;
+					if (disallowTouchThreshold > 0 && maxDiff > disallowTouchThreshold)
+					{
+						HandleEnd(TouchStatus.Canceled);
+						return;
+					}
+					var view = sender as AView;
```
pictos, line 192: Just to double check.
Are we sure that this value will never be `null`?

XamarinCommunityToolkit/Effects/Touch/PlatformTouchEffect.android.cs
```diff
@@ -0,0 +1,340 @@
...
+		}
+
+		ColorStateList GetColorStateList()
+		{
+			int color;
+			var defaultcolor = effect.NativeAnimationColor;
+			if (defaultcolor != Forms.Color.Default)
+				color = defaultcolor.ToAndroid();
+			else
+				color = Color.Argb(64, 128, 128, 128);
```
pictos, line 307: Can we make this a const?

XamarinCommunityToolkit/Effects/Touch/PlatformTouchEffect.android.cs
```diff
@@ -0,0 +1,340 @@
...
+
+			viewOverlay.Right = group.Width;
+			viewOverlay.Bottom = group.Height;
+		}
+
+		sealed class AccessibilityListener : Java.Lang.Object,
+											 AccessibilityManager.IAccessibilityStateChangeListener,
+											 AccessibilityManager.ITouchExplorationStateChangeListener
+		{
+			readonly PlatformTouchEffect platformTouchEffect;
```
pictos, line 328: Since here we are in the Android world, I think that we can make this a `WeakReference`. In my head, this can avoid memory leaks.

XamarinCommunityToolkit/Effects/Touch/PlatformTouchEffect.ios.cs
```diff
@@ -0,0 +1,201 @@
...
+			gesture?.Dispose();
+			gesture = null;
+			effect.Control = null;
+			effect = null;
+		}
+	}
+
+	sealed class TouchUITapGestureRecognizer : UIGestureRecognizer
+	{
+		TouchEffect effect;
```
pictos, line 50: can we make this a `WeakReference`, the same motivation as Android listener.

XamarinCommunityToolkit/Effects/Touch/PlatformTouchEffect.macos.cs
```diff
@@ -0,0 +1,188 @@
...
+using Foundation;
+using Xamarin.Forms;
+using Xamarin.CommunityToolkit.macOS.Effects;
+using Xamarin.CommunityToolkit.Effects;
+
+[assembly: ExportEffect(typeof(PlatformTouchEffect), nameof(TouchEffect))]
+
+namespace Xamarin.CommunityToolkit.macOS.Effects
+{
+	[Preserve(AllMembers = true)]
```
pictos, line 12: can we remove this? Since we want to make sure that the Linker will remove unused code.

XamarinCommunityToolkit/Effects/Touch/PlatformTouchEffect.macos.cs
```diff
@@ -0,0 +1,188 @@
...
+
+			gesture?.Dispose();
+			gesture = null;
+		}
+	}
+
+	sealed class MouseTrackingView : NSView
+	{
+		NSTrackingArea trackingArea;
+		TouchEffect effect;
```
pictos, line 56: `WeakReference`

XamarinCommunityToolkit/Effects/Touch/PlatformTouchEffect.tizen.cs
```diff
@@ -0,0 +1,166 @@
...
+using Xamarin.Forms.Platform.Tizen;
+using Xamarin.CommunityToolkit.Tizen.Effects;
+using Xamarin.CommunityToolkit.Effects;
+using EColor = ElmSharp.Color;
+
+[assembly: ExportEffect(typeof(PlatformTouchEffect), nameof(TouchEffect))]
+
+namespace Xamarin.CommunityToolkit.Tizen.Effects
+{
+	[Preserve(AllMembers = true)]
```
pictos, line 13: Can you remove this, please?

XamarinCommunityToolkit/Effects/Touch/TouchEffect.shared.cs
```diff
@@ -0,0 +1,1208 @@
...
+
+		public static readonly BindableProperty ShouldMakeChildrenInputTransparentProperty = BindableProperty.CreateAttached(
+			nameof(ShouldMakeChildrenInputTransparent),
+			typeof(bool),
+			typeof(TouchEffect),
+			true,
+			propertyChanged: (bindable, oldValue, newValue) =>
+			{
+				GetFrom(bindable)?.SetChildrenInputTransparent((bool)newValue);
+				TryGenerateEffect(bindable, oldValue, newValue);
```
pictos, line 37: Can we move this, and others like this, to an external method?

XamarinCommunityToolkit/Effects/Touch/TouchEffect.shared.cs
```diff
@@ -0,0 +1,1208 @@
...
+
+		VisualElement control;
+
+		public TouchEffect()
+			: base(EffectIds.TouchEffect)
+		{
+#if __ANDROID__
+			if (System.DateTime.Now.Ticks < 0)
+				_ = new Xamarin.CommunityToolkit.Android.Effects.PlatformTouchEffect();
+#elif __IOS__
```
pictos, line 571: You can add `macOS` here as well.

XamarinCommunityToolkit/Effects/Touch/TouchEffect.shared.cs
```diff
@@ -0,0 +1,1208 @@
...
+		public Aspect PressedBackgroundImageAspect => GetPressedBackgroundImageAspect(Control);
+
+		public bool ShouldSetImageOnAnimationEnd => GetShouldSetImageOnAnimationEnd(Control);
+
+		internal bool CanExecute
+			=> IsAvailable &&
+			Control.IsEnabled &&
+			(Command?.CanExecute(CommandParameter) ?? true);
+
+		internal VisualElement Control
```
pictos, line 1084: I'm not sure about this name... On Xamarin we use `Control` to specify the `Native Control`. Since this is a `VisualElement` we suggest that we use another name, like: `Element`, `View`, etc. Thoughts?

XamarinCommunityToolkit/Effects/Touch/TouchEffect.shared.cs
```diff
@@ -0,0 +1,1208 @@
+﻿using System.Windows.Input;
+using Xamarin.Forms;
+using System;
+using System.Linq;
+
+namespace Xamarin.CommunityToolkit.Effects
+{
+	public class TouchEffect : RoutingEffect
```
pictos, line 8: This class is very large. What do you think to split it into `partial classes`? We can name the file in a way that it's easy to find what it contains, like `TouchEffectProperties.shared.cs`, `TouchEffectTouchHandler.shared.cs` and so on.

pictos (commented):
Dude, this is huge work, well done! I didn't run the code, since it's a Draft PR, but I did a look into the code and suggest some modifications. Please let me know what do you think.

XamarinCommunityToolkit/Effects/Touch/GestureManager.shared.cs
AndreiMisiukevich: Unfortunately, we cannot do it because 4 parallel animations change different color parameters R G B A.
We should use the relevant value directly from the control's background color

XamarinCommunityToolkit/Effects/Touch/PlatformTouchEffect.android.cs
AndreiMisiukevich: Yeah, 100% cuz we subscribe it as ```View.Touch += OnTouch;``` where Views is AView

XamarinCommunityToolkit/Effects/Touch/PlatformTouchEffect.ios.cs
AndreiMisiukevich: We clear this reference in the Dispose method. Are you sure we need it?

XamarinCommunityToolkit/Effects/Touch/PlatformTouchEffect.android.cs
AndreiMisiukevich: Worked around via Dispose method (clear this ref there. Made it non-read-only)

XamarinCommunityToolkit/Effects/Touch/PlatformTouchEffect.macos.cs
AndreiMisiukevich: Clear it in Dispose method

XamarinCommunityToolkit/Effects/Touch/TouchEffect.shared.cs
AndreiMisiukevich: What's the preprocessor directive for macOS and Tizen and UWP?

XamarinCommunityToolkit/Effects/Touch/TouchEffect.shared.cs
AndreiMisiukevich: Do you really think it will help?) Almost all of these lines are related to API (define various properties)
if we will try to do it, TouchEffectProperties will contain 900/1000 lines of code 😄

jfversluis:
Haha wow that bot has been crazy here 😆 sorry about that!

jfversluis:
C'monnnnn GitHub! Not all those delete entries. Then I might as well just left them there 🤦

pictos:
@AndreiMisiukevich this PR also solves the #500 ?

pictos:
@AndreiMisiukevich we merged a huge change, can you solve the conflicts?

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/PlatformTouchEffect.uwp.cs
```diff
@@ -0,0 +1,207 @@
...
+		{
+			effect = TouchEffect.PickFrom(Element);
+			if (effect?.IsDisabled ?? true)
+				return;
+
+			effect.Element = Element as VisualElement;
+			if (effect.NativeAnimation)
+			{
+				var nativeControl = Container;
+				if (String.IsNullOrEmpty(nativeControl.Name))
```
pictos, line 35: this is for me, just to make me happy xP
```suggestion
				if (string.IsNullOrEmpty(nativeControl.Name))
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/PlatformTouchEffect.uwp.cs
```diff
@@ -0,0 +1,207 @@
...
+					nativeControl.Name = Guid.NewGuid().ToString();
+
+				if (nativeControl.Resources.ContainsKey("PointerDownAnimation"))
+				{
+					pointerDownStoryboard = (Storyboard)nativeControl.Resources["PointerDownAnimation"];
+				}
```
pictos, line 41:
```suggestion
				if (nativeControl.Resources.ContainsKey("PointerDownAnimation"))
					pointerDownStoryboard = (Storyboard)nativeControl.Resources["PointerDownAnimation"];
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/PlatformTouchEffect.uwp.cs
```diff
@@ -0,0 +1,207 @@
...
+					var downThemeAnimation = new PointerDownThemeAnimation();
+					Storyboard.SetTargetName(downThemeAnimation, nativeControl.Name);
+					pointerDownStoryboard.Children.Add(downThemeAnimation);
+					nativeControl.Resources.Add(new KeyValuePair<object, object>("PointerDownAnimation", pointerDownStoryboard));
+				}
+
+				if (nativeControl.Resources.ContainsKey("PointerUpAnimation"))
+				{
+					pointerUpStoryboard = (Storyboard)nativeControl.Resources["PointerUpAnimation"];
+				}
```
pictos, line 54:
```suggestion
				if (nativeControl.Resources.ContainsKey("PointerUpAnimation"))
					pointerUpStoryboard = (Storyboard)nativeControl.Resources["PointerUpAnimation"];
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/PlatformTouchEffect.uwp.cs
```diff
@@ -0,0 +1,207 @@
...
+				Container.PointerCanceled -= OnPointerCanceled;
+				Container.PointerExited -= OnPointerExited;
+				Container.PointerEntered -= OnPointerEntered;
+				Container.PointerCaptureLost -= OnPointerCaptureLost;
+
+				isPressed = false;
+			}
+		}
+
+		private void OnPointerEntered(object sender, PointerRoutedEventArgs e)
```
pictos, line 96:
```suggestion
		void OnPointerEntered(object sender, PointerRoutedEventArgs e)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/PlatformTouchEffect.uwp.cs
```diff
@@ -0,0 +1,207 @@
...
+			effect?.HandleHover(HoverStatus.Entered);
+
+			if (isPressed)
+			{
+				effect?.HandleTouch(TouchStatus.Started);
+				AnimateTilt(pointerDownStoryboard);
+			}
+		}
+
+		private void OnPointerExited(object sender, PointerRoutedEventArgs e)
```
pictos, line 110:
```suggestion
		void OnPointerExited(object sender, PointerRoutedEventArgs e)
```

pictos: please remove other privates as well (:

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/TouchEffect.shared.cs
```diff
@@ -0,0 +1,1103 @@
+﻿using System.Windows.Input;
+using Xamarin.Forms;
+using System;
+using System.Linq;
+
+namespace Xamarin.CommunityToolkit.Effects
+{
+	public class TouchEffect : RoutingEffect
+	{
+		public event EventHandler<TouchStatusChangedEventArgs> StatusChanged;
```
pictos, line 10: Should we use `WeakeEventManager`? Just wondering if makes sense to make this a default on XCT.
@brminnick thoughts?

pictos (changes requested):
@AndreiMisiukevich nice work, I tested the `LongPressCommand` and doesn't work. I don't know if I did something wrong, but the GC got crazy. (UWP)

image

```xml
<StackLayout HorizontalOptions="Fill" Padding="24" BackgroundColor="LightGray"
             xct:TouchEffect.NativeAnimation="True"
             xct:TouchEffect.NativeAnimationShadowRadius="80"
             xct:TouchEffect.NativeAnimationColor="Red"
             xct:TouchEffect.LongPressDuration="2"
             xct:TouchEffect.LongPressCommand="{Binding Command, Source={x:Reference Page}}">
    <Label TextColor="Black" HorizontalTextAlignment="Center" Text="LongPressCommand"/>
    <Label TextColor="Black" HorizontalTextAlignment="Center" Text="NativeAnimation=true"/>
    <Label TextColor="Black" HorizontalTextAlignment="Center" Text="NativeAnimationColor=Red"/>
    <Label TextColor="Black" HorizontalTextAlignment="Center" Text="NativeAnimationShadowRadius=80"/>
</StackLayout>
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/PlatformTouchEffect.uwp.cs
AndreiMisiukevich: Me too 👍

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/TouchEffect.shared.cs
AndreiMisiukevich: Anyway, I think it can't be worse if we add it.

Added

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/GestureManager.shared.cs
```diff
@@ -0,0 +1,548 @@
...
+				await GetAnimationTask(sender, state, hoverState).ConfigureAwait(false);
+				return;
+			}
+			do
+			{
+				var rippleState = isToggled.HasValue && isToggled.Value
+					? TouchState.Normal
+					: TouchState.Pressed;
+
+				await GetAnimationTask(sender, rippleState, hoverState).ConfigureAwait(false);
```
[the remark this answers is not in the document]
AndreiMisiukevich: xct:TouchEffect.LongPressDuration="2" -- 2 milliseconds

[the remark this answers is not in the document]
AndreiMisiukevich: I removed configure await there just in case

AndreiMisiukevich:
@pictos thanks for testing.
I think that this PR is broken due to sample-platform csproj files are broken now. @VladislavAntonyuk proposed to submit a PR. Once it's merged, I believe, LongPress will work here as well

AndreiMisiukevich:
I fixed iOS locally. You can check iOS for now

AndreiMisiukevich:
@pictos I found the reason of the memory leak. It was my bad. Fixed)

VladislavAntonyuk:
> @pictos thanks for testing.
> I think that this PR is broken due to sample-platform csproj files are broken now. @VladislavAntonyuk proposed to submit a PR. Once it's merged, I believe, LongPress will work here as well

#613

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/Enums/TouchStatus.shared.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Xamarin.CommunityToolkit.Effects
+{
+	public enum TouchStatus
+	{
+		Started,
+		Completed,
```
jsuarezruiz, line 6: Would a `Running` value (or similar) make sense to notify when touch started but is still running and not completed or canceled?

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/PlatformTouchEffect.uwp.cs
```diff
@@ -0,0 +1,205 @@
...
+			if (isPressed)
+			{
+				effect?.HandleTouch(TouchStatus.Canceled);
+				AnimateTilt(pointerUpStoryboard);
+			}
+
+			effect?.HandleHover(HoverStatus.Exited);
+		}
+
+		private void OnPointerCanceled(object sender, PointerRoutedEventArgs e)
```
jsuarezruiz, line 122: The access modifier is not necessary here.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/GestureManager.shared.cs
```diff
@@ -0,0 +1,565 @@
...
+				color = GetBackgroundColor(hoveredBackgroundColor);
+
+			if (duration <= 0)
+			{
+				control.BackgroundColor = color;
+				return Task.FromResult(true);
+			}
+
+			var animationCompletionSource = new TaskCompletionSource<bool>();
+			new Animation
```
jsuarezruiz, line 323: This is very nice, but, what if we create some extensions to animate certain properties like colors, etc? (Example: https://github.com/jsuarezruiz/Xamanimation/blob/master/Xamanimation/Extensions/ColorExtensions.cs)
It will be a little bit more effort but, we can have extensions to animate certain properties in a simple way.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/TouchEffect.shared.cs
```diff
@@ -0,0 +1,1145 @@
...
+		public static readonly BindableProperty NormalAnimationEasingProperty = BindableProperty.CreateAttached(
+			nameof(NormalAnimationEasing),
+			typeof(Easing),
+			typeof(TouchEffect),
+			null,
+			propertyChanged: TryGenerateEffect);
+
+		public static readonly BindableProperty HoveredAnimationDurationProperty = BindableProperty.CreateAttached(
+			nameof(HoveredAnimationDuration),
+			typeof(int),
```
jsuarezruiz, line 343: Using a `Timespan` together with a TypeConverter perhaps opens up more options.

jsuarezruiz (commented):
Thanks very much for the contribution. It is quite big! There are some enums that I wonder if we can group, etc. and I have to keep testing on more platforms.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/Enums/TouchStatus.shared.cs
AndreiMisiukevich: There are two separate properties (Status and State)

State: Normal or Pressed (I guess, the Pressed state is what you need)
Status: Started, Completed & Canceled (Started means the touch is **running** now)

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/TouchEffect.shared.cs
AndreiMisiukevich: Decided to leave int due to attached properties (discussed over Discord)

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/GestureManager.shared.cs
AndreiMisiukevich: Done

AndreiMisiukevich:
@jsuarezruiz regarding enums
> There are some enums that I wonder if we can group, etc

I am not sure we should group them for preserving clear names for every enum (Normal + Pressed for touch and Normal + Hovered for hover etc.)

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/PlatformTouchEffect.uwp.cs
```diff
@@ -0,0 +1,205 @@
...
+
+		Storyboard pointerUpStoryboard;
+
+		protected override void OnAttached()
+		{
+			effect = TouchEffect.PickFrom(Element);
+			if (effect?.IsDisabled ?? true)
+				return;
+
+			effect.Element = Element as VisualElement;
```
pictos, line 33: @AndreiMisiukevich Would not be better to throw an exception if the `Element` isn't a `VisualElement` (VE), that way is very clear to the user that the type should be a VE. Just like we use to cast the `BindableObject` on `BindableProperties`
```suggestion
			effect.Element = (VisualElement)Element;
```
This also applies to other platforms.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/PlatformTouchEffect.uwp.cs
```diff
@@ -0,0 +1,205 @@
...
+			if (effect.NativeAnimation)
+			{
+				var nativeControl = Container;
+				if (string.IsNullOrEmpty(nativeControl.Name))
+					nativeControl.Name = Guid.NewGuid().ToString();
+
+				if (nativeControl.Resources.ContainsKey("PointerDownAnimation"))
```
pictos, line 40: how about creating a `const string`?

pictos (approved):
@AndreiMisiukevich amazing work! I also tested with the collection and works very well! My comments don't invalidate this PR.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/PlatformTouchEffect.uwp.cs
AndreiMisiukevich: Done

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/PlatformTouchEffect.uwp.cs
AndreiMisiukevich: Done

pictos (approved):

jsuarezruiz (approved):

sthewissen (approved):

SagarPanwala:
@AndreiMisiukevich I'm not criticizing, I wanted this from long time, but is this good in terms of performance as well?

AndreiMisiukevich:
> @AndreiMisiukevich I'm not criticizing, I wanted this from long time, but is this good in terms of performance as well?

I think it's not bad (I have never received any complaints) :)

sethom:
Love this project. Great work! Quick question.. I'm using an old version of TouchView (4.0.21). I'm using the Completed event to detect when the touch stops. Is there an event that fires now when the view is first touched or some way for me to do that?

AndreiMisiukevich:
Hello, yes it exists) You still can add TouchEffect as a regular effect and use its Completed event.

sethom:
Andrei.. The completed event fires at the end of the touch. How do I detected when touch starts? There's not "Started" or "Pressed" counterpart to Completed event that I am able to find.

AndreiMisiukevich:
> Andrei.. The completed event fires at the end of the touch. How do I detected when touch starts? There's not "Started" or "Pressed" counterpart to Completed event that I am able to find.

You can use StateChanged event e.g.

sethom:
Thanks..

haavamoa:
Really nice work! I have a question: When you add a touch effect on a item of a list, the item's touch effect reacts when I scroll the list. This feels unnatural as I didn't really mean to "tap" it, but I tapped it to be able to scroll. Is there a way to turn this off somehow? 🤔

AndreiMisiukevich:
@haavamoa I can imagine only one possible solution. Add a new property that will manage delay before animation start...

haavamoa:
> @haavamoa I can imagine only one possible solution. Add a new property that will manage delay before animation start...

Hm, yeah. But won't that also delay the tap when I want to tap it and not scroll? 🤔

AndreiMisiukevich:
> > @haavamoa I can imagine only one possible solution. Add a new property that will manage delay before animation start...
> 
> Hm, yeah. But won't that also delay the tap when I want to tap it and not scroll? 🤔

reasonable)

Well, I think that's impossible (

haavamoa:
Maybe subscribe to scrolled event and detach / disable the touch effect for the items only when it's scrolled would work. 🤔

AlleSchonWeg:
Hi @haavamoa , @AndreiMisiukevich 
is it possible to solve the touch on scrolling issue with a GestureDetector (https://stackoverflow.com/questions/24202671/scroll-listview-in-ontouch-method) ?
So, if scrolling is detected then cancel the touch effect. 🤔

AndreiMisiukevich:
Hey @AlleSchonWeg 
Set ```xct:TouchEffect.DisallowTouchThreshold="1"``` or maybe 5

AlleSchonWeg:
Hi @AndreiMisiukevich,

tested the setting, but nothing changed:
```XAML
<BindableLayout.ItemTemplate>
								<DataTemplate>
									<frames:myFrame xct:TouchEffect.DisallowTouchThreshold="1" xct:TouchEffect.NativeAnimation="True">
										<frames:myFrame.GestureRecognizers>
											<TapGestureRecognizer x:DataType="viewmodels:model"
																  Tapped="abcde"
																  CommandParameter="{Binding .}" />
									
										</frames:myFrame.GestureRecognizers>
									</frames:myFrame>
								</DataTemplate>
							</BindableLayout.ItemTemplate>
```

AndreiMisiukevich:
@AlleSchonWeg hm, I guess you should fill a ticket.
Do  you have an idea how to fix that?

AlleSchonWeg:
@AndreiMisiukevich 
For testing i created a 
`gestureDetector = new GestureDetector(new YourGestureDetector());` like the StackOverflow sample above.
in the OnAttached method and then in `void OnTouch(object? sender, AView.TouchEventArgs e)` i tested the event:
```C#
			if (gestureDetector.OnTouchEvent(e.Event) && e.Event?.ActionMasked != MotionEventActions.Down)
			{
				//you have detected the gesture you were looking for
				//Do something about it
				System.Diagnostics.Debug.WriteLine("sdsadad");
				return;
				//return true; //because you want to notify that you handled the event so it won't propogare further (optional, depending on what behavior you want)   
			}

```

```C#
	internal class YourGestureDetector : GestureDetector.SimpleOnGestureListener
	{
		//always return true in OnDown
		public override bool OnDown(MotionEvent? e)
		{
			return true;
		}

		public override bool OnScroll(MotionEvent? e1, MotionEvent? e2, float distanceX, float distanceY)
		{
			return true;
		}

		public override bool OnFling(MotionEvent? e1, MotionEvent? e2, float velocityX, float velocityY)
		{
			return true;
		}
	}
```
If i use my test scrollview:
```C#
            <Grid>
                <Grid.RowDefinitions>
                    <RowDefinition Height="100" />
                </Grid.RowDefinitions>
                <Grid.ColumnDefinitions>
                    <ColumnDefinition Width="AUTO" />
                </Grid.ColumnDefinitions>
                <ScrollView Orientation="Both"
                            Grid.Column="0"
                            Grid.Row="0">
                    <StackLayout Orientation="Horizontal"
                                 Spacing="10">
                        <Label Text="TITLE"
                               TextColor="Black"
                               xct:TouchEffect.NativeAnimation="True"
                               FontSize="Large" />
                        <Label Text="TITLE"
                               TextColor="Black"
                               xct:TouchEffect.NativeAnimation="True"
                               FontSize="Large" />
                        <Label Text="TITLE"
                               TextColor="Black"
                               xct:TouchEffect.NativeAnimation="True"
                               FontSize="Large" />
                        <Label Text="TITLE"
                               TextColor="Black"
                               xct:TouchEffect.NativeAnimation="True"
                               FontSize="Large" />
                        <Label Text="TITLE"
                               TextColor="Black"
                               xct:TouchEffect.NativeAnimation="True"
                               FontSize="Large" />
                        <Label Text="TITLE"
                               TextColor="Black"
                               xct:TouchEffect.NativeAnimation="True"
                               FontSize="Large" />
                                         <Label Text="TITLE"
                               TextColor="Black"
                               xct:TouchEffect.NativeAnimation="True"
                               FontSize="Large" />
                                         <Label Text="TITLE"
                               TextColor="Black"
                               xct:TouchEffect.NativeAnimation="True"
                               FontSize="Large" />
                                         <Label Text="TITLE"
                               TextColor="Black"
                               xct:TouchEffect.NativeAnimation="True"
                               FontSize="Large" />
                    </StackLayout>
                </ScrollView>
            </Grid>
```
The scroll gesture is detected. But i have no plan, how your effect works. Perhaps you could take my sample and modify it for your effect?

AndreiMisiukevich:
Can you raise a separate ticket with the described problem @AlleSchonWeg ?

AlleSchonWeg:
@AndreiMisiukevich 
Done: https://github.com/xamarin/XamarinCommunityToolkit/issues/1261

PR #616, opened 2020-11-26
apply flash change only when it is available

### LINKED ISSUE #614: [Bug] Setting flash and taking photo crashes some ipads
### Description

Not all iPads have flash capability. Currently CameraView crashes when I try to set flash on it and iPad does not have it in hardware. CameraView should check for FlashAvailable property before trying to set it ( https://developer.apple.com/documentation/avfoundation/avcapturedevice/1624627-flashavailable ). 

### Steps to Reproduce

1. Take iPad that does not have flash (f.e. iPad mini 5)
2. Launch Sample app from this repo
3. Disable Video switch
4. Tap "Snap Picture"
5. Observe long stacktrace on your computer

### Expected Behavior

Setting flash on device without hardware support for it should do nothing or return false, etc. It should not crash. Preferably CameraView could expose flash availability so the app could decide to show flash controls or not.

### Actual Behavior
Setting flash on device without hardware support for it crashes the app doing so.

### Basic Information

- Version with issue: 1.0.0-pre5
- Last known good version: none known
- IDE: VS for mac
- Platform Target Frameworks: 
  - iOS:  14.0
- Affected Devices: iPad mini 5 (and probably other cheap iPads without flash, works on pro)

### Workaround
one have to check for flash hardware before trying to set the flash

nostah:
### Description of Change ###

Do not try to set flash on iOS when it is not available in hardware.
There are no new tests for this change as there seem to be no straightforward way to mock the AVCaptureDevice.
Sample that exhibits this behaviour is already in repo.

### Bugs Fixed ###

- Fixes #614 

### API Changes ###

none

### PR Checklist ###

- [ x Changes adhere to coding standard


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/xamarin/XamarinCommunityToolkit?pullRequest=616) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/xamarin/XamarinCommunityToolkit?pullRequest=616) before we can accept your contribution.<br/><br/>:x: nostah [sign now](https://cla.dotnetfoundation.org/xamarin/XamarinCommunityToolkit?pullRequest=616)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/xamarin/XamarinCommunityToolkit?pullRequest=616) it.</sub>

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs
```diff
@@ -305,7 +305,7 @@ public void FinishedRecording(AVCaptureFileOutput captureOutput, NSUrl outputFil
 
 		public void SwitchFlash(CameraFlashMode newFlashMode)
 		{
-			if (isAvailable && device != null && newFlashMode != flashMode)
+			if (isAvailable && device != null && newFlashMode != flashMode && device.FlashAvailable)
```
AndreiMisiukevich, line 308: Shouldn't we move this check to the private ```void SwitchFlash()``` method?

I assume a crash appears inside that method and that method is invoking from different places, not only here.

AndreiMisiukevich: Actually, I do not understand why it crashes. There are checks which should prevent crashes + try-catch block
```if (device.IsFlashModeSupported(AVCaptureFlashMode.Off))```

Do you have a stack-trace?

Comment (nostah) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs:308]: then the check should be in both places as this method first changes local variable and then goes to private method

Comment (nostah) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs:308]: there are checks - yes, but the crash is during taking photo in other method. this happens becase - like I wrote above - it changes first local var and then checks. `TakePhoto()` uses local variable value and crashes because it has things set that can be illegal for some devices.

Comment (AndreiMisiukevich) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs:308]: I don't think it's a problem if we change the field's value (cuz it won't be used)

But my question is still relevant... Why does it even crash since checks exist?

Comment (AndreiMisiukevich) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs:308]: Well, then the code should be refactored a little more.
Cuz now you check only ```device.FlashAvailable``` 
But sometimes we should check ```device.TorchAvailable```. It depends on what value newFlashMode has.

Comment (nostah) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs:308]: can one have torch without flash? interesting. let me check what I can do.

Comment (AndreiMisiukevich) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs:308]: > can one have torch without flash? interesting. let me check what I can do.
Not sure, but let's imaging another case (similar to yours) but when anyone trying to enable torch when it is not supported (flash is supported)
Your code will allow changing field's value, but then it will lead to a crash in TouchPhotoAsync (because Torch is not suppported).

Comment (nostah) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs]: yes, that's true. I've added check for torch as well.

File: src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs [hunk 2]
```diff
@@ -305,7 +305,13 @@ public void FinishedRecording(AVCaptureFileOutput captureOutput, NSUrl outputFil
...
 		public void SwitchFlash(CameraFlashMode newFlashMode)
 		{
-			if (isAvailable && device != null && newFlashMode != flashMode)
+			if (!isAvailable || device == null || newFlashMode == flashMode)
+			{
+				return;
+			}
+
+			if ((flashMode == CameraFlashMode.Torch && device.TorchAvailable)
+				|| (flashMode != CameraFlashMode.Torch && device.FlashAvailable))
```
AndreiMisiukevich, line 314:
```suggestion
			var isSupported = newFlashMode == CameraFlashMode.Torch
				? device.TorchAvailable
				: device.FlashAvailable;

			if (isSupported && isAvailable && device != null && newFlashMode != flashMode)
```

AndreiMisiukevich: I guess we should compare **newFlashMode**
Please consider my proposed solution

pictos:
I'll wait until the @AndreiMisiukevich requests are addressed to review this one.

AndreiMisiukevich (approved):

AndreiMisiukevich:
> I'll wait until the @AndreiMisiukevich requests are addressed to review this one.

Done :)

jfversluis:
@nostah how does this one work exactly? Did you test it? Because when I now run the sample app and go to the CameraView it crashes immediately because `device` is null at that point when initializing.

AndreiMisiukevich:
> @nostah how does this one work exactly? Did you test it? Because when I now run the sample app and go to the CameraView it crashes immediately because `device` is null at that point when initializing.

My bad.
Seems I forgot to pull code, that's why I didn't get an exception.

I will fix it

AndreiMisiukevich:
> ~@nostah how does this one work exactly? Did you test it? Because when I now run the sample app and go to the CameraView it crashes immediately because `device` is null at that point when initializing.~
> 
> Aah actually this might be because I'm on the Simulator :D

I am preparing the fix for simulators

PR #465, opened 2020-10-19
AvatarView Fix image alignment

VladislavAntonyuk:
### Description of Change ###

Fix Absolute layout image alignment for AspectFit

### Issue
Fixes: #466 

### Behavioral Changes ###

Before:
image

After:
image

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [ ] Has samples (if omitted, state reason in description)
- [x] Rebased on top of main at time of PR
- [x] Changes adhere to coding standard

- [ ] Updated [documentation](https://github.com/MicrosoftDocs/xamarin-communitytoolkit)

PR #620, opened 2020-11-27
[Bug] AvatarView Aspect

### LINKED ISSUE #599: [Bug] AvatarView "bleeding" from Color set on background Frame
### Description

Sometimes the Color from the background frame is visible on the edges of the image on AvatarView - I think its due to some alignment issues, going on when AvatarView is put inside more "complex" layout pages.

### Steps to Reproduce

1. Put AvatarView inside a StackLayout with a WidthRequest twice the size of the AvatarView Size property
2. You will now see that the background Frame gets the size of the StackLayout, but the Image is still "square"

### Expected Behavior

The background Frame should keep the same size as the Image (and the Size property)

### Actual Behavior

In some scenaries the background Frame picks up the size of the parent control.

### Basic Information

- Version with issue: 1.0.0-pre5
- IDE: Visual Studio Mac 8.8
- Platform Target Frameworks: 
  - iOS:  14.2
  - Android: Not tested on Android
  - UWP:  Not tested on UWP

### Reproduction imagery

left image is a standard Image control - Image to the right is AvatarView - see the yellow "bleeding" :

image

Change Color property on AvatarView to Red - makes it even more visible :

image

This is from my app where I have the bleeding issue - AvatarView is places inside Grid column same size as AvatarView :

image

If I make my Grid colums horizontally larger the AvatarView Frame resizes - but not the AvatarView image :

image

Tried doing the same on the XCT samples page :

image

AndreiMisiukevich:
### Description of Change ###
The bug was introduced during fixing another bug in https://github.com/xamarin/XamarinCommunityToolkit/pull/465
We should force Image to fill all space in case Aspect has value AspectFill or Fill.

So, we preserve existing behavior for AspectFit but return back to initial behavior for Fill options.

image

### Bugs Fixed ###
- Fixes #599 

### API Changes ###
None

### Behavioral Changes ###
The image will be spread over the all space that is taken by its Frame.

### PR Checklist ###

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/AvatarView/AvatarView.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/AvatarView/AvatarView.shared.cs
@@ -128,7 +128,7 @@ namespace Xamarin.CommunityToolkit.UI.Views
 		{
 			IsClippedToBounds = true;
 			MainLayout.Children.Add(Label, new Rectangle(0.5, 0.5, -1, -1), AbsoluteLayoutFlags.PositionProportional);
-			MainLayout.Children.Add(Image, new Rectangle(0.5, 0.5, -1, -1), AbsoluteLayoutFlags.PositionProportional);
+			MainLayout.Children.Add(Image, new Rectangle(0, 0, 1, 1), AbsoluteLayoutFlags.All);
 			control.IsClippedToBounds = true;
 			control.HasShadow = false;
 			control.Padding = 0;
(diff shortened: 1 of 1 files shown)


jfversluis (approved):

tonyfonager:
I just tested this, the images still suffer from "bleeding" at the edges :

image

AndreiMisiukevich:
> I just tested this, the images still suffer from "bleeding" at the edges :
> 
> image

can you build a sample? I cannot reproduce

AndreiMisiukevich:
That's possible if your images have transparent areas
I tried on iOS and cannot reproduce

PR #628, opened 2020-12-01
VariableMultiConverter + tests

### LINKED ISSUE #618: [Enhancement] Various MultiBinding converters
## Summary
A recent version of Xamarin.Forms introduces the `MultiBinding`.

https://docs.microsoft.com/en-us/xamarin/xamarin-forms/app-fundamentals/data-binding/multibinding

This allows for a single binding to take in multiple booleans and decide things based on that. However, it also takes in a converter that indicates whether or not all booleans must be true, any boolean can be true, etc. The docs above even include a sample converter. This looks like an ideal candidate to put a series of these converters in the toolkit.

## API Changes
- [x] Add VariableMultiConverter

## Intended Use Case
```xaml
             xmlns:local="clr-namespace:DataBindingDemos"
             x:Class="DataBindingDemos.MultiBindingConverterPage"
             Title="MultiBinding Converter demo">

    <ContentPage.Resources>
        <local:VariableMultiConverter x:Key="AllTrueConverter" ConditionType="All" />
        <local:VariableMultiConverter x:Key="AnyTrueConverter" ConditionType="Any" />
        <local:VariableMultiConverter x:Key="TwoTrueConverter" ConditionType="Count" Count="2" />
        <local:InverterConverter x:Key="InverterConverter" />
    </ContentPage.Resources>

    <CheckBox>
        <CheckBox.IsChecked>
            <MultiBinding Converter="{StaticResource AllTrueConverter}">
                <Binding Path="Employee.IsOver16" />
                <Binding Path="Employee.HasPassedTest" />
                <Binding Path="Employee.IsSuspended"
                         Converter="{StaticResource InverterConverter}" />
            </MultiBinding>
        </CheckBox.IsChecked>
    </CheckBox>
</ContentPage>  
```

## Who Will Do The Work?

- [X] I am willing to take this on myself

sthewissen:
### Description of Change ###
Adds the VariableMultiConverter as described in the enhancement issue below.

### Bugs Fixed ###
- Fixes #618

### API Changes ###
Added: 
 
- `VariableMultiConverter` and...

```csharp
public enum MultiBindingCondition
{
	All = 0,
	Any = 1,
	Count = 2
}
```
### PR Checklist ###

- [X] Has tests (if omitted, state reason in description)


sthewissen:
Feel free to already take a look / review. I'll be adding a sample next and some docs (in the other repo).

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/MultiBindingCondition.shared.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Xamarin.CommunityToolkit.Helpers
+{
+	public enum MultiBindingCondition
+	{
+		All = 0,
+		Any = 1,
+		Count = 2
+	}
```
AndreiMisiukevich, line 8:
```csharp
	[Flags]
	public enum MultiBindingCondition
	{
		None = 1,
		Any = 2,
		All = 4,
		Count = 8
	}
```

In this way, I can get "True" if **None** of converters returned **True** (All false)

Or, for example, **None** or **All** True, otherwise False (Flags combination)

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/VariableMultiConverter.shared.cs
```diff
@@ -0,0 +1,58 @@
...
+
+		public object[] ConvertBack(object value, Type[] targetTypes, object parameter, CultureInfo culture)
+		{
+			if (!(value is bool b) || targetTypes.Any(t => !t.IsAssignableFrom(typeof(bool))))
+				return null;
+
+			if (b)
+				return targetTypes.Select(t => ConditionType == MultiBindingCondition.All ? (object)true : (object)false).ToArray();
+			else
+				return null;
```
AndreiMisiukevich, line 55: Can you please convert it to a ternary conditional operator?

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/VariableMultiConverter.shared.cs
```diff
@@ -0,0 +1,58 @@
...
+				MultiBindingCondition.Any => false,
+				MultiBindingCondition.Count => Count == amount,
+				MultiBindingCondition.All => true,
+				_ => true,
+			};
+		}
+
+		public object[] ConvertBack(object value, Type[] targetTypes, object parameter, CultureInfo culture)
+		{
+			if (!(value is bool b) || targetTypes.Any(t => !t.IsAssignableFrom(typeof(bool))))
```
AndreiMisiukevich, line 49: let's avoid unclear names for non-lambda parameters (I mean, can we rename **b** to **boolValue**?)

AndreiMisiukevich (changes requested):
Great work!

I have a question:
What do you think about the approach when we avoid enum at all.

Count = -1 (Means Any)
Count = 0 (Means none)
Count = {Value} (Means Count)
Count = int.Max or actual children.Count) (Means All)

I am not sure about this approach. That's more about your thoughts. I wonder what do you think))

sthewissen:
> Great work!
> 
> I have a question:
> What do you think about the approach when we avoid enum at all.
> 
> Count = -1 (Means Any)
> Count = 0 (Means none)
> Count = {Value} (Means Count)
> Count = int.Max or actual children.Count) (Means All)
> 
> I am not sure about this approach. That's more about your thoughts. I wonder what do you think))

Thanks for the review!

Regarding your suggestion, I think it will be a bit too ambiguous for the user if we use `Count` that way. To me, using `MaxInt` or -1 to indicate these scenarios feels a bit counterintuitive and unclear. I agree that it would clean up the code further, but I'm not sure if that should take priority over having some more clarity out-of-the-box for the user.

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/VariableMultiConverter.shared.cs
```diff
@@ -0,0 +1,55 @@
...
+				_ => true,
+			};
+		}
+
+		public object[] ConvertBack(object value, Type[] targetTypes, object parameter, CultureInfo culture)
+		{
+			if (!(value is bool boolValue) || targetTypes.Any(t => !t.IsAssignableFrom(typeof(bool))))
+				return null;
+
+			return boolValue ? targetTypes.Select(t => ConditionType == MultiBindingCondition.All ? (object)true : (object)false).ToArray()	: null;
```
AndreiMisiukevich, line 52:
```csharp
targetTypes.Select(t => ConditionType == MultiBindingCondition.All ? (object)true : (object)false).ToArray()
```
Let's change to

```csharp
targetTypes.Select(t => ConditionType == MultiBindingCondition.All).OfType<object>().ToArray()
```

OR

```csharp
targetTypes.Select(t => (object)(ConditionType == MultiBindingCondition.All)).ToArray()
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/VariableMultiConverter.shared.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
+using System.Globalization;
+using System.Linq;
+using Xamarin.CommunityToolkit.Helpers;
+using Xamarin.Forms;
+using Xamarin.Forms.Internals;
+
+namespace Xamarin.CommunityToolkit.Converters
+{
+	public class VariableMultiConverter : IMultiValueConverter
```
AndreiMisiukevich, line 10: Should it implement Converter Extension somehow (perhaps we will have to create a new extension class for multy converter)

src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj
```diff
@@ -23,8 +23,20 @@
     <RepositoryUrl>https://github.com/xamarin/XamarinCommunityToolkit</RepositoryUrl>
     <PackageReleaseNotes>
       
+      
+      
+      
+      
+      
+
```
AndreiMisiukevich, line 31:
```suggestion
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj
```diff
@@ -23,8 +23,20 @@
...
+      
+      
       <!-- See if we can host these on MS docs? -->
     
+    
+    
+    
+    
+    
+
```
AndreiMisiukevich, line 39:
```suggestion
```

sthewissen:
Visual Studio sometimes does dumb things.

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/MultiBindingCondition.shared.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Xamarin.CommunityToolkit.Helpers
+{
+	public enum MultiBindingCondition
```
AndreiMisiukevich, line 3: ```[Flags]``` attribute is missed if we want to support multiple options)

sthewissen: But why would you? They all kind of mutually exclude one another right?

AndreiMisiukevich: Perhaps it's useless, but I don't think they all exclude others.

4 converters and I want to say that **valid** ALL or None.

But maybe let's skip it for now

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/MultiBindingCondition.shared.cs
```diff
@@ -0,0 +1,10 @@
...
+	{
+		None = 1,
+		All = 2,
+		Any = 4,
+		Count = 8
```
AndreiMisiukevich, line 8: Then maybe we can leave them as default (without specifying numbers) 0, 1, 2, 3

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/VariableMultiConverter.shared.cs
```diff
@@ -0,0 +1,56 @@
...
+		public MultiBindingCondition ConditionType { get; set; }
+
+		public int Count { get; set; }
+
+		public object Convert(object[] values, Type targetType, object parameter, CultureInfo culture)
+		{
+			if (values == null || !targetType.IsAssignableFrom(typeof(bool)))
+				return false;
+
+			var amount = 0;
```
AndreiMisiukevich, line 22: I propose to simplify the code a little.

```csharp

var boolValues = values.OfType<bool>();
if (boolValues.Count() != values.Count())
	return false;

var count = boolValues.Where(v => v).Count();

var amount = ConditionType switch {
	MultiBindingCondition.Any => 1,
	MultiBindingCondition.Count => Count,
	_ => boolValues.Count()
};

return amount == count;
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/VariableMultiConverter.shared.cs
```diff
@@ -0,0 +1,56 @@
...
+
+			foreach (var value in values)
```
AndreiMisiukevich, line 24: BTW, this code doesn't handle **None** value if I am not wrong.

Also, I have a question about **Count**. It means "ExactValue". Perhaps it would be handy to have the possibility to set "At least {Count}" or "Not greater than {Count}"

What do you think?

Sorry, if I am too picky :) But I like this idea very much

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/VariableMultiConverter.shared.cs
sthewissen: This won't work entirely, right? Since for `Any` it can also be 2 and be valid. `Count` will be 2, but it will expect 1. I refactored a bit to support that too.

AndreiMisiukevich: Yes, you are right. It would work for ```>=```

sthewissen:
Added a sample

9ndnbxWzav

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/VariableMultiValueConverter.shared.cs
```diff
@@ -0,0 +1,48 @@
...
+
+			if (boolValues.Count() != values.Count())
+				return false;
+
+			var count = boolValues.Where(v => v).Count();
+
+			return ConditionType switch
+			{
+				MultiBindingCondition.Any => count >= 1,
+				MultiBindingCondition.None => count == 0,
```
AndreiMisiukevich, line 32: Any: GreaterThan **Count=0**
None: Exact **Count=0**

:)

Do you think we should preserve them for convenient usage?

AndreiMisiukevich: Probably yes, we should. I am just pointing to this case 👍

sthewissen: I agree that usage goes above a few fewer lines of code in this case :)

AndreiMisiukevich: I guess Count has 0 as the default value :) So, it should be an equal amount of lines

sthewissen: Ok ok, same amount of lines :)

sthewissen:
Docs also PRed in. LGTM :)

AndreiMisiukevich (approved):
LGTM

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/VariableMultiValueConverter.shared.cs
```diff
@@ -0,0 +1,48 @@
...
+		{
+			if (values == null)
+				return false;
+
+			var boolValues = values.OfType<bool>();
+
+			if (boolValues.Count() != values.Count())
+				return false;
+
+			var count = boolValues.Where(v => v).Count();
```
pictos, line 27: Do we need this Where? If don't we can calculate the `Count` for `boolValues` just one time.

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/VariableMultiValueConverter.shared.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+	public class VariableMultiValueConverter : MultiValueConverterExtension, IMultiValueConverter
+	{
+		public MultiBindingCondition ConditionType { get; set; }
+
+		public int Count { get; set; }
+
+		public object Convert(object[] values, Type targetType, object parameter, CultureInfo culture)
+		{
+			if (values == null)
```
pictos, line 19:
```suggestion
			if (values == null || values.Length== 0)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/VariableMultiValueConverter.shared.cs
```diff
@@ -0,0 +1,48 @@
...
+				return false;
+
+			var boolValues = values.OfType<bool>();
+
+			if (boolValues.Count() != values.Count())
```
pictos, line 24:
```suggestion
			var boolValues = values.OfType<bool>().ToArray();

			if (boolValues.Length != values.Length)
```

pictos (changes requested):
What do you think about these suggestions?

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/VariableMultiValueConverter.shared.cs
AndreiMisiukevich: @pictos do you mean ```var count = boolValues.Count(v => v);``` ?

pictos: Yes

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/VariableMultiValueConverter.shared.cs
sthewissen: This one creates kind of a situation. Let's say you have a variable list of values, and you want `None` of the values to be true. However, it could also be that the list has no values. Then `None` would theoretically be true as there are no booleans in there that are true. I wouldn't personally add length-based escapes and allow for the list to also be empty and continue functioning as it does.

pictos: I didn't see this use case. I think it's ok to keep the current implementation.

AndreiMisiukevich (approved):

pictos (approved):
Really nice (:

PR #629, opened 2020-12-01
CameraView Rename CameraCaptureOptions and remove SavePhotoToFile

jfversluis:
### Description of Change ###

This change removes the `SavePhotoToFile` option for now. It isn't implemented on iOS and the user has no control over where things are saved on Android and UWP. This seems not very useful to me.

Also renamed `CameraCaptureOptions` to `CameraCaptureMode` which seems more fitting to me.

### Bugs Fixed ###
N/A

### API Changes ###

Changed:

 - `enum CameraCaptureOptions` => `enum CameraCaptureMode`
 - `CaptureOptions` => `CaptureMode`
 
Removed:

- `bool SavePhotoToFile`

### Behavioral Changes ###

Breaking change for users already using CameraView and implementing these properties

### PR Checklist ###


AndreiMisiukevich:
@jfversluis ```Error CS1061: 'CameraView' does not contain a definition for 'CaptureOptions'```
Can you check pls?

jfversluis:
Do not merge! Checking what's going on with CameraView (on iOS) and also if it has to do with these changes or anything

jfversluis:
Hm interesting enough this seems to also influence if the image is saved to the iOS gallery or not

jfversluis:
OK think we should make it more like the MediaPlugin with the options object: https://github.com/jamesmontemagno/MediaPlugin#saving-photovideo-to-camera-rollgallery

Not sure if I/we can make that happen for v1, in which case I propose to still do the rename of `CameraCaptureOptions` and remove this `SavePhotoToFile` property and implement it this way later so we don't have to introduce any breaking changes by then. Thoughts?

jfversluis:
Created a new issue for the `SavePhotoToFile` property. With that I think, if you agree with this idea, this PR is ready to go.

AndreiMisiukevich (dismissed):

pictos (changes requested):
I think that we should remove the `Path` property, from the `MediaCapturedEventArgs`. If we keep it there, the user will expect a value.

jfversluis:
Great suggestion @pictos !

Done for now

pictos (approved):

yoyokits:
The documentation is still using SavePhotoToFile.
[CameraView Documentation](https://docs.microsoft.com/en-us/xamarin/community-toolkit/views/cameraview)

PR #639, opened 2020-12-02
Update BadgeIndicatorContainer visibility when BadgeView is created

### LINKED ISSUE #638: [Bug] BadgeView BadgeIndicatorContainer is visible if the Text property is 0 at the start
### Description
When showing a page containing a  `BadgeView ` with it Text property is initially equals to 0, the badge is visible with "0" as text while it shouldn't be.

### Steps to Reproduce

1. In the sample project change [`Counter` initial value](https://github.com/xamarin/XamarinCommunityToolkit/blob/1860937eacde3ced52e32fffff08a7592e8754b1/src/CommunityToolkit/Xamarin.CommunityToolkit.Sample/ViewModels/Views/BadgeViewViewModel.cs#L10) to 0.

### Expected Behavior
The BadgeIndicatorContainer should not be visible.
### Actual Behavior
The BadgeIndicatorContainer  is visible with value 0.
### Basic Information

- Version with issue:
- Last known good version: NA
- IDE: VS2019
- Platform Target Frameworks: ALL
- Android Support Library Version: NA
- Nuget Packages: NA
- Affected Devices: NA

### LINKED ISSUE #640: [Bug] BadgeView doesn't display when Text is empty
### Description
BadgeView doesn't display when initial Text value is null or empty.

### Steps to Reproduce

1. Have a badgeview with Content.
2. Set its Text to string.Empty or null
3. Watch the badgeview and its content be IsVisible = false

### Expected Behavior
The content is visible, the badge is not as there is no value.

### Actual Behavior
There is no content

### Basic Information

- Version with issue: Commit 4f6f80e9
- Last known good version: never
- IDE: VS 19
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Nuget Packages:
- Affected Devices:

### Reproduction imagery

Not overly helpful as the visual is to be invisible.

### Reproduction Link

Add to BadgeSamplePage
```
                <xct:BadgeView
                    BackgroundColor="Red"
                    Text=""
                    TextColor="White">
                    <Label Text="Initial Empty Value" />
                </xct:BadgeView>
```

Cfun1:
### Description of Change ###
Initially set `IsInvisible="true"` for `BadgeIndicatorContainer` so that it will be invisible when `BadgeView` is created and it `Text` property is initially equals to 0.

### Bugs Fixed ###
- Fixes #640

Screenshot_1607114251

### Behavioral Changes ###
The `BadgeIndicatorContainer` will be invisible if when `BadgeView` is created it `Text` property is initially equals to 0.

-  For some reasons the badge position is changed after the modifications in this PR from top right to top left.

### PR Checklist ###


src/CommunityToolkit/Xamarin.CommunityToolkit/Views/BadgeView/BadgeView.shared.cs
```diff
@@ -184,7 +184,8 @@ static Grid CreateIndicatorContainerElement()
 		   => new Grid
 		   {
 			   HorizontalOptions = LayoutOptions.Start,
-			   VerticalOptions = LayoutOptions.Start
+			   VerticalOptions = LayoutOptions.Start,
+   			   IsVisible = false
```
AndreiMisiukevich, line 188:
```suggestion
			   IsVisible = false
```

AndreiMisiukevich (commented):
LGTM

Cfun1:
@AndreiMisiukevich bindable property name  has the same name as the enum Type should it be renamed? https://github.com/xamarin/XamarinCommunityToolkit/blob/1860937eacde3ced52e32fffff08a7592e8754b1/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/BadgeView/BadgeView.shared.cs#L29
 but this is not what causing the position to change.

AndreiMisiukevich:
No, let's leave as is

I think we need to call ```UpdateBadgeViewPlacement()``` in OnTextChanged before UpdateLayout
It should solve our problem

Cfun1:
Exactly what I was about to test right now :)

AndreiMisiukevich:
Oh, probably it won't help(

Seem we should observe ```BadgeIndicatorContainer.IsVisible``` and when this property changes we will have to call ```UpdateBadgeViewPlacement(true)``

Cfun1:
I tested you are right It becomes visible before changing the position

Cfun1:
> 
> 
> Oh, probably it won't help(
> 
> Seem we should observe `BadgeIndicatorContainer.IsVisible` and when this property changes we will have to call `UpdateBadgeViewPlacement(true)`

I tried with (also with `PropertyChanged `)
```
		static Grid CreateIndicatorContainerElement()
		{
			var newgrid = new Grid
			{
				HorizontalOptions = LayoutOptions.Start,
				VerticalOptions = LayoutOptions.Start,
				IsVisible = false
			};
			newgrid.PropertyChanging += IndicatorContainerElement_PropertyChanged;
			return newgrid;
		}

		public static void IndicatorContainerElement_PropertyChanged(object sender, PropertyChangingEventArgs e)
		{
			if (e.PropertyName.Equals(nameof(IsVisible)))
				((sender as Grid).Parent as BadgeView)?.UpdateBadgeViewPlacement(true);
		}
```

Unfortunately it is not working maybe  because `UpdateVisibilityAsync` is running async or I am missing something else.

Cfun1:
I am for now testing without animation only you may need to set  `IsAnimated="False"` in xaml.

Another suggestion is to call `UpdateBadgeViewPlacement` like:

From https://github.com/xamarin/XamarinCommunityToolkit/blob/1860937eacde3ced52e32fffff08a7592e8754b1/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/BadgeView/BadgeView.shared.cs#L351-L352

To:
```
else
{
		BadgeIndicatorContainer.IsVisible = badgeIsVisible;
		await UpdateBadgeViewPlacement(true);
 }
```

it is updating the position but for some reasons the dimensions are not good (small) when incrementing first time from 0 up to 3, after 4 it becomes good.

Cfun1:
It seems working as expected, now is ready to be tested.

Cfun1:
Ignore the last commit as it appears it creating an infinite loop when badge position is other than TopRight.

Cfun1:
I have changed below line for debugging

https://github.com/xamarin/XamarinCommunityToolkit/blob/1860937eacde3ced52e32fffff08a7592e8754b1/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/BadgeView/BadgeView.shared.cs#L255
To
`if ((BadgeContent.Height <= 0 && BadgeContent.Width <= 0) || (BadgeText.Height <= 0 && BadgeText.Width <= 0))`

and found that `BadgeText.Height` and `BadgeText.Width` equals -1 while it is visible and at the same time `BadgeContent.Width` has a correct value. I don't know why `BadgeText` are not updated on the on the first increment of text value from 0 to 1.

With the modification mentioned in this comment and by ignoring the last commit, the badge position is updated but only after 2 textchanged events (means from value 0 to 2)

Cfun1:
I have changed below line for debugging

https://github.com/xamarin/XamarinCommunityToolkit/blob/1860937eacde3ced52e32fffff08a7592e8754b1/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/BadgeView/BadgeView.shared.cs#L255
To
`if ((BadgeContent.Height <= 0 && BadgeContent.Width <= 0) || (BadgeText.Height <= 0 && BadgeText.Width <= 0))`

and found that `BadgeText.Height` and `BadgeText.Width` equals -1 while it is visible and at the same time `BadgeContent.Width` has a correct value. I don't know why `BadgeText` are not updated on the on the first increment of text value from 0 to 1.

Status now: The badge position is updated but only after 2 TextChanged events (means from value 0 to 2)

AndreiMisiukevich:
Hmm, interesting case. I'll take a look and try to fix myself (I am not sure I will have success haha)

AndreiMisiukevich:
I guess this badge should stick to the top right corner
image

Now it's a little above

AndreiMisiukevich:
@Cfun1 https://github.com/xamarin/XamarinCommunityToolkit/pull/654

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/BadgeView/BadgeView.shared.cs
```diff
@@ -230,7 +231,7 @@ void UpdateLayout()
 			BadgeIndicatorBackground.BorderColor = BorderColor;
 			BadgeIndicatorBackground.HasShadow = HasShadow;
 
-			BadgeText.Text = Text;
+			BadgeText.Text = string.IsNullOrEmpty(Text) ? "0" : Text;
```
AndreiMisiukevich, line 234: I don't think it's the proper way.
I think we should just delete the check inside UpdateVisibilityAsync

Cfun1: I have tried with https://github.com/xamarin/XamarinCommunityToolkit/blob/1860937eacde3ced52e32fffff08a7592e8754b1/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/BadgeView/BadgeView.shared.cs#L325-329 commented out but it was not working.

Cfun1:
> 
> 
> I guess this badge should stick to the top right corner
> image
> 
> Now it's a little above

true, I am thinking we need to change the way how margins are calculated.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/BadgeView/BadgeView.shared.cs
```diff
@@ -268,7 +269,7 @@ void UpdateBadgeViewPlacement(bool force = false)
 			{
 				const double Padding = 6;
 				var size = Math.Max(BadgeText.Height, BadgeText.Width) + Padding;
-				BadgeIndicatorBackground.HeightRequest = size;
+				BadgeIndicatorBackground.HeightRequest = BadgeText.Height + Padding;
```
VladislavAntonyuk, line 272: do we need to add Padding twice?

AndreiMisiukevich:
#638 is fixed

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/BadgeView/BadgeView.shared.cs
```diff
@@ -272,7 +272,7 @@ void UpdateBadgeViewPlacement(bool force = false)
...
```
AndreiMisiukevich, line 275: it doesn't seem to be a proper fix cuz indicator should be stuck to the corner. Now it's above

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/BadgeView/BadgeView.shared.cs
AndreiMisiukevich: maybe let's set here Text as it was before, but maybe we will hide indicator if the Value is "0" or null or emtpy?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/BadgeView/BadgeView.shared.cs
AndreiMisiukevich: Also, currently, it's supposed to have equal width and height. So for allowing non-square / non-circle shapes, I think, we should expose a new property

AndreiMisiukevich:
Please, focus on #640 in this PR :)
Does it sound OK for you?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/BadgeView/BadgeView.shared.cs
Cfun1: I think current margins calculations are supposing to not have a big difference between Width and Height.

AndreiMisiukevich: ``var size = Math.Max(BadgeText.Height, BadgeText.Width) + Padding```
It always takes "greatest" parameter now

Cfun1:
Of course no issue :)

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/BadgeView/BadgeView.shared.cs
```diff
@@ -328,12 +328,13 @@ async Task UpdateVisibilityAsync()
...
 			{
-				IsVisible = false;
+				BadgeIndicatorBackground.IsVisible = false;
 				return;
 			}
 
 			var badgeIsVisible = !AutoHide || !badgeText.Trim().Equals("0");
-
+			BadgeIndicatorBackground.IsVisible = badgeIsVisible;
+
```
AndreiMisiukevich, line 337: Seems this line is redundant (We set badgeIsVisible below)

Cfun1: we set for `BadgeIndicatorContainer` not for `BadgeIndicatorBackground` in line 357. If this line removed together with 331 then a red frame will be visible.

jfversluis (approved):

PR #651, opened 2020-12-03
Fixed fatal error on media element

### LINKED ISSUE #558: [Bug] MediaElement crash when trying to play on iOS ( removing status observer error )
### Description

MediaElement crashes when trying to play video

### Steps to Reproduce

1. Create a page with a MediaElement tag

```
<xct:MediaElement x:Name="VideoPlayer" Source="{Binding Url}" ShowsPlaybackControls="True" IsLooping="False" MediaEnded="VideoPlayer_MediaEnded" AutoPlay="True" />
```

(In my case I have a PopupPage from Rg.Plugins.Popup, but I don't think that is the issue)

2. Open the page

```
await PopupNavigation.Instance.PushAsync(page);
```

### Expected Behavior

MediaElement opens and plays normally

### Actual Behavior

App crashes with an exception when video is about to play.

Exception thrown and app crashes:
Objective-C exception thrown.  Name: NSRangeException Reason: Cannot remove an observer <__XamarinObjectObserver 0x282bd7a00> for the key path "status" from <AVPlayerItem 0x282d4ae30> because it is not registered as an observer.

### Basic Information

- Version with issue: 1.0.0-preview1022

- Platform Target Frameworks:
  - iOS:  12.4.9
- Nuget Packages: Rg.Plugins.Popup

### Workaround

I am not aware of any workaround.

I just upgraded from using Xamarin.Forms preview MediaElement to CommunityToolkit,
while XF one had issues, I had my own MediaElementRenderer which worked,
but since I upgraded to CommunityToolkit, my own renderer can't be used because some types are now internal

I can't register on SeekRequested / StateRequested, etc, and I cannot use my own renderer,

### Reproduction Link

I may create a blank project to reproduce this issue later if the information above is not sufficient.

pictos:
### Description of Change ###

Changed the clean-up of the control to `Dispose` method.

For future references, when we have an observer like this:
`avPlayerViewController?.Player?.RemoveObserver(volumeObserver, "volume");` where we need to specify a key, we can't remove it and after Dispose it (e.g. `volumeObserver.Dispose()`) because that will lead an exception. If you just remove the observer (`RemoveObserver(volumeObserver, "volume")`) you will see in the output that you don't dipose it, and maybe this object can be alive in the memory (I guess). But if you just call the Dispose method (`volumeObserver.Dispose()`) nothing breaks and you don't see the warning in the output window.

### Bugs Fixed ###

- Fixes #558

### API Changes ###

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
@@ -358,10 +358,10 @@ namespace Xamarin.CommunityToolkit.UI.Views
 
 			if (e.OldElement != null)
 			{
-				Element.PropertyChanged -= OnElementPropertyChanged;
-				Element.SeekRequested -= MediaElementSeekRequested;
-				Element.StateRequested -= MediaElementStateRequested;
-				Element.PositionRequested -= MediaElementPositionRequested;
+				e.OldElement.PropertyChanged -= OnElementPropertyChanged;
+				e.OldElement.SeekRequested -= MediaElementSeekRequested;
+				e.OldElement.StateRequested -= MediaElementStateRequested;
+				e.OldElement.PositionRequested -= MediaElementPositionRequested;
 				SetKeepScreenOn(false);
 
 				// stop video if playing
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
@@ -82,7 +82,7 @@ namespace Xamarin.CommunityToolkit.UI.Views
 				}
 
 				var item = new AVPlayerItem(asset);
-				RemoveStatusObserver();
+				DisposeObservers(ref statusObserver);
 
 				statusObserver = (NSObject)item.AddObserver("status", NSKeyValueObservingOptions.New, ObserveStatus);
 
@@ -128,12 +128,6 @@ namespace Xamarin.CommunityToolkit.UI.Views
 			}
 			else
 				throw new ArgumentException("uri");
-		}
-
-		protected void RemoveStatusObserver()
-		{
-			statusObserver?.Dispose();
-			statusObserver = null;
 		}
 
 		protected virtual void ObserveRate(NSObservedChange e)
@@ -282,15 +276,15 @@ namespace Xamarin.CommunityToolkit.UI.Views
 			var audioSession = AVAudioSession.SharedInstance();
 			var err = audioSession.SetCategory(AVAudioSession.CategoryPlayback);
 
-			if (!(err is null))
+			if (err != null)
 				Log.Warning("MediaElement", "Failed to set AVAudioSession Category {0}", err.Code);
 
 			audioSession.SetMode(AVAudioSession.ModeMoviePlayback, out err);
-			if (!(err is null))
+			if (err != null)
 				Log.Warning("MediaElement", "Failed to set AVAudioSession Mode {0}", err.Code);
 
 			err = audioSession.SetActive(true);
-			if (!(err is null))
+			if (err != null)
 				Log.Warning("MediaElement", "Failed to set AVAudioSession Active {0}", err.Code);
 
 			if (avPlayerViewController.Player != null)
@@ -333,7 +327,7 @@ namespace Xamarin.CommunityToolkit.UI.Views
 
 					var err = AVAudioSession.SharedInstance().SetActive(false);
 
-					if (!(err is null))
+					if (err != null)
 						Log.Warning("MediaElement", "Failed to set AVAudioSession Inactive {0}", err.Code);
 
 					break;
@@ -362,26 +356,6 @@ namespace Xamarin.CommunityToolkit.UI.Views
 		{
 			base.OnElementChanged(e);
 
-			if (e.NewElement != null)
-			{
-				SetNativeControl(avPlayerViewController.View);
-
-				Element.PropertyChanged += OnElementPropertyChanged;
-				Element.SeekRequested += MediaElementSeekRequested;
-				Element.StateRequested += MediaElementStateRequested;
-				Element.PositionRequested += MediaElementPositionRequested;
-
-				avPlayerViewController.ShowsPlaybackControls = Element.ShowsPlaybackControls;
-				avPlayerViewController.VideoGravity = AspectToGravity(Element.Aspect);
-				if (Element.KeepScreenOn)
-					SetKeepScreenOn(true);
-
-				playedToEndObserver = NSNotificationCenter.DefaultCenter.AddObserver(AVPlayerItem.DidPlayToEndTimeNotification, PlayedToEnd);
-
-				UpdateBackgroundColor();
-				UpdateSource();
-			}
-
 			if (e.OldElement != null)
 			{
 				Element.PropertyChanged -= OnElementPropertyChanged;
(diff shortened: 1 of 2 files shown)


src/CommunityToolkit/Xamarin.CommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
```diff
@@ -423,5 +385,44 @@ protected override void OnElementChanged(ElementChangedEventArgs<MediaElement> e
 		}
 
 		protected virtual void UpdateBackgroundColor() => BackgroundColor = Element.BackgroundColor.ToUIColor();
+
+		protected override void Dispose(bool disposing)
```
AndreiMisiukevich, line 389: Dispose is called once (or twice)
Can't we get a scenario where OnElementChanged is called multiple times from the view (Sharing Renderer among different Elements). Probably no, at least I've never seen it :)

But in such scenario, we will get memory leak, because we subscribe on events and create IDisposable observer there

 ```csharp
				Element.PropertyChanged += OnElementPropertyChanged;
				Element.SeekRequested += MediaElementSeekRequested;
				Element.StateRequested += MediaElementStateRequested;
				Element.PositionRequested += MediaElementPositionRequested;
```

Comment (pictos) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs:389]: @AndreiMisiukevich. So we should keep the unsubscribe code inside the `e.OldElement != null` or on both? I don't if that throws an exception if we don't have more events to unsubscribe.

Comment (AndreiMisiukevich) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs:389]: Don't you think it makes sense to preserve the logic where it exists now?

Comment (pictos) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs:389]: To me make more sense to clean up resources on the Dispose method, because we can prevent the `System.ObjectDisposedException` more easily. Anyway, I'm ok to move the code to the current location

Comment (AndreiMisiukevich) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs:389]: Renderers are designed to be possible to share a single renderer among several Elements.
But I am not sure that this feature is used. (One element - one renderer.)

@jsuarezruiz @sthewissen ?

File: src/CommunityToolkit/Xamarin.CommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs [hunk 2]
```diff
@@ -364,62 +362,58 @@ protected override void OnElementChanged(ElementChangedEventArgs<MediaElement> e
...
+				if (Element.KeepScreenOn)
+					SetKeepScreenOn(true);
+
+				playedToEndObserver = NSNotificationCenter.DefaultCenter.AddObserver(AVPlayerItem.DidPlayToEndTimeNotification, PlayedToEnd);
+
+				UpdateBackgroundColor();
+				UpdateSource();
+			}
+
 			if (e.OldElement != null)
```
AndreiMisiukevich, line 385: please move Old Element logic to be above New Element one.
We should do it for making sure we are not disposing of necessary observers

AndreiMisiukevich: Btw, I propose to add a new method **Dispose**

```csharp
void Dispose (ref IDisposable disposable)
{
	disposable?.Dispose();
	disposable = null;
}
```

So we can use it everywhere as
```csharp
Dispose (ref playedToEndObserver);
Dispose (ref rateObserver);
```
etc.

pictos: Implemented

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
```diff
@@ -366,39 +358,29 @@ protected override void OnElementChanged(ElementChangedEventArgs<MediaElement> e
...
 			{
-				e.OldElement.PropertyChanged -= OnElementPropertyChanged;
-				e.OldElement.SeekRequested -= MediaElementSeekRequested;
-				e.OldElement.StateRequested -= MediaElementStateRequested;
-				e.OldElement.PositionRequested -= MediaElementPositionRequested;
+				Element.PropertyChanged -= OnElementPropertyChanged;
+				Element.SeekRequested -= MediaElementSeekRequested;
+				Element.StateRequested -= MediaElementStateRequested;
+				Element.PositionRequested -= MediaElementPositionRequested;
```
AndreiMisiukevich, line 364: I guess we still should use **e.OldElement** otherwise we will get NullReferenceException when Renderer unassigned from Element.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
```diff
@@ -423,5 +406,17 @@ protected override void OnElementChanged(ElementChangedEventArgs<MediaElement> e
...
+		{
+			disposable?.Dispose();
+			disposable = null;
+		}
+
+		protected void DisposeObservers(ref NSObject disposable)
+		{
+			disposable?.Dispose();
+			disposable = null;
+		}
```
AndreiMisiukevich, line 420: can we leave only the first method? (IDisposable) ?
**DisposeObserver** sounds better, do you agree?

pictos: Some observers are NSObject and do the same code, said that I think that's good to have the `NSObject` overload

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
pictos: Done

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
AndreiMisiukevich: Yes) But NSObject implements IDisposable, otherwise you wouldn't have the `Dispose` method)

pictos: Yes, but if you try to use the `ref IDisposable` method the compiler will complain.
image

AndreiMisiukevich: Ah, okay. Interesting)
Then maybe let's store all observers as NSObjects instead IDisposable?

pictos: It's better to have it as `IDisposable` when I need to use a `key` to assign the observer, because the `AddObserver` method returns `IDisposable` as you can [see here](https://docs.microsoft.com/en-us/dotnet/api/Foundation.NSObject.AddObserver?view=xamarin-ios-sdk-12#Foundation_NSObject_AddObserver_System_String_Foundation_NSKeyValueObservingOptions_System_Action_Foundation_NSObservedChange__) on the docs.

AndreiMisiukevich (approved):

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/MediaElement/iOS/MediaElementRenderer.ios.cs
AndreiMisiukevich: But we cast it to NSObject in code now

Why cannot we preserve them as NSObject? It worked before :)
```csharp
rateObserver = (NSObject)avPlayerViewController.Player.AddObserver("rate", NSKeyValueObservingOptions.New, ObserveRate);
```

pictos: I missed that cast, we don't need it. I don't about the history of this code, so I guess that the original author didn't know that it's ok to have this as `IDisposable`.

AndreiMisiukevich (approved):

AndreiMisiukevich (approved):

PR #674, opened 2020-12-11
LifeCycleEvents

### LINKED ISSUE #556: [Spec]  LifeCycle events for controls
# [The feature] 

For now, XF doesn't have good lifecycle events for controls, we just have the `OnAppearing` and `OnDisappearing` events/overrides on Pages and very often this isn't enough, for some scenarios. Would be good if we can have a way to know when any view (Button, ContentView, etc) is visible/loaded and invisible/unloaded.

# API

### Events

| API | Description |
| [event EventHandler Loaded] | this event will be triggered when the view is loaded on the screen. |
| [event EventHandler UnLoaded] | this event will be triggered when the view is unloaded on the screen.  |

# Scenarios

This can be used for any control/view that the dev needs to know when it's loaded and/or unloaded from the screen.

```xml
<StackLayout>
       <StackLayout.Effects>
            <effects:ViewLifecycleEffect Loaded="OnViewLoaded" Unloaded="OnViewUnloaded"/>
        </StackLayout.Effects>

        <Button.Effects>
            <effects:ViewLifecycleEffect Loaded="OnViewLoaded" Unloaded="OnViewUnloaded"/>
        </Button.Effects>
    </Button>
  <Entry />
        <Entry.Effects>
            <effects:ViewLifecycleEffect Loaded="OnViewLoaded" Unloaded="OnViewUnloaded"/>
        </Entry.Effects>
    </Entry>
</StackLayout>
```

```csharp
   void OnViewLoaded(object sender, EventArgs e)
    {
            if (sender is Button)
                 // Do button stuffs
            else if (sender is Entry)
                 // Do entry stuffs
            else if (sender is StackLayout)
                 // Do stackLayout on stuffs
    }
 
    void OnViewUnloaded(object sender, EventArgs e)
    {
            if (sender is Button)
                 // Do button stuffs
            else if (sender is Entry)
                 // Do entry stuffs
            else if (sender is StackLayout)
                 // Do stackLayout on stuffs
    }
```

# Platform Compatibility
- Target Frameworks: 
  - iOS:  
  - Android: 
  - UWP: 

# Difficulty : [low/medium/high]
Medium

## Who Will Do The Work?

pictos:
### Important

I don't have an expectation to merge this before the first stable release, it's just to make it easier for people to try this feature and provide valuable feedback.
Also, it's ready for code-review.

### Description of Change ###

This PR adds two LifeCycle events for controls across iOS, Android and UWP. With this devs can know when control or layout (more precisely a `XF.Element`) is rendered and ready for use.

This implementation was based on [this article](https://www.pshul.com/2018/03/27/xamarin-forms-add-views-lifecycle-events/) and this [XF issue](https://github.com/xamarin/Xamarin.Forms/issues/2210)

### Bugs Fixed ###

- Fixes #556 

### API Changes ###

Added: 

```csharp
	public class LifeCycleEffect : RoutingEffect
	{
		public event EventHandler Loaded;

		public event EventHandler Unloaded;
		public void RaiseLoadedEvent(Element element) => Loaded?.Invoke(element, EventArgs.Empty);

		public void RaiseUnloadedEvent(Element element) => Unloaded?.Invoke(element, EventArgs.Empty);
        }
```
- and the Androi, iOS and UWP implementation

### Behavioral Changes ###

Now the users can know when a `Xamarin.Forms.Element` is rendered and ready for use.

### PR Checklist ###


src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Effects
+{
+	public class LifeCycleEffect : RoutingEffect
+	{
+		public event EventHandler Loaded;
+
+		public event EventHandler Unloaded;
```
AndreiMisiukevich, line 10: I propose adding corresponding commands as well.

Also, would be great to make this effect attachable (similar to TouchEffect)

```xaml
<StackLayout xct:LifecycleEffect.UnloadedCommand="{Binding Command}">
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Effects
+{
+	public class LifeCycleEffect : RoutingEffect
```
AndreiMisiukevich, line 6: I prefer Life**c**ycleEffect (lowercased "c") 

https://developer.android.com/guide/components/activities/activity-lifecycle

AndreiMisiukevich:
BTW, it can be achieved via behavior (We can observe "Renderer" property change)

The first time this property changed means Renderer assigned
The second time means Renderer unassigned.

But, probably, the effect approach is even better

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffectRouter.uwp.cs
```diff
@@ -0,0 +1,45 @@
...
+
+		void OnNativeViewLoaded(object sender, RoutedEventArgs e) => lifeCycleEffect.RaiseLoadedEvent(Element);
+
+		void OnNativeViewUnloaded(object sender, RoutedEventArgs e)
+		{
+			lifeCycleEffect.RaiseUnloadedEvent(Element);
+			nativeView.Unloaded -= OnNativeViewUnloaded;
+			nativeView.Loaded -= OnNativeViewLoaded;
+			lifeCycleEffect = null;
+			nativeView = null;
```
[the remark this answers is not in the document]
pictos: Not really, because if you leave the screen the `OnDeatached` method isn't called after the `Unloaded` event, in other words, the Effect will be detached before the control is unloaded and we want to make sure that doesn't happen. And if the control is Unloaded, the Effect will be detached already.

I'm still thinking about how to make this more _beaultiful_, maybe an external method and calling on both the `OnDeatached` method and the `Unloaded` event.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
```diff
@@ -0,0 +1,31 @@
...
+				_ = new Xamarin.CommunityToolkit.iOS.Effects.LifeCycleEffectRouter();
+#elif UWP
+			if (System.DateTime.Now.Ticks < 0)
+				_ = new Xamarin.CommunityToolkit.UWP.Effects.LifeCycleEffectRouter();
+#endif
+		}
+
+		public void RaiseLoadedEvent(Element element) => Loaded?.Invoke(element, EventArgs.Empty);
+
+		public void RaiseUnloadedEvent(Element element) => Unloaded?.Invoke(element, EventArgs.Empty);
```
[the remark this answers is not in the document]
pictos: Yeah, that works as well xD

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
pictos: > Also, would be great to make this effect attachable (similar to TouchEffect)

For commands yes, I'm not sure if we can do it for events

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
pictos: As I'm always messing with names 😅

pictos:
> BTW, it can be achieved via behavior (We can observe "Renderer" property change)

Yes, we can, but this "Renderer" property is `internal` so the XF team can do break changes on it, so I want to avoid it. Also, this is more like a playground to see how it behaves and (maybe) be a reference for the Maui team.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
AndreiMisiukevich: > > Also, would be great to make this effect attachable (similar to TouchEffect)
> 
> For commands yes, I'm not sure if we can do it for events

I think we can't. But command would work

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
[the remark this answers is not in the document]
AndreiMisiukevich: Maybe let's make them internal? It will prevent issues when users call this method on their side

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffectRouter.uwp.cs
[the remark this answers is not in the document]
AndreiMisiukevich: Got it.
I'm just thinking about the case when you add effect programmatically and then remove it.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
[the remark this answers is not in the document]
pictos: Yeah, I'll commit this change by tomorrow.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
pictos: @AndreiMisiukevich I was thinking about the syntax and since we can't do attached props. for events, I think that is a bad idea to make it for Commands, I can image a syntax similar to the `GestureRecognizers`

```xml
<TapGestureRecognizer Tapped="This is an event"  Command = ""/>
```

So that way the effect will be

```xml
<xct:LifecycleEffect Loaded="Event"  LoadedCommand="Command" />
```

Thoughts?

IeuanWalker:
@pictos works well, I'm using it in a new control - https://github.com/IeuanWalker/Xamarin.Forms.CustomSwitch

pictos:
@IeuanWalker thank you so much for testing my code <3. There's something that you think can be improved?

IeuanWalker:
@pictos no prob. No all seams good. Over the next week I'll try it out in our production app, that would utilise both the loaded and unloaded events.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
AndreiMisiukevich: That's fine

src/CommunityToolkit/Xamarin.CommunityToolkit.Sample/Pages/Effects/LifeCycleEffectPage.xaml.cs
```diff
@@ -0,0 +1,50 @@
...
+	public partial class LifeCycleEffectPage
+	{
+		public LifeCycleEffectPage()
+		{
+			InitializeComponent();
+		}
+
+		void LifeCycleEffect_Loaded(object sender, EventArgs e)
+		{
+			if (sender is Button)
```
jsuarezruiz, line 22: Maybe we can include a Label in the UI that is updated every time a loaded/unloaded event is invoked (Whoever tries the sample will see everything in debug and release mode).

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
jsuarezruiz: Thinking in MVVM, I find it interesting to add commands.

ederbond:
What I miss the most when creating my XF controls is that all the XF Elements had a virtual Dispose method that I could override to dispose my own stuffs.

brminnick (commented):
Cool PR! I just have two cosmetic recommendations to tighten up the code a bit 👍

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
pictos: About the command, and MVVM. On `event` I provide the control as `sender`, that way the user can perform actions per control. If we add Command would be good (or necessary) to provide the `CommandParameter`, and it should be the control and that is against the MVVM because the VM will have a reference from the View (e.g. the control parents).

Thoughts?

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffectRouter.ios.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+	public class LifeCycleEffectRouter : PlatformEffect
+	{
+		const NSKeyValueObservingOptions observingOptions = NSKeyValueObservingOptions.Initial | NSKeyValueObservingOptions.OldNew | NSKeyValueObservingOptions.Prior;
+		LifecycleEffect lifeCycleEffect;
+		IDisposable isLoadedObserverDisposable;
+
+		protected override void OnAttached()
+		{
+			lifeCycleEffect = Element.Effects.OfType<LifecycleEffect>().FirstOrDefault();
```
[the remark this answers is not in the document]
pictos: nice catch!

src/CommunityToolkit/Xamarin.CommunityToolkit.Sample/Pages/Effects/LifeCycleEffectPage.xaml.cs
pictos: Nice suggestion!

AndreiMisiukevich:
@pictos are we ready to merge it?

pictos:
@AndreiMisiukevich not yet. I need to change the iOS implementation.

jfversluis:
Can we retarget this one to develop? Thanks! :)

jsuarezruiz:
@pictos Could you rebase to fix the current conflicts?. Thanks!.

pictos:
FYI, I should work on this next week.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffectRouter.ios.cs
```diff
@@ -0,0 +1,45 @@
...
+	public class LifeCycleEffectRouter : PlatformEffect
+	{
+		LifecycleEffect lifeCycleEffect;
+
+		protected override void OnAttached()
+		{
+			lifeCycleEffect = Element.Effects.OfType<LifecycleEffect>().FirstOrDefault() ??
+				throw new ArgumentNullException($"The effect {nameof(LifecycleEffect)} can't be null.");
+
+			Element.PropertyChanged += OnPropertyChanged;
```
pictos, line 21: On iOS, AFAIK doesn't exist a way to know the UIView lifecycle, that way I need to use the Renderer property. It still needs to be an effect, because I need to access the native renderer.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
```diff
@@ -24,8 +47,8 @@ public LifecycleEffect()
 #endif
 		}
 
-		internal void RaiseLoadedEvent(Element element) => Loaded?.Invoke(element, EventArgs.Empty);
+		internal void RaiseLoadedEvent(Element element) => eventManager.RaiseEvent(element, EventArgs.Empty, "Loaded");
```
brminnick, line 50: Let's use the `nameof` operator in lieu of a hard-coded string 👍 
```csharp
eventManager.RaiseEvent(element, EventArgs.Empty, nameof(Loaded));
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
```diff
@@ -24,8 +47,8 @@ public LifecycleEffect()
...
 
-		internal void RaiseUnloadedEvent(Element element) => Unloaded?.Invoke(element, EventArgs.Empty);
+		internal void RaiseUnloadedEvent(Element element) => eventManager.RaiseEvent(element, EventArgs.Empty, "Unloaded");
```
brminnick, line 52: Let's use the `nameof` operator in lieu of a hard-coded string 👍 
```csharp
eventManager.RaiseEvent(element, EventArgs.Empty, nameof(Unloaded));
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
```diff
@@ -0,0 +1,54 @@
...
+#if __ANDROID__
+			if (System.DateTime.Now.Ticks < 0)
+				_ = new Xamarin.CommunityToolkit.Android.Effects.LifeCycleEffectRouter();
+#elif __IOS__
+			if (System.DateTime.Now.Ticks < 0)
+				_ = new Xamarin.CommunityToolkit.iOS.Effects.LifeCycleEffectRouter();
+#elif UWP
+			if (System.DateTime.Now.Ticks < 0)
+				_ = new Xamarin.CommunityToolkit.UWP.Effects.LifeCycleEffectRouter();
+#endif
```
brminnick, line 47: Just curious - how come we're initializing an object and not assigning it to anything? Is to prevent the linker from removing the platform-specific implementation?

brminnick (changes requested):

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
brminnick: Now that I think about it, it's gotta be for the linker, because `System.DateTime.Now.Ticks < 0` will never be true, right? 😃

brminnick: Maybe we wrap this in a `#region` and add a comment about this being for the linker to prevent devs in the future from removing this block of code?

```csharp
#region Add work-around to prevent linker from removing the platform-specific implementation 

#if __ANDROID__
				_ = new Xamarin.CommunityToolkit.Android.Effects.LifeCycleEffectRouter();
#elif __IOS__
				_ = new Xamarin.CommunityToolkit.iOS.Effects.LifeCycleEffectRouter();
#elif UWP
				_ = new Xamarin.CommunityToolkit.UWP.Effects.LifeCycleEffectRouter();
#endif

#endregion

Cfun1: @brminnick  I remember I have asked the same question in the past and @AndreiMisiukevich answered me that is to trick the linker. to preserver, indeed like you thought :).

pictos: Yes @brminnick this is a trick to do a static reference to make linker don't remove the native implementation, and it will not remove just if someone uses the effect. And the compiler will do optimization on this part because this will never be called. I learned it from [here](https://github.com/xamarin/Essentials/pull/1209#discussion_r405051562).

@brminnick about the regions, do you agree to do it in another PR? That way we can do it for all effects

brminnick: > do you agree to do it in another PR? That way we can do it for all effects 

No problem!

I'll open an Issue now and assign it to myself so that I don't forget 👍

brminnick: https://github.com/xamarin/XamarinCommunityToolkit/issues/910 ✅

brminnick (approved):

AndreiMisiukevich (approved):

Cfun1:
@pictos I have opened an issue regarding missing documentation so that we won't forget about.

pictos:
@Cfun1 thanks <3

ryanharding:
Hi, sorry for my ignorance, so with this being merged, will it be in the next release? i.e. 1.0.4?

acaliaro:
> Hi, sorry for my ignorance, so with this being merged, will it be in the next release? i.e. 1.0.4?

maybe you can find it in the nightly build

Cfun1:
@ryanharding it will be in the v1.1.0 pre-release version  https://github.com/xamarin/XamarinCommunityToolkit/releases/tag/1.1.0-pre1

jfversluis:
@ryanching and all, I don't mind you asking, so don't see this as trying to be smart, I'd just like you to know for yourself when something is coming :)

Have a look to the right of this issue, there is the Milestone section which says 1.1. This means it will make it's first appearance in 1.1! And the pre-release was released just today.

Hope that clears it up a little for you

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/LifeCycle/LifeCycleEffect.shared.cs
jor1196: Could you integrate it into Shell? 
I imagine calling it this way: 
Shell.Current.CurrentPage.Loaded
Shell.Current.CurrentPage.Unloaded

PR #736, opened 2021-01-08
Add BaseCommand.ChangeCanExecute()

brminnick:
This PR is base on community feedback from @maxkoshevoi: https://github.com/xamarin/XamarinCommunityToolkit/discussions/709#discussioncomment-270001

### Description of Change ###

Because Xamarin.Forms.Command uses `ChangeCanExecute()`, adding `ChangeCanExecute()` to both `AsyncCommand` and `AsyncValueCommand` enables an easier, more seamless, transition from `Xamarin.Forms.Command` to `Xamarin.CommunityToolkit.ObjectModel.AsyncCommand`.

### API Changes ###

Adds `BaseCommand.ChangeCanExecute()`
- `BaseCommand.ChangeCanExecute()` leverages `BaseCommand.RaiseCanExecuteChanged()`

### Behavioral Changes ###

None, `BaseCommand.ChangeCanExecute()` leverages `BaseCommand.RaiseCanExecuteChanged()`:

```csharp
public void ChangeCanExecute() => RaiseCanExecuteChanged();
```

### PR Checklist ###

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/BaseCommand.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/BaseCommand.shared.cs
@@ -77,12 +77,17 @@ namespace Xamarin.CommunityToolkit.ObjectModel.Internals
 		};
 
 		/// <summary>
-		/// Raises the CanExecuteChanged event.
+		/// Raises the `ICommand.CanExecuteChanged` event.
 		/// </summary>
 		public void RaiseCanExecuteChanged()
 		{
 			// Automatically marshall to the Main Thread to adhere to the way that Xamarin.Forms automatically marshalls binding events to Main Thread
 			Device.BeginInvokeOnMainThread(() => weakEventManager.RaiseEvent(this, EventArgs.Empty, nameof(CanExecuteChanged)));
 		}
+
+		/// <summary>
+		/// Raises the `ICommand.CanExecuteChanged` event. Performs the same functionality as RaiseCanExecuteChanged leveraging the Xamarin.Forms.Command nomencalture to make the transition from Xamarin.Forms.Command easier.
+		/// </summary>
+		public void ChangeCanExecute() => RaiseCanExecuteChanged();
 	}
 }
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/BaseCommand.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/BaseCommand.shared.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.ComponentModel;
 using Xamarin.CommunityToolkit.Helpers;
 using Xamarin.Forms;
 
@@ -86,9 +87,9 @@ namespace Xamarin.CommunityToolkit.ObjectModel.Internals
 		}
 
 		/// <summary>
-		/// Raises the `ICommand.CanExecuteChanged` event.
+		/// Raises the `ICommand.CanExecuteChanged` event. Recommend using RaiseCanExecuteChanged() instead.
 		/// </summary>
-		[Obsolete("Use RaiseCanExecuteChanged() instead")]
+		[EditorBrowsable(EditorBrowsableState.Never)]
 		public void ChangeCanExecute() => RaiseCanExecuteChanged();
 	}
 }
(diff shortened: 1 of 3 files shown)


brminnick:
Assigning to @jfversluis for review since he was looped into the original discussion: https://github.com/xamarin/XamarinCommunityToolkit/discussions/709#discussioncomment-270001

src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/BaseCommand.shared.cs
```diff
@@ -77,12 +77,18 @@ protected int ExecutionCount
...
 		{
 			// Automatically marshall to the Main Thread to adhere to the way that Xamarin.Forms automatically marshalls binding events to Main Thread
 			Device.BeginInvokeOnMainThread(() => weakEventManager.RaiseEvent(this, EventArgs.Empty, nameof(CanExecuteChanged)));
 		}
+
+		/// <summary>
+		/// Raises the `ICommand.CanExecuteChanged` event.
+		/// </summary>
+		[Obsolete("Use RaiseCanExecuteChanged() instead")]
+		public void ChangeCanExecute() => RaiseCanExecuteChanged();
```
AndreiMisiukevich, line 92: Well, we added a new already obsolete method :)

I guess an obsolete attribute should be added to RaiseCanExecuteChanged, right?

AndreiMisiukevich (changes requested):

src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/BaseCommand.shared.cs
brminnick: Hey Andrei! 

I decided to tag it with `[Obsolete]` to reduce the overhead of maintaining and documenting two methods that do the same thing.

I chose to add the `[Obsolete]` tag to `ChangeCanExecute` rather than `RaiseCanExecuteChanged` for two reasons:
- The documentation, samples, and tests already include and leverage `RaiseCanExecuteChanged`
- `RaiseCanExecuteChanged` is a better description of what is actually happening under the hood; it invokes `event ICommand.CanExecuteChanged`

brminnick: Another option is to replace `[Obsolete]` with `[EditorBrowsable(EditorBrowsableState.Never)]`

maxkoshevoi: `RaiseCanExecuteChanged` is a better name, but `ChangeCanExecute` provides backwards compatibility with `Command` class (and interface). 
I would also deprecate `RaiseCanExecuteChanged`. If `ChangeCanExecute` would to be renamed (or deprecated and replaces with `RaiseCanExecuteChanged`) this needs to be done in `Command` class too to provide more seamless experience when user needs to switch from `Command` to `AsyncCommand` or backwards.

brminnick: There is no `ChangeCanExecute` in the [`ICommand` interface](https://docs.microsoft.com/dotnet/api/system.windows.input.icommand?view=net-5.0&WT.mc_id=mobile-0000-bramin). `ChangeCanExecute` is just the method name that the creators of Xamarin.Forms came up with; it is not an industry standard.

Adding `ChangeCanExecute` does allow for a seamless transition from `Xamarin.Forms.Command` to `AsyncCommand`.

maxkoshevoi: You are current about `ICommand` interface. Need to really start double checking what I'm talking about. This is really confusing method.

But this doesn't change that `Command` has it. Why having method with the same name and functionality in `AsyncCommand` wouldn't provide seamless transition?

pictos:
@brminnick since this PR makes an API Obsolete I think that is an API change, so will be better to target the `develop` branch. Thoughts?

brminnick:
@pictos It doesn't make an API Obsolete. It adds an API, improving the migration experience from Xamarin.Forms.Command to AsyncCommand.

pictos:
@brminnick my mistake so, thanks for explaining that. Yeah, I would say that we are good to go then.

pictos (approved):

PR #749, opened 2021-01-11
Fix culture not persisted to new pages

### LINKED ISSUE #745: [Bug] {xct:Translate String} works only for currently open pages
### Description
If user changes language with `LocalizationResourceManager.Current.SetCulture()`, all strings localized with `{xct:Translate String}` on **currency open** pages get updated to new language, but if user opens new page it will be localized with old language.

### Steps to Reproduce

1. Open some page (page has language A)
2. Change culture to language B with `LocalizationResourceManager.Current.SetCulture()`
3. Open new page with some string localized with `{xct:Translate String}` 
4. New page will be localized with language A instead of B

### Expected Behavior
Newly opened pages respect current language

### Actual Behavior
Newly opened pages localized using language app was opened with

### Basic Information

- Version with issue: 1.0.0
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android: 10
  - UWP:   
- Nuget Packages:
- Affected Devices:

### Workaround
Manually localize all strings from the code

maxkoshevoi:
### Description of Change ###
Turns out `Thread.CurrentThread.CurrentUICulture` is unreliable way to store culture. When new page is created `CurrentUICulture` is reset to default for some reason (even though `ManagedThreadId` is the same).

- Created private field to store culture, so we don't depend on anyone to store it (when resource designer is created by Xamarin it also creates it (see image bellow), so it seems like good approach)
image
- Also refactored `LocalizationResourceManager` a bit while I have the opportunity
  - Added second `Init(ResourceManager resource, CultureInfo initialCulture)` so culture can be set right away. 
  - ~~Added setter to `CurrentCulture` and deprecated `SetCulture()`. Culture is handled via properties everywhere (`CultureInfo.CurrentCulture`, `Thread.CurrentThread.CurrentUICulture`, `AppResources.Culture`), so it feels like `LocalizationResourceManager.CurrentCulture` should have setter too.~~

### Bugs Fixed ###

- Fixes #745 

### API Changes ###

Added: 
 
- `void Init(ResourceManager resource, CultureInfo initialCulture)`
- Setter for `CurrentCulture`

Changed:

 - `Thread.CurrentThread.CurrentUICulture` => `CultureInfo currentCulture`

### Behavioral Changes ###

- Culture now stored internally instead of using `Thread.CurrentThread.CurrentUICulture` 
- Also `indexer` used `Thread.CurrentThread.CurrentUICulture` while `GetValue` used `CultureInfo.CurrentCulture`. Now they provide the same results.

### PR Checklist ###

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs
@@ -29,13 +29,12 @@ namespace Xamarin.CommunityToolkit.Helpers
 		public string this[string text] =>
 			GetValue(text);
 
-		[Obsolete("Use " + nameof(CurrentCulture) + " to set culture")]
 		public void SetCulture(CultureInfo language) => CurrentCulture = language;
 
 		public CultureInfo CurrentCulture
 		{
 			get => currentCulture;
-			set
+			private set
 			{
 				currentCulture = value;
 				Invalidate();
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs
@@ -20,7 +20,7 @@ namespace Xamarin.CommunityToolkit.Helpers
 		public void Init(ResourceManager resource, CultureInfo initialCulture)
 		{
 			resourceManager = resource;
-			CurrentCulture = initialCulture;
+			currentCulture = initialCulture;
 		}
 
 		public string GetValue(string text) =>
@@ -29,17 +29,13 @@ namespace Xamarin.CommunityToolkit.Helpers
 		public string this[string text] =>
 			GetValue(text);
 
-		public void SetCulture(CultureInfo language) => CurrentCulture = language;
-
-		public CultureInfo CurrentCulture
+		public void SetCulture(CultureInfo language)
 		{
-			get => currentCulture;
-			private set
-			{
-				currentCulture = value;
-				Invalidate();
-			}
+			currentCulture = language;
+			Invalidate();
 		}
+
+		public CultureInfo CurrentCulture => currentCulture;
 
 		public event PropertyChangedEventHandler PropertyChanged;
 
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs
@@ -1,4 +1,5 @@
-﻿using System.ComponentModel;
+﻿using System;
+using System.ComponentModel;
 using System.Globalization;
 using System.Resources;
 using System.Threading;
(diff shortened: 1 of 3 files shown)


src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs
```diff
@@ -11,30 +12,40 @@ public class LocalizationResourceManager : INotifyPropertyChanged
...
+			CurrentCulture = initialCulture;
 		}
 
-		public string this[string text] => resourceManager.GetString(text, CurrentCulture);
+		public string GetValue(string text) =>
+			resourceManager.GetString(text, CurrentCulture);
 
-		public void SetCulture(CultureInfo language)
-		{
-			Thread.CurrentThread.CurrentUICulture = language;
```
sthewissen, line 24: Should we not still set the current thread culture somewhere? Some other elements (date formats e.g.) also depend on this I would suppose. Or do we leave that responsibility to the user and only concern ourselves with text resources retrieved through our Translate extension?

maxkoshevoi: With traditional localization approach there is `CultureInfo resourceCulture` static field inside AppResources (or wherever you named it) class and setting it does nothing except changes culture of the strings that will be returned from it.

```cs
/// <summary>
///   Overrides the current thread's CurrentUICulture property for all
///   resource lookups using this strongly typed resource class.
/// </summary>
[global::System.ComponentModel.EditorBrowsableAttribute(global::System.ComponentModel.EditorBrowsableState.Advanced)]
public static global::System.Globalization.CultureInfo Culture {
    get {
        return resourceCulture;
    }
    set {
        resourceCulture = value;
    }
}
```

I like this approach since there is no hidden side effects. If user wants to change `CultureInfo.CurrentCulture`, `Thread.CurrentThread.CurrentUICulture`, `AppResources.Culture` or anything else when `LocalizationResourceManager`'s culture is changes they are welcome to do so like this:
```cs
LocalizationResourceManager.Current.Init(LN.ResourceManager);
LocalizationResourceManager.Current.PropertyChanged += (_, _) => LN.Culture = LocalizationResourceManager.Current.CurrentCulture;
```

This is what I use in my code, so I can still call AppResource dirrectly (`LN.String`) to get localized string instead of `LocalizationResourceManager.Current["String"]` since it's shorter and doesn't contain hardcoded strings.

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs
```diff
@@ -11,30 +12,40 @@ public class LocalizationResourceManager : INotifyPropertyChanged
...
-		public CultureInfo CurrentCulture => Thread.CurrentThread.CurrentUICulture;
+		public CultureInfo CurrentCulture
+		{
+			get => currentCulture;
+			set
+			{
+				currentCulture = value;
+				Invalidate();
+			}
+		}
```
pictos, line 43: I have a question about this change... Is this change something in the usage? If no I would say to keep the prop `private set` and use the `SetCulture` method to handle it. 

I know that is good to use the C# features, but this isn't a good reason to Obsolete an API.

pictos (changes requested):
@maxkoshevoi Thanks for this fix! I have just one point that I want to understand better.

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs
maxkoshevoi: Makes sense. Just figured since package is new and this method is generally not used much, it's still time to move to using property instead. It's not only good use of C#, it's how everything that updates culture does it (`CultureInfo.CurrentCulture`, `Thread.CurrentThread.CurrentUICulture`, `AppResources.Culture`).

Are you sure it's worth sticking to this method? I really feel like property is the way to go here (especially considering that we a moving to backing field, so this method no longer has any external side effects)

pictos: @maxkoshevoi if this change isn't the fix for the bug, I would say to keep at it is (with method instead of Prop) and it's just to avoid to make the API Obsolete. Anyway, we need the input of other maintainers before. 
If we agree to Obsolete this API this will be an API change and should target the develop branch.

@AndreiMisiukevich , @sthewissen , @jfversluis , @jsuarezruiz, what do you think about it?

maxkoshevoi: @pictos This change isn't the fix for the bug. This fix is sort of a blocker since `LocalizationResourceManager` cannot be reliably used until it's merged, so we can split this PR into two. One for fixing the bug and second for deprecating the `SetCulture` method.

pictos: @maxkoshevoi  that sounds good! Before opening a PR to deprecate the method, can you open an issue first?

maxkoshevoi: @pictos Removed `[Obsolete]` attribute from `SetMethod` and made `CurrentCulture` setter private. I'll create an issue to reverse that later today,

pictos: @maxkoshevoi Thank you so much! For the help!

pictos (approved):

pictos:
@maxkoshevoi I took the liberty to do a small change on the code, please let me know your thoughts.

maxkoshevoi:
> @maxkoshevoi I took the liberty to do a small change on the code, please let me know your thoughts.

No problem. I moved all logic into the property because I wanted to deprecate the method, but since we are not doing this right now, it can stay in the method.

AndreiMisiukevich (approved):

PR #750, opened 2021-01-11
Add LocalizedString class for automatic  ViewModel string updates

maxkoshevoi:
Related to #743

### Description of Change ###

With new `LocalizationResourceManager` and `TranslateExtension` it's now easy to change app's language on the fly. One issue is that only strings localized with `{xct:Translate String}` (aka using xaml) are automatically updated on language change.
Strings that are set from the code need to be updated manually.

- Added class `LocalizedString` that automatically updates strings declared in ViewModel.
- Added `INotifyPropertyChanged.WeakSubscribe` extension method. `LocalizedString` needs it to be garbage collected when there are no references to it other then subscription to `LocalizationResourceManager.PropertyChanged`.

Example usage:
```cs
public LocalizedString AppVersion { get; } = new(() => string.Format(LN.Version, AppInfo.VersionString));
```
```xaml
<Label Text="{Binding AppVersion.Localized}"/>
```

```cs
private LocalizedString title;
public LocalizedString Title { get => title; set => SetProperty(ref title, value); }
...
Title = new(() => string.Join(", ", selectedEntities.Select(se => se.Name)));
or
Title = (Func<string>)() => string.Join(", ", selectedEntities.Select(se => se.Name));
```
```xaml
             xmlns:xct="http://xamarin.com/schemas/2020/toolkit"
             Title="{Binding Title.Localized}">
```
**Note:** I tried to do it without `.Localized` by overriding `ToString()`, but if xaml references object itself, instead of it's property, no amount of `OnPropertyChanged(null)` will force UI to refresh. [More info.](https://forums.xamarin.com/discussion/65678/how-to-get-inotifypropertychanged-for-the-whole-object)

### API Changes ###

Added: 
 
- `class LocalizedString`
- `void WeakSubscribe<T>(this INotifyPropertyChanged target, T subscriber, Action<T, object, EventArgs> action)`

### Behavioral Changes ###

No changes

### PR Checklist ###


maxkoshevoi:
@pictos Could you also take a look at this PR please?

pictos:
@maxkoshevoi sure, can you update the branch first? And I'll checkout and test here

maxkoshevoi:
@pictos Done

pictos:
@maxkoshevoi can you add your new feature on our sample app, I guess this [the class ](https://github.com/xamarin/XamarinCommunityToolkit/blob/main/samples/XCT.Sample/ViewModels/SettingViewModel.cs) that you should add the feature as a sample

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizedString.shared.cs
```diff
@@ -0,0 +1,42 @@
...
+			Generator = generator;
+			localizationManager.PropertyChanged += (sender, args) => Invalidate();
+		}
+
+		Func<string> generator;
+
+		public Func<string> Generator
+		{
+			get => generator;
+			set
```
pictos, line 25: Do we want that the Generator isn't immutable?

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizedString.shared.cs
```diff
@@ -0,0 +1,42 @@
...
+	{
+		public LocalizedString(Func<string> generator = null)
+            : this(LocalizationResourceManager.Current, generator)
+        {
+        }
+
+		public LocalizedString(LocalizationResourceManager localizationManager, Func<string> generator = null)
+		{
+			Generator = generator;
+			localizationManager.PropertyChanged += (sender, args) => Invalidate();
```
pictos, line 17: I'm a little concerned about this event. This isn't a single class and the user can create a lot of instances of this (maybe without intention) and we never unsubscribe this event what can lead to a memory leak and odd behaviors. So I see two solutions for it:

1. Implement the IDisposable pattern;
2. Use the singleton pattern (as we did for `LocalizationResourceManager`), that way we will just one instance of it for all app lifecycle.

I would say that option 2 will be simple and enough for this scenario. Please let me know your thoughts.

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizedString.shared.cs
```diff
@@ -0,0 +1,42 @@
...
+			set
+			{
+				generator = value;
+				Invalidate();
+			}
+		}
+
+		public string Localized => Generator?.Invoke();
+
+		public static implicit operator LocalizedString(Func<string> func) => new LocalizedString(func);
```
pictos, line 34: Nica catch! But the linker could remove it, so we need to alert the linker.

```suggestion
        [Xamarin.Forms.Internals.Preserve(Conditional = true)]
		public static implicit operator LocalizedString(Func<string> func) => new LocalizedString(func);
```

pictos (changes requested):
@maxkoshevoi Thanks for this gem! I have some comments on it, feel free to ping me any time with any question! Also, is this a new feature?

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizedString.shared.cs
maxkoshevoi: Hm, good point. 
1) `IDisposable` won't help if nobody would call `Dispose`
2) Yes, I think this is the way, but we need to be able to have different instances of `LocalizedString` with different generators. Maybe we can use `PropertyChanged` event from `LocalizationResourceManager`, but then `LocalizedString` must be immutable

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizedString.shared.cs
maxkoshevoi: Yes, I wanted to give user ability to change generator instead of creating new instance, but maybe this functionality is not needed after all

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizedString.shared.cs
pictos: 1. We can override the Destructor of the class and let the GC call the Dispose from there, but you know isn't too safe because we can't control the GC.
2. Why do we need different generators?

maxkoshevoi: 1. Yes, this would work. I think this is the best option
2. In case strings will need to be changed. If generator is immutable the only way to change the string is to create a new instance. It's fine approach, just wanted to give more options. Yeah, changing generator would be conducing, better make it private

maxkoshevoi:
> @maxkoshevoi Thanks for this gem! I have some comments on it, feel free to ping me any time with any question! Also, is this a new feature?

Yes, this is new feature, glad you liked it =) I needed to somehow deal with StringFormat not being a thing (#743) so come up with this class

pictos:
@maxkoshevoi since this is a new feature you need to target the `develop` branch, I think that you will not have problems on that since they are pretty much the same.

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizedString.shared.cs
maxkoshevoi: Unfortunately it's not very useful since lambda expressions have no type, and `new(()=>{})` looks better then `(Func<string>)()=>{}`, but maybe this will be useful to someone.

Thanks for `Preserve`! I'll add it

pictos: > Unfortunately it's not very useful since lambda expressions have no type

For that, you can create an overload with an `Action`, o make it more generic. I need to play a little bit more with this Generator Func to see the usage and came with better (or bad) ideas.

```csharp
class LocalizedString<T>
{
     public Func<T> Generator { get ; set ; } 
}
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizedString.shared.cs
maxkoshevoi: Implemented `IDisposable`

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizedString.shared.cs
maxkoshevoi: Made generator private

maxkoshevoi:
> @maxkoshevoi can you add your new feature on our sample app, I guess this [the class ](https://github.com/xamarin/XamarinCommunityToolkit/blob/main/samples/XCT.Sample/ViewModels/SettingViewModel.cs) that you should add the feature as a sample

Done. Added app version string

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizedString.shared.cs
```diff
@@ -1,42 +1,40 @@
...
 
 		public event PropertyChangedEventHandler PropertyChanged;
 
-		public void Invalidate()
+		void Invalidate(object sender, PropertyChangedEventArgs e)
 			=> PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(null));
+
+		public void Dispose() => localizationManager.PropertyChanged -= Invalidate;
+
+		~LocalizedString() => Dispose();
```
pictos, line 37: If you are using the VS it will provide a code snippet for it, when you press `ctrl + .` But if you want I can setup it later or move this to the singleton pattern. :D

maxkoshevoi: I don't quite understand what do you mean by singleton. We need to have ability to create new `LocalizedString`s.
I there something wrong with current implementation of `IDisposable`?

pictos: Singleton is when we use one instance of that object during all application life-cycle, we do [this here](https://github.com/xamarin/XamarinCommunityToolkit/blob/4595d82a4b7fe383fc4da751d08239b33895642e/src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs#L11)

`public static LocalizationResourceManager Current { get; } = new LocalizationResourceManager();`

So we don't expose a constructor, normally we set a private constructor, and expose for users the object instance on that `Current` prop. that way we avoid more than one instance.

Is that make sense for you?

maxkoshevoi: I know what singleton is =)
I just don't understand how can we use it here. `LocalisationManager` has only one instance for a reason: it's state needs to be shared throughout the application.

In case of `LoclizedSitring`, each one has its own state (generator). How do you propose to make `LocalizedString` a singleton?

pictos: I'll take a look at that tomorrow, it's time to sleep for me (:

maxkoshevoi: Yeah, it's almost 5AM for me 😄
Good night

maxkoshevoi:
> @maxkoshevoi since this is a new feature you need to target the `develop` branch, I think that you will not have problems on that since they are pretty much the same.

Retargeted to develop. Do I need to revert commits that are not mine? (were present in main, but not in develop)

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizedString.shared.cs
maxkoshevoi: Lambda expressions don't have type, so it's impossible to write implicit converting from them. But it's still might be useful if someone will cast manually, or would have method that generates the string.

- Action doesn't make sense here, since the whole point of generator is to return value that will be displayed on the page, and Actions don't return value. Also, you'll still need to explicitly cast lambda expression to action for implicit conversion to work, so there is no difference here.
- Making class generic wouldn't help with implicitly casting lambda expressions, but would allow more methods to be converted to `LocalizedString`. Page will call `ToString` on the result anyway, but I think point of `LocalizedString` is to be a string, so it doesn't need to be generic. Could you provide some use case for generic `LocalizedString`?

jfversluis:
@maxkoshevoi do you by any change know how to do a rebase? :) That would be great!

maxkoshevoi:
> @maxkoshevoi do you by any chance know how to do a rebase? :) That would be great!

Didn't know how to do a rebase, and I think I partially figured it out. Rebased only my commits, but older commits are still here. How do I remove them?

pictos:
@maxkoshevoi looks like that your rebase changed other files, can you review it? If you want help, please let us know, we are here to help with anything :D

maxkoshevoi:
> @maxkoshevoi looks like that your rebase changed other files, can you review it? If you want help, please let us know, we are here to help with anything :D

@picos If you mean `AppResources` files, those are all my changes. They are autogenerated. Tested sample app with latest commit (40a9390), localization works as expected.

Also modified project file for Sample app a little so that `AppResources` will be auto regenerated on each change.

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizedString.shared.cs
maxkoshevoi: @pictos Inherited `LocalizedString` from `ObservableObject`. Now it should be garbage collected as soon as view model doesn't have reference to it (since `ObservableObject` uses week event), so destructor will trigger and it'll unsubscribe from `LocalizationResourceManager`. 
Does this look good to you?

pictos: Actually, Finalizers/deconstructor aren't deterministic, since they are triggered by GC we can't say that will be triggered. So the issue couldn't be a memory leak (but this doesn't invalidate memory leak), but other instances listen to that event and lead to odd behaviors. Said that I want to avoid it if it's possible. But you don't need to change your code right now, let's define what will be the best approach first.

For that can you explain better that `Func<string> Generate`? I feel that I didn't get his intent.

maxkoshevoi: Could you please clarify what kind of add behaviors we are trying to avoid? Listening to `localizationManager.PropertyChanged` is valid behavior for `LocalizedString` even we are not using it, since it actually will have different value when `localizationManager.PropertyChanged` is triggered. If object that's not already in use report that its value has changed that shouldn't be a problem.

I did some testing and `LocalizedString` will never be garbage collected until `Dispose` is called since it's subscribed to `localizationManager` (so finalizer is useless). We'll need to subscribe using `WeekReference` or something for `LocalizedString` to be automatically collected when the only reference to it is in `localizationManager`. Not sure how to do this for now.

> For that can you explain better that `Func<string> Generate`? I feel that I didn't get his intent.

`LocalizedString` is intended to be companion of `xct:Translate` extension. `Translate` is used in `Views` and automatically updates it's value when culture is changed, `LocalizedString` does the same thing, but is used in `ViewModels`. The only thing that's changing in `Translate` and `LocalizedString` is value of localized string (since different AppResource is used for different culture). That's why `LocalizedString` accepts only `Func<string>`. There is no point in putting `Func<object>` into it. If you have an object some string properties of which are read from AppResources, only those properties should be `LocalizedString`, not the whole object.

Intended use of `LocalizedString` class is to be placed instead of `string` properties in `ViewModel` if value of those properties is dependent on current culture.

I hope that makes sense. If it's still not clear, why `LocalizedString` accept only `Func<string>`, could you please provide some code sample of when accepting `Func<T>` would be useful?

maxkoshevoi: I think I got it. This extension method (inspired by [this article](https://faithlife.codes/blog/2008/08/event_subscription_using_weak_references/)) allows for objects that are subscribed using it to be collected while they are subscribed. I tested it and it works
image

```cs
public static class INotifyPropertyChangedEx
{
	public static void WeakSubscribe<T>(this INotifyPropertyChanged target, T subscriber, Action<T, object, EventArgs> action)
	{
		var weakSubscriber = new WeakReference(subscriber, false);
		target.PropertyChanged += handler;

		void handler(object sender, PropertyChangedEventArgs e)
		{
			var s = (T)weakSubscriber.Target;
			if (s is null)
			{
				target.PropertyChanged -= handler;
				return;
			}

			action.Invoke(s, sender, e);
		}
	}
}
```

Here is constructor of `WeekLocalizedString`

```cs
public WeekLocalizedString(LocalizationResourceManager localizationManager, Func<string> generator = null)
{
    this.generator = generator;
    localizationManager.WeakSubscribe(this, (t, _, _) =>
    {
        Console.WriteLine($"Week \"{t.Localized}\" OnPropertyChanged");
        t.OnPropertyChanged(null);
    });
}
```

maxkoshevoi: @pictos Refactored `LocalizedString` to use `WeakSubscribe` (and added test to ensure it is disposed when there are no references to it)

pictos: @maxkoshevoi I'll take a look by tomorrow... In meantime, could a MarkupExtension solve this issue in a more simple way?

pictos: BTW thanks for the info and article, I'll read all of those tomorrow. Have a good night :D

maxkoshevoi: @pictos Good point about MarkupExtension. I think it could, but I'm not sure. Don't really know what markup extensions capable of

maxkoshevoi: After sleeping on it, I don't think you can solve this with `MarkupExtension`. There are two reasons why strings can be provided from the ViewModel, instead on hardcoded into the View: 
1) They are dynamic. Different strings are used depending on the state of the view. This maybe can be solved by `MarkupExtension` that takes resource name from ViewModel and fetches its localized value.
2) And most common reason is that strings are generated via some logic. This cannot be solved with `MarkupExtension` since `MarkupExtension` will have how way to regenerate string with values from new culture.

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizedString.shared.cs
```diff
@@ -0,0 +1,30 @@
...
+		{
+			this.generator = generator;
+			localizationManager.WeakSubscribe(this, (t, sender, e) => t.OnPropertyChanged(null));
+		}
+
+		public string Localized => generator?.Invoke();
+
+		[Preserve(Conditional = true)]
+		public static implicit operator LocalizedString(Func<string> func) => new LocalizedString(func);
+	}
```
pictos, line 28: Can we add another implicit operator to convert a `string` into `LocalizedString`? That way we can just type this on VM

`public LocalizedString AppVersion { get; } = string.Format(AppResources.Version, AppInfo.VersionString);`
```suggestion
		[Preserve(Conditional = true)]
		public static implicit operator LocalizedString(Func<string> func) => new LocalizedString(func);
		
		[Preserve(Conditional = true)]
		public static implicit operator LocalizedString(string text) => new LocalizedString(() => text);
	}
```

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Helpers/LocalizedStringTests/LocalizedStringTests.cs
```diff
@@ -0,0 +1,70 @@
...
+using Xunit;
+
+namespace Xamarin.CommunityToolkit.UnitTests.Helpers.LocalizedStringTests
+{
+	public class LocalizedStringTests
+	{
+		public LocalizedStringTests()
+		{
+			resourceManager = new MockResourceManager();
+			localizationManager = new LocalizationResourceManager();
```
pictos, line 15: We shouldn't create new instances of this class. I already created a PR to fix it. And I just saw it thanks to your PR ❤
```suggestion
			localizationManager = LocalizationResourceManager.Current;
```

src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/Extensions/INotifyPropertyChangedEx.shared.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.ComponentModel;
+
+namespace Xamarin.CommunityToolkit.ObjectModel.Extensions
+{
+	public static class INotifyPropertyChangedEx
+	{
+		public static void WeakSubscribe<T>(this INotifyPropertyChanged target, T subscriber, Action<T, object, EventArgs> action)
```
pictos, line 8: I just love it! 
@brminnick do you think that we can add this helper and others (described [here](https://faithlife.codes/blog/2008/08/event_subscription_using_weak_references/) into your WeakEventManager? I think they could be very helpful for devs!

src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/Extensions/INotifyPropertyChangedEx.shared.cs
```diff
@@ -0,0 +1,26 @@
...
+		{
+			var weakSubscriber = new WeakReference(subscriber, false);
+			target.PropertyChanged += handler;
+
+			void handler(object sender, PropertyChangedEventArgs e)
+			{
+				var s = (T)weakSubscriber.Target;
+				if (s is null)
```
pictos, line 16: bear with, but we don't use pattern matching to check for null.
```suggestion
				if (s == null)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/Extensions/INotifyPropertyChangedEx.shared.cs
```diff
@@ -0,0 +1,26 @@
...
+				{
+					target.PropertyChanged -= handler;
+					return;
+				}
+
+				action.Invoke(s, sender, e);
```
pictos, line 22: Just to be safe... Or we can validate all args on this method.
```suggestion
				action?.Invoke(s, sender, e);
```

pictos (changes requested):
@maxkoshevoi Thank you very much for pushing this up and addressing my comments. I learned a lot with this PR and I hope you do too. For me are getting there, these are, probably, my last comments and after that, I think that we are good to go.

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizedString.shared.cs
maxkoshevoi: This (with proposed implicit conversion): 
`public LocalizedString AppVersion { get; } = string.Format(AppResources.Version, AppInfo.VersionString);`
Would be the same as this 
`public LocalizedString AppVersion { get; } = () => "Version: 1.0.0";`

If `LocalizedString` would receive already generated string there would be no way to modify it for different culture (string is just as array of `char`s, there is no culture information in it. And if culture changes, it would stay the same array or `char`s). This is why we need a function that can generate string on demand.

Closest thing to what you are proposing is to do implicit conversion from `FormattableString`, but [it's evaluation can be done only once](https://livebook.manning.com/book/c-sharp-in-depth-fourth-edition/chapter-9/point-9410-193-193-0), so it's also not an option (and I don't expect interpolated strings to be used is such way anyway).

src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/Extensions/INotifyPropertyChangedEx.shared.cs
maxkoshevoi: Glad you liked it, I had lots of fun creating it =)
This is already a public extension method, so users would be able to use it on anything that implements `INotifyPropertyChanged` (that's why I added it as an extension method instead of member of `LocalizationResourceManager`)

src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/Extensions/INotifyPropertyChangedEx.shared.cs
maxkoshevoi: Yeah, forgot about it. Do you now why this is the rule here? `is` doesn't call any equality overloads, so it always returns `true` if reference type has no value.

src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/Extensions/INotifyPropertyChangedEx.shared.cs
maxkoshevoi: Hm, good point. But I would do a null check earlier. If `action` is null, there is no point in subscribing in the first place.

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Helpers/LocalizedStringTests/LocalizedStringTests.cs
maxkoshevoi: Even for testing? A bit strange to call `Init()`on the same instance multiple times, but okay.

maxkoshevoi:
@pictos All done. Thanks for your feedback! `LocalizedString` is now much better than it was in the first commit thanks to it.

samples/XCT.Sample/ViewModels/SettingViewModel.cs
```diff
@@ -44,7 +47,7 @@ void LoadLanguages()
 				{ new Language(AppResources.English, "en") },
 				{ new Language(AppResources.Spanish, "es") }
 			};
-			SelectedLanguage = SupportedLanguages.FirstOrDefault(pro => pro.CI == LocalizationResourceManager.Current.CurrentCulture.TwoLetterISOLanguageName);
+			SelectedLanguage = SupportedLanguages.Single(pro => pro.CI == LocalizationResourceManager.Current.CurrentCulture.TwoLetterISOLanguageName);
```
pictos, line 50: This will crash if the app is using a different language. I, for example, am using "pt-BR" as Language, and that Single can't return a match result, that way the app crashes on me. Can you add a verification?
This was my workaround, isn't pretty but can give you some inspiration. 

```csharp
			Language? s = SupportedLanguages.FirstOrDefault(pro => pro.CI == LocalizationResourceManager.Current.CurrentCulture.TwoLetterISOLanguageName);

			if (s == null)
			{
				LocalizationResourceManager.Current.SetCulture(new CultureInfo("en"));
				s = new Language("Enlighs", "en");
			}

			SelectedLanguage = s.Value;
```

pictos (changes requested):
I'm testing here on my end, and I found two errors. One is in the comment below, another one happens on UWP when I see a Binding error. I need to tap two times in the button, to change the AppVersion language. Could you check it?

image

src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/Extensions/INotifyPropertyChangedEx.shared.cs
pictos: Yeap, I agree with you that `is null` and `is { }` ( != null) is better to check nulls. I'm not sure about it but I believe that's because of the Xamarin pattern. All Xamarin projects (Forms, Essentials, etc) use the `==` and `!=` pattern.

src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/Extensions/INotifyPropertyChangedEx.shared.cs
pictos: Yeah, I saw that is public for everyone to use it <3. But I want to see if we can improve our WeakEvent manager implementation and also bring the other methods

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizedString.shared.cs
pictos: Yeah, you're right... I forgot that detail. Good catch

src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/Extensions/INotifyPropertyChangedEx.shared.cs
maxkoshevoi: Got it. BTW as of C# 9 you can use `is not null`

samples/XCT.Sample/ViewModels/SettingViewModel.cs
maxkoshevoi: Missed that.. I think, I'll just revert that change since it's out of scope of this PR, and fix just brings unneeded complexity

maxkoshevoi:
> I need to tap two times in the button, to change the AppVersion language.

You need to tab button twice to change localization of everything on the page (not only AppVersion), so this is the problem with `LocalizationResourceManager`.
And you won't believe it, but #749 fixes it as well. Just merge it into develop =)

Regarding binding error, it was here before this PR. Notice how language select Picker becomes empty if you click the button two or more times in a row. This empty string understandably "cannot be converted to type 'Xamarin.CommunityToolkit.Sample.Models.Language'".

pictos:
@maxkoshevoi thanks for taking a look at those, I'll run other tests on this PR, hoping to approve by tomorrow.

pictos (approved):

themronion:
@maxkoshevoi So, before when we set our strings like this - someString = AppResources.LocText and it wasn't updated. We now should use ur class instead of string and it will be updated when we switch the language. Am i right?

maxkoshevoi:
> @maxkoshevoi So, before when we set our strings like this - someString = AppResources.LocText and it wasn't updated. We now should use ur class instead of string and it will be updated when we switch the language. Am i right?

Yes, that's exactly the point! You should replace every string-typed property in your ViewModel with `LocalizedString` one and add `.Localized` to it in xaml. After this it will be automatically updated when culture changes.

themronion:
And this works now with the current ResourceManager in the main branch or only with your latest pull requests?

maxkoshevoi:
> And this works now with the current ResourceManager in the main branch or only with your latest pull requests?

This works with current `LocalizationResourceManager` in the main branch. You can copy `LocalizedString` class in your project and use it right now (`LocalizationResourceManager` has a bug though. It's fixed, but fix is not released yet #749)

themronion:
Ok, i ll try it out, thx. I hope this is finally solves the problem of app reloading to switch the language completely

themronion:
> > And this works now with the current ResourceManager in the main branch or only with your latest pull requests?
> 
> This works with current `LocalizationResourceManager` in the main branch. You can copy `LocalizedString` class in your project and use it right now (`LocalizationResourceManager` has a bug though. It's fixed, but fix is not released yet #749)

Unfortunately failed to just add the class to my project - it didn't find the extension method WeakSubscribe in localization manager

PR #763, opened 2021-01-15
Added new TextCaseType to TextCaseConverter

JoacimWall:
Added support for first char to upper rest to lower case.

### Description of Change ###

Added new enum and logic for that enum to convert a string to first char to upper rest to lower case.

Sample No need for project sample i think:
<Label Text="{Binding Source={x:Static resource:AppResources.Logout}, Converter={StaticResource TextCaseConverter}, ConverterParameter={x:Static converters:TextCaseType.FirstUpperRestLower}}"/>

Regarding  PR check below i really don't know how to fix the rest check boxes. 
  ### Bugs Fixed ###

- Fixes #

### API Changes ###

None
Added: 
 
Added 
FirstUpperRestLower to the TextCaseType enum
Changed:

none

### Behavioral Changes ###

none

### PR Checklist ###

- [X] Has tests (if omitted, state reason in description)

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
@@ -39,7 +39,7 @@ namespace Xamarin.CommunityToolkit.Converters
 				TextCaseType.Upper => value?.ToUpperInvariant(),
 				TextCaseType.FirstUpperRestLower => !string.IsNullOrWhiteSpace(value) 
 					? value.Substring(0, 1).ToUpperInvariant() + value.Substring(1).ToLowerInvariant()
-					: value
+					: value,
 				_ => value
 			};
 
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
@@ -38,7 +38,6 @@ namespace Xamarin.CommunityToolkit.Converters
 				TextCaseType.Lower => value?.ToLowerInvariant(),
 				TextCaseType.Upper => value?.ToUpperInvariant(),
 				TextCaseType.FirstUpperRestLower => value?.Substring(0, 1).ToUpperInvariant() + value?.Substring(1).ToLowerInvariant(),
-
 				_ => value
 			};
 
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
@@ -37,7 +37,7 @@ namespace Xamarin.CommunityToolkit.Converters
 			{
 				TextCaseType.Lower => value?.ToLowerInvariant(),
 				TextCaseType.Upper => value?.ToUpperInvariant(),
-				TextCaseType.FirstUpperRestLower => !string. IsNullOrWhiteSpace(value) 
+				TextCaseType.FirstUpperRestLower => !string.IsNullOrWhiteSpace(value) 
 					? value.Substring(0, 1).ToUpperInvariant() + value.Substring(1).ToLowerInvariant()
 					: value
 				_ => value
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
@@ -37,6 +37,8 @@ namespace Xamarin.CommunityToolkit.Converters
 			{
 				TextCaseType.Lower => value?.ToLowerInvariant(),
 				TextCaseType.Upper => value?.ToUpperInvariant(),
+				TextCaseType.FirstUpperRestLower => value?.Substring(0, 1).ToUpperInvariant() + value?.Substring(1).ToLowerInvariant(),
+
 				_ => value
 			};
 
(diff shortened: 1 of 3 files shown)


AndreiMisiukevich:
@JoacimWall please sign the CLA

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
```diff
@@ -37,6 +37,8 @@ object Convert(string value, object parameter)
 			{
 				TextCaseType.Lower => value?.ToLowerInvariant(),
 				TextCaseType.Upper => value?.ToUpperInvariant(),
+				TextCaseType.FirstUpperRestLower => value?.Substring(0, 1).ToUpperInvariant() + value?.Substring(1).ToLowerInvariant(),
+
```
AndreiMisiukevich, line 41:
```suggestion
```

AndreiMisiukevich (changes requested):
Please add unit tests (and perhaps update the sample app).

Current implementation causes the exception if string is empty or has 1 character

JoacimWall:
> Please add unit tests (and perhaps update the sample app).
> Added 
[InlineData(test, TextCaseType.FirstUpperRestLower, "Test")]
[InlineData(t, TextCaseType.FirstUpperRestLower, "T")]
[InlineData("", TextCaseType.FirstUpperRestLower, "")]
> Current implementation causes the exception if string is empty or has 1 character
Fixed only if string empty .1 char was working.

The project referens in xamarin.communityToolkit 
<ProjectReference Include="..\..\Markup\Xamarin.CommunityToolkit.Markup\Xamarin.CommunityToolkit.Markup.csproj" />  
does not match filesystem needed to change to ProjectReference Include="..\Markup...
to be able to run test. I don't check in this change.

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
```diff
@@ -37,6 +37,7 @@ object Convert(string value, object parameter)
 			{
 				TextCaseType.Lower => value?.ToLowerInvariant(),
 				TextCaseType.Upper => value?.ToUpperInvariant(),
+				TextCaseType.FirstUpperRestLower => string.IsNullOrEmpty(value) ? string.Empty : value?.Substring(0, 1).ToUpperInvariant() + value?.Substring(1).ToLowerInvariant(),
```
AndreiMisiukevich, line 40:
```suggestion
				TextCaseType.FirstUpperRestLower => !string. IsNullOrWhiteSpace(value) 
					? value.Substring(0, 1).ToUpperInvariant() + value.Substring(1).ToLowerInvariant()
					: value
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
```diff
@@ -37,6 +37,9 @@ object Convert(string value, object parameter)
 			{
 				TextCaseType.Lower => value?.ToLowerInvariant(),
 				TextCaseType.Upper => value?.ToUpperInvariant(),
+				TextCaseType.FirstUpperRestLower => !string. IsNullOrWhiteSpace(value)
```
AndreiMisiukevich, line 40:
```suggestion
				TextCaseType.FirstUpperRestLower => !string.IsNullOrWhiteSpace(value) 
```

AndreiMisiukevich (approved):

sthewissen:
I would perhaps rename the enum value to `FirstCharacterUppercase` as that already implies the rest is lowercase.

Furthermore, I wonder if this is meant to be something like TitleCase, which is actually culture dependant and can be done through a TextInfo object gathered from the current app culture?

AndreiMisiukevich:
> I would perhaps rename the enum value to `FirstCharacterUppercase` as that already implies the rest is lowercase.
> 
> Furthermore, I wonder if this is meant to be something like TitleCase, which is actually culture dependant and can be done through a TextInfo object gathered from the current app culture?

Title case sounds good, but just "FirstUpper" won't be enough (It sounds like "first upper and other as is")

JoacimWall:
> Please add unit tests (and perhaps update the sample app).
> 
> Current implementation causes the exception if string is empty or has 1 character

Ok

> I would perhaps rename the enum value to `FirstCharacterUppercase` as that already implies the rest is lowercase.
> 
> Furthermore, I wonder if this is meant to be something like TitleCase, which is actually culture dependant and can be done through a TextInfo object gathered from the current app culture?

Hi 
the ide is that if you have a string that is HELLO it should resoult in Hello. and if the input string is hello it should result in Hello.  In the app we use it we get strings from Resource file som time this string is HELLLO  then later som oe on desging says we need the samt string somewhere elles but this time it should be Hello then we dont want add an extra string to the resource file and all translations resurcefiles just use a converter. 

as you can se in the converter code it just not make the frist upper ot also make the rest lower 
value?.Substring(0, 1).ToUpperInvariant() + value?.Substring(1).ToLowerInvariant()
//Regards joacim

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
```diff
@@ -37,6 +37,9 @@ object Convert(string value, object parameter)
 			{
 				TextCaseType.Lower => value?.ToLowerInvariant(),
 				TextCaseType.Upper => value?.ToUpperInvariant(),
+				TextCaseType.FirstUpperRestLower => !string.IsNullOrWhiteSpace(value) 
+					? value.Substring(0, 1).ToUpperInvariant() + value.Substring(1).ToLowerInvariant()
+					: value
```
pictos, line 42: you missed the `,` 
```suggestion
					: value,
```

pictos (changes requested):
Just a small observation (:

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
AndreiMisiukevich: It was my bad, I guess)

pictos (approved):
@JoacimWall Thank you for your contribution ❣

JoacimWall:
Hi 
One question when will the feature be in the release version?
//Regards Joacim

AndreiMisiukevich:
@JoacimWall I guess you can find it in the pre-release version of NuGet now.

PR #766, opened 2021-01-16
Deprecate LocalizationResourceManager.SetCulture

### LINKED ISSUE #757: [Spec]  Deprecate LocalizationResourceManager.SetCulture method in favor of CurrentCulture property
# [The feature] 

Add setter to `CurrentCulture` and deprecate `SetCulture()`. 

Culture is handled via properties everywhere (`CultureInfo.CurrentCulture`, `Thread.CurrentThread.CurrentUICulture`, `AppResources.Culture`), so it feels like `LocalizationResourceManager.CurrentCulture` should have setter too. 
Especially accounting for when #749 will be merged culture will be stored in the backing field anyway. Feels wrong having get-only property to get value of a backing field and separate method to set it

# API

## [ class ]

`LocalizationResourceManager`

### Changes

```cs
[Obsolete("Use " + nameof(CurrentCulture) + " to set culture")]
public void SetCulture(CultureInfo language) => CurrentCulture = language;

public CultureInfo CurrentCulture
{
	get => currentCulture;
	set
	{
		currentCulture = value;
		Invalidate();
	}
}
```
 
# Scenarios

# Platform Compatibility
- Target Frameworks: 
  - iOS:  
  - Android: 
  - UWP: 
# Backward Compatibility

Users will see warning when using `SetCulture()`

# Difficulty : low

maxkoshevoi:
### Description of Change ###

Deprecates `SetCulture` in favor of `CurrentCulture`

### Bugs Fixed ###

- Fixes #757

### API Changes ###
Added: 
 
- `CurrentCulture { set; }`

Removed:

 - `LocalizationResourceManager.Invalidate`

### PR Checklist ###


jfversluis:
@maxkoshevoi I think that one was merged? Could you update to the right branch and rebase?

maxkoshevoi:
@jfversluis Done

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs
```diff
@@ -29,18 +29,14 @@ public void Init(ResourceManager resource, CultureInfo initialCulture)
 		public string this[string text] =>
 			GetValue(text);
 
-		public void SetCulture(CultureInfo language)
+		[Obsolete("Use " + nameof(CurrentCulture) + " to set culture")]
```
AndreiMisiukevich, line 32: Please add EditorBrowsableState.Never attribute

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs
```diff
@@ -29,18 +29,14 @@ public void Init(ResourceManager resource, CultureInfo initialCulture)
...
+			get => currentCulture;
+			set => SetProperty(ref currentCulture, value, null);
 		}
-
-		public CultureInfo CurrentCulture => currentCulture;
-
-		public event PropertyChangedEventHandler PropertyChanged;
-
-		public void Invalidate() =>
-			PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(null));
```
AndreiMisiukevich, line 43: @jfversluis can we just remove it? Or we should mark it as obsolete and editorbrowsablestate.never

maxkoshevoi: This method doesn't make sense anymore. Culture is stored in the private field, and when we change it, we always call `PropertyChanged` ourselves. So `public Invalidate` method will just needlessly send an event.

It was needed earlier since culture was stored in `Thread.CurrentThread.CurrentUICulture` that could be changed outside of `LocalizationResourceManager`

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs
```diff
@@ -29,18 +30,15 @@ public void Init(ResourceManager resource, CultureInfo initialCulture)
 		public string this[string text] =>
 			GetValue(text);
 
-		public void SetCulture(CultureInfo language)
+		[Obsolete("Use " + nameof(CurrentCulture) + " to set culture")]
```
pictos, line 33:
```suggestion
		[Obsolete("Please, use " + nameof(CurrentCulture) + " to set culture")]
```

pictos (approved):

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs
jfversluis: I see what you're saying, I think it would make sense to first release a version where it's deprecated though. That way people will get a heads up that it is going away and gives them the opportunity to change their code at their pace instead of just dropping a breaking change over the fence that they have to deal with without any notice upfront.

Does that make sense?

maxkoshevoi: Makes sense, returned the method. When are we planning to remove everything that is marked as `Obsolete`?

jfversluis: I would say 1 version obsolete, next version it's gone :)

jfversluis:
If you could fix the conflict @maxkoshevoi I think we can merge! I would do it, but I'm unsure what is the right change here :)

PR #279, opened 2020-09-09
AsyncCommand

YZahringer:
### Description of Change ###

Async implementation of `ICommand`.

Based on [MVVM Helpers](https://github.com/jamesmontemagno/mvvm-helpers/blob/master/MvvmHelpers/Commands/AsyncCommand.cs) and [AsyncAwaitBestPractices](https://github.com/brminnick/AsyncAwaitBestPractices/blob/main/Src/AsyncAwaitBestPractices.MVVM/AsyncCommand.cs).

### Bugs Fixed ###

- Fixes #272 

### API Changes ###
```csharp
public interface IAsyncCommand<in T> : ICommand
{
    /// <summary>
    /// Allow simultaneous/re-entrant execution of async Command. This is reflected on CanExecute result.
    /// Default value is False.
    /// </summary>
    bool AllowMultipleExecution { get; set; }

    /// <summary>
    /// Executes the Command as a Task
    /// </summary>
    /// <returns>The Task to execute</returns>
    /// <param name="parameter">Data used by the command. If the command does not require data to be passed, this object can be set to null.</param>
    Task ExecuteAsync(T parameter);

    /// <summary>
    /// Raises the CanExecuteChanged event.
    /// </summary>
    void RaiseCanExecuteChanged();
}

public interface IAsyncCommand : IAsyncCommand<object>
{
}
```
`internal WeakEventManager` added until it becomes public on XF.

### Behavioral Changes ###

None

### PR Checklist ###

- [x] Has tests (if omitted, state reason in description)
- [x] Has samples (if omitted, state reason in description)
- [x] Rebased on top of main at time of PR
- [x] Changes adhere to coding standard
- [ ] Updated documentation

PR #772, opened 2021-01-18
Alter AsyncCommand parameters to match Command ones, and add CommandHelper (#771)

### LINKED ISSUE #771: [Enhancement] Update AsyncCommand constructor parameters to match Command ones
## Summary

Related to #709

This repository has `AsyncCommand` class that accompanies Xamarin's `Command` class, but for `async` methods. 
It's great, but has two issues:
1) Sometimes it accepts `object` as parameter. This is not ideal. We should use static typing whenever we can (that's why generics were created). And we can do this here with little API change. 
2) `AsyncCommand` and `Command` should accept the same parameters (that are common for both classes), but right not this is not always the case:
- `canExecute` parameter for `AsyncCommand<T>` accept `Func<object, bool>` instead of `Func<T, bool>`? It's Func<T, bool> for regular `Command<T>`.
- `canExecute` parameter for `AsyncCommand` (without T) accept `Func<object, bool>` instead of `Func<bool>`? It's `Func<bool>` for regular `Command`.

They should the same parameters so user can seamlessly switch between them (methods become async and sync again all the time).

### Common place to create all the commands

Also it would be great to create some `CommandHelper.Create()` that would automatically resolve what command (sync or async) you're trying to create and return an instance of it. This is helpful because id some method will become async or sync, `Create()` would automatically return command of needed type. And in general, when creating a command user wouldn't need to think what type it would be, they just need to call `Create` on whatever method they have.

## Intended Use Case

User would be able to change `new Command(a, b)` to `new AsyncCommand(a, b)` (and `new Command<T>(a, b)` to `new AsyncCommand<T>(a, b)`) and `a` and `b` would still be valid arguments.

## Who Will Do The Work?

maxkoshevoi:
### Description of Change ###

- Updated `AsyncCommand` constructor parameters so they are consistent with `Command`s ones.
- Created `CommandHelper` helper class with `Create` method that can create both types of command based on provided parameters.
- Updated sample app to use `CommandHelper`
- Updated sample app to use `xct:EventToCommandBehavior` to map `Appearing` event to ViewModel's command

### Bugs Fixed ###

- Fixes #771
- Related to #709

### API Changes ###

- `AsyncCommand<T>` is now equivalent to `AsyncCommand<T, T>` instead of `AsyncCommand<T, object>`
- `AsyncCommand`s `canExecue` parameter has type `Func<bool>` instead of `Func<object, bool>`
- `static CommandHelper.Create()`

User can still pass `object`s into `canExecue`, but they'll need to explicitly ont-in into that: `AsyncCommand<T, object>`

### Behavioral Changes ###

This is breaking change that will affect users that pass `canExecute` as `Func<object, bool>` instead of creating with two type arguments to specify type of `canExecute`. So this change will push those users to a better (more statically checked) code.

### PR Checklist ###

### DIFF
--- a/samples/XCT.Sample/Pages/AboutPage.xaml
+++ b/samples/XCT.Sample/Pages/AboutPage.xaml
@@ -14,7 +14,11 @@
     <pages:BasePage.BindingContext>
         <viewmodels:AboutViewModel />
     </pages:BasePage.BindingContext>
-    
+
+    <ContentPage.Behaviors>
+        <xct:EventToCommandBehavior EventName="Appearing" Command="{Binding PageAppearingCommand}"/>
+    </ContentPage.Behaviors>
+
     <pages:BasePage.Content>
         <StackLayout Padding="{StaticResource ContentPadding}">
             <Label Text="Thank you Xamarin Community Toolkit contributors"
--- a/samples/XCT.Sample/Pages/AboutPage.xaml.cs
+++ b/samples/XCT.Sample/Pages/AboutPage.xaml.cs
@@ -1,9 +1,4 @@
 ﻿using System;
-using System.Linq;
-using Octokit;
-using Xamarin.CommunityToolkit.Sample.ViewModels;
-using Xamarin.Essentials;
-using Xamarin.Forms;
 
 namespace Xamarin.CommunityToolkit.Sample.Pages
 {
@@ -11,12 +6,6 @@ namespace Xamarin.CommunityToolkit.Sample.Pages
 	{
 		public AboutPage()
 			=> InitializeComponent();
-
-		protected override async void OnAppearing()
-		{
-			base.OnAppearing();
-			await ((AboutViewModel)BindingContext).OnAppearing();
-		}
 
 		async void OnCloseClicked(object sender, EventArgs e)
 			=> await Navigation.PopModalAsync();
--- a/samples/XCT.Sample/Pages/Base/BasePage.cs
+++ b/samples/XCT.Sample/Pages/Base/BasePage.cs
@@ -1,6 +1,6 @@
 ﻿using System;
 using System.Windows.Input;
-using Xamarin.CommunityToolkit.ObjectModel;
+using Xamarin.CommunityToolkit.Helpers;
 using Xamarin.CommunityToolkit.Sample.Models;
 using Xamarin.Forms;
 
@@ -8,12 +8,12 @@ namespace Xamarin.CommunityToolkit.Sample.Pages
 {
 	public class BasePage : ContentPage
 	{
-		ICommand navigateCommand;
+		public BasePage() =>
+			NavigateCommand = CommandHelper.Create<SectionModel>(sectionModel => Navigation.PushAsync(PreparePage(sectionModel)));
 
 		public Color DetailColor { get; set; }
 
-		public ICommand NavigateCommand => navigateCommand ??= new AsyncCommand<SectionModel>(sectionModel
-			=> Navigation.PushAsync(PreparePage(sectionModel)));
+		public ICommand NavigateCommand { get; }
 
 		Page PreparePage(SectionModel model)
 		{
--- a/samples/XCT.Sample/Pages/Converters/ByteArrayToImageSourcePage.xaml
+++ b/samples/XCT.Sample/Pages/Converters/ByteArrayToImageSourcePage.xaml
@@ -16,6 +16,10 @@
          </ResourceDictionary>
     </ContentPage.Resources>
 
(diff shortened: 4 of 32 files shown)


maxkoshevoi:
@brminnick, @AndreiMisiukevich Could you please take a look at this?

brminnick:
Maksym,
I'm sorry to have to close this PR after you have spent hours of your time working on this. 

Ten days ago, in the [Discussions tab of this repo](https://github.com/xamarin/XamarinCommunityToolkit/discussions/709), we discussed these changes and why they are not recommended.

**tl;dr** This is a breaking change that removes & duplicates existing functionality and violates the [`ICommand` spec](https://docs.microsoft.com/dotnet/api/system.windows.input.icommand?view=netcore-3.1&WT.mc_id=xamarincommunitytoolkit-github-bramin).

Below are the answers from our [previous discussion](https://github.com/xamarin/XamarinCommunityToolkit/discussions/709), copy/pasted here for visibility:

## [Question 1 ](https://github.com/xamarin/XamarinCommunityToolkit/discussions/709#discussioncomment-269961)

> Why does `canExecute` parameter for `AsyncCommand<T>` accept `Func<object, bool>` instead of `Func<T, bool>`? It's `Func<T, bool>` for regular `Command<T>`.

### Answer
 In `AsyncCommand<T>`, `T` is the type for the parameter for `AsyncCommand.Execute(T parameter)`. 

We [purposely did not use `T` for `canExecute<object, bool>`](https://github.com/xamarin/XamarinCommunityToolkit/pull/279#issuecomment-692882191) to allow for scenarios where `CanExecute` and and `Execute` use different parameters. For example, here is a case where we are using an `int` for the Execute parameter and a `bool` for the CanExecute parameter:

```csharp
public ICommand PullToRefreshCommand { get; } = new AsyncCommand((int) count => FetchNewData(count), (bool) isBusy => !isBusy)
```

This implementation best honors the paradigm of [`ICommand`](https://docs.microsoft.com/dotnet/api/system.windows.input.icommand?view=netcore-3.1&WT.mc_id=xamarincommunitytoolkit-github-bramin) which doesn't require `Execute` to use the same parameter type as `CanExecute`.

I strongly disagree with Xamarin.Forms' original implementation of `Command<T>` that assigns `T` to both `Execute` and `CanExecute`, because the [`ICommand Specification`](https://docs.microsoft.com/dotnet/api/system.windows.input.icommand?view=netcore-3.1&WT.mc_id=xamarincommunitytoolkit-github-bramin) does not require `Execute` and `CanExecute` to use the same parameter type.

### Recommendation

For a type-safe `canExecute`, I recommend using `AsyncCommand<TExecute, TCanExecute>` like so:

```csharp
public ICommand PullToRefreshCommand { get; } = new AsyncCommand<int,bool>(count => FetchNewData(count), isBusy => !isBusy)
```

## [Question 2](https://github.com/xamarin/XamarinCommunityToolkit/discussions/709#discussioncomment-269981)

> Why does `canExecute` parameter for `AsyncCommand` (without `T`)  accept `Func<object, bool>` instead of `Func<bool>`? It's `Func<bool>` for regular `Command`.

### Answer

Similar to the [Answer for Question #1](https://github.com/xamarin/XamarinCommunityToolkit/discussions/709#discussioncomment-269961), `Execute` and `CanExecute` are separate and are not required by the [ICommand Specification](https://docs.microsoft.com/dotnet/api/system.windows.input.icommand?view=netcore-3.1&WT.mc_id=xamarincommunitytoolkit-github-bramin) to use the same parameter type. 

I.e. Foregoing an input parameter for `Execute` does predicate that you must forego an input parameter for `CanExecute`.

Here is an example where `Execute` does not use an input parameter and `CanExecute` _does_ use an input parameter:

```csharp
public ICommand PullToRefreshCommand { get; } = new AsyncCommand(() => FetchNewData(count), (bool) isBusy => !isBusy)
```

Again, I strongly disagree with Xamarin.Forms' implementation of `Command` as it breaks the `ICommand` paradigm.

maxkoshevoi:
@brminnick Could you please take a closer look at `AsyncCommad` implementation in this PR. 
1) It still allows to pass parameters with different types to `Execute` and `CanExecute`, but now user need to explicitly specify those types, so it doesn't break `ICommand Specification`. 
2) No `AsyncCommad` constructor parameters default to `object`. If user needs to pass `object` in `Execute` or `CanExecute` it needs to be explicitly specified.  
3) As I described in PR description: This is breaking change that will affect users that pass `canExecute` as `Func<object, bool>` instead of creating with two type arguments to specify type of `canExecute`. So this change will push those users to a better (more statically checked) code.

Could you please explain why do you disagree with all or any of those points?

PR #797, opened 2021-01-23
Add CommandFactory class

### LINKED ISSUE #798: [Enhancement] Add CommandFactory class
## Summary

## API Changes

## Intended Use Case

## Who Will Do The Work?

maxkoshevoi:
#### Common place to create all the commands

Added `CommandFactory.Create()` that automatically resolves what command (sync or async) user is trying to create and returns an instance of it. 

**Note**: 
- `CommandFactory.Create` accepts exactly the same arguments for `Command` and `AsyncCommad`
- `CommandFactory.Create` converts `Command` constructors to match `AsyncCommad` ones 
  - `Command` now can be created with two type arguments
  - If `Command` is created with one type argument, `canExecute` has `Func<bool>` type
  - `CommandFactory` never creates `Command<T>` since it forces both `execute` and `canExecute` to accept argument with the same type, which ([as motioned by @brminnick](https://github.com/xamarin/XamarinCommunityToolkit/issues/771#issuecomment-762391817)) violates `ICommand` spec

#### What was done
- Created `CommandFactory` helper class with `Create` method that can create both types of command based on provided parameters.
- Updated sample app to use `CommandFactory`
- Updated sample app to use `xct:EventToCommandBehavior` to map `Appearing` event to ViewModel's command

#### Intended use case

Just replace `new Command` or `new AsyncCommand` with `CommandFactory.Create`:
image

In this example `CommandFactory` will return `AsyncCommand`
image

### Bugs Fixed ###

- Fixes #798

### API Changes ###

### Behavioral Changes ###

None

### PR Checklist ###


maxkoshevoi:
@brminnick Could you please take a look at this? 
I think you'll like this one since now even `Command` doesn't violate `ICommand` spec

roubachof:
IMHO CommandFactory would be a better name 😀

samples/XCT.Sample/ViewModels/AboutViewModel.cs
```diff
@@ -18,7 +18,18 @@ public class AboutViewModel : BaseViewModel
 
 		string emptyViewText = "Loading data...";
 
-		ICommand selectedContributorCommand;
+		public AboutViewModel()
+		{
+			PageAppearingCommand = CommandFactory.Create(OnAppearing);
+			SelectedContributorCommand = CommandFactory.Create(async () =>
+			{
+				if (SelectedContributor is null)
```
pictos, line 26:
```suggestion
				if (SelectedContributor == null)
```

samples/XCT.Sample/ViewModels/Markup/SearchViewModel.cs
```diff
@@ -54,17 +52,24 @@ public SearchViewModel()
...
-		public ICommand OpenTwitterSearchCommand => openTwitterSearchCommand ??= new RelayCommandAsync(OpenTwitterSearch);
+		public ICommand OpenTwitterSearchCommand { get; }
 
-		public ICommand OpenHelpCommand => openHelpCommand ??= new RelayCommandAsync(OpenHelp);
+		public ICommand OpenHelpCommand { get; }
 
-		void Back() { }
+		void Back()
+		{
+		}
```
pictos, line 72: can you revert this change, please?

samples/XCT.Sample/ViewModels/Views/GravatarImageViewModel.cs
```diff
@@ -9,17 +11,7 @@ public class GravatarImageViewModel : BaseViewModel
...
-			DefaultGravatar.Blank,
-			DefaultGravatar.FileNotFound,
-			DefaultGravatar.Identicon,
-			DefaultGravatar.MonsterId,
-			DefaultGravatar.MysteryPerson,
-			DefaultGravatar.Retro,
-			DefaultGravatar.Robohash,
-			DefaultGravatar.Wavatar
-		};
+		public DefaultGravatar[] Defaults { get; } = Enum.GetValues(typeof(DefaultGravatar)).Cast<DefaultGravatar>().ToArray();
```
pictos, line 14: Can you revert this, please? I think that idea to use all Enums explicitly is to make it easier for devs to find all possible values.

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/CommandFactory.shared.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System;
+using System.Threading.Tasks;
+using Xamarin.CommunityToolkit.ObjectModel;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Helpers
+{
+	public static class CommandFactory
+	{
+		public static Command Create(Action execute, Func<bool> canExecute = null) =>
```
pictos, line 10: Why don't return `ICommand` just like the other `Create` methods that return `IAsyncCommand`?

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/CommandFactory.shared.cs
```diff
@@ -0,0 +1,67 @@
...
+			Func<TExecute, Task> execute,
+			Func<TCanExecute, bool> canExecute = null,
+			Action<Exception> onException = null,
+			bool continueOnCapturedContext = false,
+			bool allowsMultipleExecutions = true) =>
+			new AsyncCommand<TExecute, TCanExecute>(execute, canExecute, onException, continueOnCapturedContext, allowsMultipleExecutions);
+
+		static Action<object> ConvertExecute<T>(Action<T> execute)
+		{
+			if (execute is null)
```
pictos, line 45: we don't use pattern matching for null check, can you fix this one and others?
```suggestion
			if (execute == null)
```

pictos (changes requested):
Hey there, thanks for your contribution. I just take a quick look at your code and came up with these sugesstions.

pictos:
@maxkoshevoi what's the issue related to this PR?

samples/XCT.Sample/ViewModels/Markup/SearchViewModel.cs
maxkoshevoi: This is style related change. There was a warning for this `void Back() { }`

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/CommandFactory.shared.cs
maxkoshevoi: `ICommand` doesn't have `ChangeCanExecute` method while `IAsyncCommand` does have `RaiseCanExecuteChanged`.
If it returned `ICommand` this would hide `ChangeCanExecute` method that is needed to notify of `canExecute` value change.

pictos: Got it. So, in our samples, should we change all `ICommand` to `Command`?

maxkoshevoi: - If ViewModel needs to execute `ChangeCanExecute()` then property must be `Command`, if not, it can be `ICommand`. 
- If property is `ICommand` both `Command` and `AsyncCommand` can be asigned to if

I don't have strong opinion in any direction here.

pictos: @maxkoshevoi Ok, agreed. Do you mind creating a sample page for that? One that we need to execute the ChangeCanExecute, just to have a reference for the devs?

samples/XCT.Sample/ViewModels/Markup/SearchViewModel.cs
pictos: Odd I can remember that's our pattern, or am I just think about another project? Well, so you can ignore this one.

maxkoshevoi: Reverted the change

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/CommandFactory.shared.cs
maxkoshevoi: That's just basic Xamarin MVVM staff. This page wouldn't demonstrate anything from XCT. Are you sure it needs to be added?

pictos: IF you think that it doesn't add any value, it's ok. I was just guessing. TBH Command isn't my main area here.

maxkoshevoi: It will have value (who doesn't like samples), but it wouldn't demonstrate anything that could be done only with XCT. `ICommand` interface and `Command` class are standard tools that Xamarin developers are already using. 

TLDR: `ChangeCanExecute` is a method that needs to be called so that Xamarin would rerun `canExecute` function for the given command and update page layout according to result.

pictos: got it!

maxkoshevoi:
> @maxkoshevoi what's the issue related to this PR?

Created #798 for this PR

pictos:
@maxkoshevoi I'll give my review by tomorrow or today late, but I think that Brain is the right person to merge it, so if he approvals it before me this is good to go.

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/CommandFactory.shared.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System;
+using System.Threading.Tasks;
+using Xamarin.CommunityToolkit.ObjectModel;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Helpers
```
brminnick, line 6: Let's move this to the namespace `Xamarin.CommunityToolkit.ObjectModel`

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/CommandFactory.shared.cs
```diff
@@ -0,0 +1,67 @@
...
+			new AsyncCommand<TExecute>(execute, canExecute, onException, continueOnCapturedContext, allowsMultipleExecutions);
+
+		public static IAsyncCommand<TExecute, TCanExecute> Create<TExecute, TCanExecute>(
+			Func<TExecute, Task> execute,
+			Func<TCanExecute, bool> canExecute = null,
+			Action<Exception> onException = null,
+			bool continueOnCapturedContext = false,
+			bool allowsMultipleExecutions = true) =>
+			new AsyncCommand<TExecute, TCanExecute>(execute, canExecute, onException, continueOnCapturedContext, allowsMultipleExecutions);
+
```
brminnick, line 42: Let's add support for `IAsyncValueCommand`

brminnick (changes requested):
Thanks @maxkoshevoi!

Let's just tweak a couple things:
- Move `CommandFactory` to the namespace `Xamarin.CommunityToolkit.ObjectModel`
- Add support for `IAsyncValueCommand`
- Write Unit Tests for `CommandFactory`

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/CommandFactory.shared.cs
maxkoshevoi: By support for `IAsyncValueCommand` you mean creating new `CreateValue()` method that would return value commands?
From what I now, Xamarin doesn't have `ValueCommand`, so what should this method return for sync actions?

That's why I didn`t include `IAsyncValueCommand` here in the first place - I couldn't figure out how to do it. Do you have any ideas on how this could be done?

maxkoshevoi:
@brminnick Updated `ConvertExecute` and `ConvertCanExecute` to throw `InvalidCommandParameterException` in case of type mismatch like `AsyncCommand` does and added tests.

There is only one thing. `AsyncCommand` throws exception if `canExecute` is null and is typed. I think it shouldn't. If `canExecute` is null (even in constructor for `AsyncCommand<T1, T2>`, `CanExecute` should always return `true` even if argument type doesn't match `T2`).

image

brminnick:
Please re-enable and fix the failing test.

The `InvalidCommandParameterException` is correct: `bool` cannot be `null`.

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/ObjectModel/ICommandTests/CommandFactoryTests/CommandFactory_Tests.cs
```diff
@@ -0,0 +1,223 @@
...
+			// Act
+
+			// Assert
+			Assert.True(command.CanExecute(true));
+			Assert.True(command.CanExecute(null));
+			Assert.True(command.CanExecute(string.Empty));
+			Assert.True(command.CanExecute(0));
+		}
+
+		[Fact(Skip = "AsyncCommand validates type here. Is this correct behavior?")]
```
brminnick, line 178: Do not skip this test.

It is correctly throwing an exception because you are passing in `null` as a parameter that is expecting  a `bool`.

brminnick (changes requested):

samples/XCT.Sample.Android/Xamarin.CommunityToolkit.Sample.Android.csproj
```diff
@@ -127,7 +127,7 @@
   <Import Project="$(MSBuildExtensionsPath)\Xamarin\Android\Xamarin.Android.CSharp.targets" />
   <ProjectExtensions>
     <VisualStudio>
-      <UserProperties XamarinHotReloadXFormsNugetUpgradeInfoBarXamarinCommunityToolkitSampleAndroidHideInfoBar="True" />
+      <UserProperties TriggeredFromHotReload="False" XamarinHotReloadXFormsNugetUpgradeInfoBarXamarinCommunityToolkitSampleAndroidHideInfoBar="True" />
```
pictos, line 130: please revert this change

pictos (changes requested):

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/CommandFactory.shared.cs
brminnick: Here is the implementation for `IAsyncValueCommand`:
```csharp
public static IAsyncValueCommand Create(
	Func<ValueTask> execute,
	Func<bool> canExecute = null,
	Action<Exception> onException = null,
	bool continueOnCapturedContext = false,
	bool allowsMultipleExecutions = true) =>
	new AsyncValueCommand(execute, canExecute, onException, continueOnCapturedContext, allowsMultipleExecutions);

public static IAsyncValueCommand<TExecute> Create<TExecute>(
	Func<TExecute, ValueTask> execute,
	Func<bool> canExecute = null,
	Action<Exception> onException = null,
	bool continueOnCapturedContext = false,
	bool allowsMultipleExecutions = true) =>
	new AsyncValueCommand<TExecute>(execute, canExecute, onException, continueOnCapturedContext, allowsMultipleExecutions);

public static IAsyncValueCommand<TExecute, TCanExecute> Create<TExecute, TCanExecute>(
	Func<TExecute, ValueTask> execute,
	Func<TCanExecute, bool> canExecute = null,
	Action<Exception> onException = null,
	bool continueOnCapturedContext = false,
	bool allowsMultipleExecutions = true) =>
	new AsyncValueCommand<TExecute, TCanExecute>(execute, canExecute, onException, continueOnCapturedContext, allowsMultipleExecutions);
```

maxkoshevoi: Oh, right. It uses different task type and can be differentiated using it. Missed that, thank you. I thought the command itself is value type.

samples/XCT.Sample.Android/Xamarin.CommunityToolkit.Sample.Android.csproj
maxkoshevoi: Reverted. It's added automatically every time I launch Android sample. Is there some reason it cannot be checked into source control?

src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/CommandFactory.shared.cs
maxkoshevoi: Hm, after adding support for `AsyncValueCommand` we have ambiguity when using async lambda as first argument.

image

maxkoshevoi: @brminnick Can we just support `Command` and `AsyncCommand` with `CommandFactory.Create` since those two are most commonly used onse?

maxkoshevoi: `AsyncValueCommand` is almost never used, so removed it from `CommandFactory`.

Here's what I mean by "almost never":
image
image

User can still create it manually if they need it

brminnick: > Can we just support Command and AsyncCommand with CommandFactory.Create since those two are most commonly used onse?

No. We cannot add a support library for `IAsyncCommand` whilst ignoring `IAsyncValueCommand`.

 I will not approve this PR without support for `ValueTask` and `IAsyncValueCommand`.

maxkoshevoi: Do you have some ideas on how to remove the ambiguity with async lambdas without forcing users to write explicit cast? The only one I came up with is to create `AsyncValueCommand` from a different method `CreateValue` or something. But it that feels kinda strange.

brminnick: With all due respect, Maksym, you are author of this PR; not me. I'd like for you to make an attempt, and I'm happy to review it.

I've already [provided you code that provides a starting point for supporting `IAsyncValueCommand`](https://github.com/xamarin/XamarinCommunityToolkit/pull/797#discussion_r564183571). I recommend massaging these provided APIs until you reach a solution.

maxkoshevoi: Thank you again for providing those APIs. 

What do you think about this approach: `Create()` is used for most commonly used `ICommand` types `Command` and `IAsyncCommand` (mainly for automatic command type resolution), and `CreateXXX()` methods are used for other types like `CreateValue()` for `IAsyncValueCommand`.

Committed changes with `CreateValue` and tests for it.

maxkoshevoi: Got it! (some kind people helped) Optional parameters can act as tiebreakers.
Added overloads for `IAsyncCommand Create` methods without optional parameters so they would be preferred over `IAsyncValueCommand Create` that does have optional parameters.

samples/XCT.Sample.Android/Xamarin.CommunityToolkit.Sample.Android.csproj
AlleSchonWeg: FYI: https://developercommunity.visualstudio.com/content/problem/1291161/xamarinandroid-csproj-is-always-shown-as-edited-in.html

brminnick:
I've updated the PR with the following:
- Add XML Documentation
- Separated `Command.Factory.Command.shared.cs`, `Command.Factory.IAsyncCommand.shared.cs` and `Command.Factory.IAsyncValueCommand.shared.cs` for easier legibility and maintenance 
- Added missing `CommandFactory.Create` methods
- Updated Unit Tests

brminnick (approved):

brminnick:
@maxkoshevoi Please open a Pull Request on the [Xamarin Community Toolkit Docs Repo] adding the documentation for `CommandFactory`: https://github.com/MicrosoftDocs/xamarin-communitytoolkit/issues/70

PR #812, opened 2021-01-29
Add StatusBarEffect and NavigationBarEffect classes to manage their color and style

### LINKED ISSUE #708: [Enhancement] Add BarStyleManager to manage navigation and status bar color
## Summary

Add methods to change status and navigation bar color and appearance (white or dark text color).

## API Changes

Here is the interface that I use:
```cs
namespace MyProject.Core.Models.InterplatformCommunication
{
    public interface IBarStyleManager
    {
        void SetStatusBarColor(string hexColor);

        void SetNavigationBarColor(string hexColor);
    }
}
```

Android implementation (modified [SO answer](https://stackoverflow.com/questions/58441077/xamarin-forms-how-to-change-the-status-bar-color-without-a-navigation-page/)):
<details><summary>Show/Hide code</summary>

```cs
using Android.Graphics;
using Android.OS;
using Android.Views;
using MyProject.Core.Models.InterplatformCommunication;
using NureTimetable.Droid.Dependences;
using Xamarin.Essentials;
using MyProject.Core.Extensions;

[assembly: Xamarin.Forms.Dependency(typeof(BarStyleManager))]
namespace MyProject.Droid.Dependences
{
    public class BarStyleManager : IBarStyleManager
    {
        public void SetStatusBarColor(string hexColor)
        {
            if (Build.VERSION.SdkInt < BuildVersionCodes.M)
            {
                return;
            }

            Color color = Color.ParseColor(hexColor);
            MainThread.BeginInvokeOnMainThread(() =>
            {
                var currentWindow = GetCurrentWindow();
                SetBarAppearance(currentWindow, color == Color.White, null);
                currentWindow.SetStatusBarColor(color);
            });
        }

        public void SetNavigationBarColor(string hexColor)
        {
            if (Build.VERSION.SdkInt < BuildVersionCodes.M)
            {
                return;
            }

            Color color = Color.ParseColor(hexColor);
            MainThread.BeginInvokeOnMainThread(() =>
            {
                var currentWindow = GetCurrentWindow();
                SetBarAppearance(currentWindow, null, color == Color.White);
                currentWindow.SetNavigationBarColor(color);
            });
        }

        private static void SetBarAppearance(Window currentWindow, bool? statusBarLight = null, bool? navigationBarLight = null)
        {
            StatusBarVisibility barAppearanceOld = 0;
            WindowInsetsControllerAppearance barAppearanceNew = 0;
            if ((int)Build.VERSION.SdkInt < 30)
            {
#pragma warning disable CS0618 // Type or member is obsolete. Using new API for Sdk 30+
                barAppearanceOld = currentWindow.DecorView.SystemUiVisibility;
#pragma warning restore CS0618 // Type or member is obsolete
            }
            else
            {
                barAppearanceNew = (WindowInsetsControllerAppearance)currentWindow.InsetsController.SystemBarsAppearance;
            }

            if (statusBarLight == true)
            {
                barAppearanceOld = barAppearanceOld.AddFlag((StatusBarVisibility)SystemUiFlags.LightStatusBar);
                barAppearanceNew = barAppearanceNew.AddFlag(WindowInsetsControllerAppearance.LightStatusBars);
            }
            else if (statusBarLight == false)
            {
                barAppearanceOld = barAppearanceOld.RemoveFlag((StatusBarVisibility)SystemUiFlags.LightStatusBar);
                barAppearanceNew = barAppearanceNew.RemoveFlag(WindowInsetsControllerAppearance.LightStatusBars);
            }
            if (navigationBarLight == true)
            {
                barAppearanceOld = barAppearanceOld.AddFlag((StatusBarVisibility)SystemUiFlags.LightNavigationBar);
                barAppearanceNew = barAppearanceNew.AddFlag(WindowInsetsControllerAppearance.LightNavigationBars);
            }
            else if (navigationBarLight == false)
            {
                barAppearanceOld = barAppearanceOld.RemoveFlag((StatusBarVisibility)SystemUiFlags.LightNavigationBar);
                barAppearanceNew = barAppearanceNew.RemoveFlag(WindowInsetsControllerAppearance.LightNavigationBars);
            }
            
            if ((int)Build.VERSION.SdkInt < 30)
            {
#pragma warning disable CS0618 // Type or member is obsolete. Using new API for Sdk 30+
                currentWindow.DecorView.SystemUiVisibility = barAppearanceOld;
#pragma warning restore CS0618 // Type or member is obsolete
            }
            else
            {
                currentWindow.InsetsController?.SetSystemBarsAppearance((int)barAppearanceNew, (int)barAppearanceNew);
            }
        }

        private Window GetCurrentWindow()
        {
            Window window = Platform.CurrentActivity.Window;
            window.ClearFlags(WindowManagerFlags.TranslucentStatus);
            window.AddFlags(WindowManagerFlags.DrawsSystemBarBackgrounds);
            return window;
        }
    }
}
```

</details>

## Intended Use Case
This will be used to more easily customize theme of the application

## Who Will Do The Work?

- [x] I am willing to take this on myself (but only for Android. I have no experience in iOS development and don't have Mac to do it)

maxkoshevoi:
Port of this PR https://github.com/xamarin/Xamarin.Forms/pull/8298

#### Important

Android implementation will [need to be updated](https://github.com/xamarin/XamarinCommunityToolkit/pull/812#discussion_r575855668) when XCT will move to Mono11+

### Description of Change ###

- Added `StatusBarEffect` for StatusBar style manipulation.
- Added `NavigationBarEffect` for NavigationBar style manipulation (Android only).

### Bugs Fixed ###

- Fixes #708
- Fixes https://github.com/xamarin/Xamarin.Forms/issues/7314

### API Changes ###

```csharp
public class StatusBarEffect
{
    public Color Color { set; }
    public StatusBarStyle Style { set; }
}

public class NavigationBarEffect
{
    public Color Color { set; }
    public NavigationBarStyle Style { set; }
}
```

### Testing ###
<details>

- Android 10 and 11:
Screenshot_1616197836

- UWP
image

- iOS
image

</details>

### PR Checklist ###


src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/BarStyle/PlatformBarStyle.android.cs
```diff
@@ -0,0 +1,164 @@
...
+				currentWindow.DecorView.SystemUiVisibility = appearanceLegacy;
+#pragma warning restore CS0618 // Type or member is obsolete
+			}
+			else
+			{
+				currentWindow.InsetsController.SetSystemBarsAppearance((int)appearance, (int)appearance);
+			}
+		}
+
+		static FormsAppCompatActivity Activity => App.Context.GetActivity() as FormsAppCompatActivity;
```
maxkoshevoi, line 162: Could someone please suggest how to het the activity here?
This doesn't work

pictos: Since you are inside the `Effect` you can grab the Context from the `Control` or `Container` property. Can you try it? If that doesn't solve your trouble our implementation on PopUp PR will solve.

maxkoshevoi: Unfortunately both `Container.Context.GetActivity()` and `Control.Context.GetActivity()` are `null`. Waiting on PopUp PR then

pictos: I think it's fine to cast the `Control.Context` to Activity.

Cfun1: It happened to me some strange behavior, not taking modifications into account when debugging with breakpoint on it, could you try to clean and rebuild `Container.Context.GetActivity()` should work.

pictos: @maxkoshevoi I fixed this issue. Now the effect is working

pictos:
@maxkoshevoi can you drop the NavigationBar implementation for this PR? Since it's just android we should work on the PlatformSpecific API. After this PR we can spec that, I already have something implemented (in a branch that I need to find) that we can use for that purpose.

maxkoshevoi:
@pictos Sorry for long delay, was on vacation.
- Removed `NavigationBar` from the PR. Where should I include it then? Changing `NavigationBar` color on Android is is as valid as changing `StatusBar` color (what interesting is that `NavigationBar` style changes automatically according to color)
- Renamed effect to `Window` to reflect that it affects global window of the application (I was also thinking about `Page`, but this might mean that color will revert to default one once user leaves the page, which is currently not the case)
- Made some other minor changes and tested the effect on Android 10 and 11 (since they have different APIs for `StatusBar` style). Everything works

samples/XCT.Sample.Android/Xamarin.CommunityToolkit.Sample.Android.csproj
```diff
@@ -16,7 +16,7 @@
     <AndroidManifest>Properties\AndroidManifest.xml</AndroidManifest>
     <MonoAndroidResourcePrefix>Resources</MonoAndroidResourcePrefix>
     <MonoAndroidAssetsPrefix>Assets</MonoAndroidAssetsPrefix>
-    <TargetFrameworkVersion>v10.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v11.0</TargetFrameworkVersion>
```
pictos, line 19: can you revert this change? This is causing the build error on CI.

samples/XCT.Sample.Android/Properties/AndroidManifest.xml
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="1" android:versionName="1.0" package="com.companyname.xamarincommunitytoolkitsample" android:installLocation="auto">
-	<uses-sdk android:minSdkVersion="21" android:targetSdkVersion="29" />
+	<uses-sdk android:minSdkVersion="21" android:targetSdkVersion="30" />
```
pictos, line 3: Can you revert this change?

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Window/PlatformWindow.android.cs
```diff
@@ -0,0 +1,129 @@
...
+
+[assembly: ExportEffect(typeof(PlatformWindow), nameof(Effects.Window))]
+
+namespace Xamarin.CommunityToolkit.Android.Effects
+{
+	public class PlatformWindow : PlatformEffect
+	{
+		protected override void OnAttached()
+		{
+			if (Effects::Window.GetStatusBarColor(Element) != (Forms.Color)Effects::Window.StatusBarColorProperty.DefaultValue)
```
pictos, line 21: Can we avoid calling the full namespace here and in other parts?

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Window/Window.shared.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System.Linq;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Effects
+{
+	public class Window : RoutingEffect
```
pictos, line 6: Can you rename this class and file adding the suffix `Effect` to follow our pattern?
```suggestion
	public class WindowEffect : RoutingEffect
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="MSBuild.Sdk.Extras/2.1.2">
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;netstandard2.0;netstandard2.1;Xamarin.iOS10;MonoAndroid10.0;Xamarin.TVOS10;Xamarin.WatchOS10;Xamarin.Mac20;tizen40</TargetFrameworks>
+    <TargetFrameworks>netstandard1.0;netstandard2.0;netstandard2.1;Xamarin.iOS10;MonoAndroid11.0;Xamarin.TVOS10;Xamarin.WatchOS10;Xamarin.Mac20;tizen40</TargetFrameworks>
```
pictos, line 4: Please, remove this MonoAndroid11 support

pictos (changes requested):
I have more change requests. Please avoid pushing changes in `csproj` and manifest/info.plist if they aren't needed.

src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj
maxkoshevoi: I moved to Android 11's ADK because it has different API to set status bar style. If I revert this change than I would need to revert support for Android 11 API for changing it.

It should still work on Android 11 (I'll check that), but this code will need to be added later when XCT will move to Android 11+.

samples/XCT.Sample.Android/Xamarin.CommunityToolkit.Sample.Android.csproj
maxkoshevoi: Is it possible to make CI support Android 11?

src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj
pictos: Yeah, but we need to discuss internally when we will move to support Android11 and that will be a hard job, because we will need to test everything and see if it works. So for now I prefer to drop the Android 11 support

pictos:
@maxkoshevoi
> @pictos Sorry for long delay, was on vacation.

> Removed NavigationBar from the PR. Where should I include it then? Changing NavigationBar color on Android is is as valid as changing StatusBar color (what interesting is that NavigationBar style changes automatically according to color)
Renamed effect to Window to reflect that it affects global window of the application (I was also thinking about Page, but this might mean that color will revert to default one once user leaves the page, which is currently not the case)
Made some other minor changes and tested the effect on Android 10 and 11 (since they have different APIs for StatusBar style). Everything works

No need to be sorry, I really hope that you had a great vacation❣Thanks for pushing this forward

samples/XCT.Sample.Android/Xamarin.CommunityToolkit.Sample.Android.csproj
pictos: Not sure about it.

pictos: For now, let's revert it. Upgrading an API needs more work to guarantee that everything works as expected and changing deprecated APIs

src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj
maxkoshevoi: @pictos Removed Android 11 support checked the code on Android 11 device. It doesn't work. If we cannot update XCT to Android 11 SDK, we cannot implement this feature (who would use it if it doesn't work on latest OS version?). 
I think we should be fine updating to Android 11 SDK (everything should still work).

maxkoshevoi: For history: here is the commit needed to support Android 11's API https://github.com/xamarin/XamarinCommunityToolkit/pull/812/commits/38e872a1c54c4956ef0bc62edbb423059eb5d426

maxkoshevoi:
@pictos Fixed you comments and added `NavigationBar` style modification logic as a separate effect `WindowEffectAndroid` (can rename it to `WindowAndroidEffect`). I tried to include it as `WindowEffect.Android.NavigationBarXxx`, but couldn't figure out how to do it.


<details><summary>Here are test results on Android 11 device (Pixel 4)</summary>

image

</details>

pictos:
@maxkoshevoi that WindowAndroidEffect should be a platform specific API, as I mentioned before we don't have it yet. I'll work on it this week and let you know when we have that merged.

maxkoshevoi:
@pictos Will it be possible to use this API through XAML (something like this: 
`<Setter Property="xct:WindowAndroidEffect.NavigationBarColor" Value="{StaticResource PrimaryColor}"/>`)?

pictos:
@maxkoshevoi we will be able to use it from xaml, will be something using [this API ](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/platform/)

maxkoshevoi:
@pictos Great!

I just wanted to include code for `NavigationBar` somewhere, so it wouldn't get left behind. Feel free to use `SetNavigationBarColor` `SetNavigationBarStyle` methods from `PlatformWindowEffectAndroid` in API you are working on.

And another thing. I don't have iOS device and cannot compile XCT with `uap10.0.19041` target (VS cannot find some classes used in `.uwp.cs` files), so could you please find someone who can work on iOS and UWP bits (porting them from https://github.com/xamarin/Xamarin.Forms/pull/8298)? I think that's about it for Android implementation.

pictos:
@maxkoshevoi can you install the UWP SDK? There's no problem if you can't. Ok, we will look to add the iOS and UWP bit ASAP.

maxkoshevoi:
@pictos I installed it (as VS suggested), and I can build uwp sample project, but cannot build XCT itself. Maybe I've installed wrong SDK

maxkoshevoi:
@pictos Yep, I installed wrong SDK.

- Ported UWP and iOS implementations from https://github.com/xamarin/Xamarin.Forms/pull/8298
- Tested UWP implementation. It works, but `StatusBarStyle` is ignored sometimes. Like it wouldn't display black text on black background (but displays white text on white background)

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Window/PlatformWindowEffect.uwp.cs
```diff
@@ -39,12 +35,55 @@ protected override void OnElementPropertyChanged(PropertyChangedEventArgs args)
...
+			var foregroundColor = Color.Default;
+			switch (style)
+			{
+				case StatusBarStyle.LightContent:
+					foregroundColor = Color.White;
+					break;
+				case StatusBarStyle.DarkContent:
+					foregroundColor = Color.Black;
+					break;
+			}
```
Cfun1, line 69:
```suggestion
			foregroundColor = style switch
			{
		     	StatusBarStyle.LightContent => Color.White,
		    	StatusBarStyle.DarkContent => 	Color.Black;
			};
```
Could we use a more compact syntax?

maxkoshevoi: Good point! Didn't noticed it could be used there. Converted that switch to switch expression

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Window/PlatformWindowEffect.android.cs
```diff
@@ -0,0 +1,105 @@
...
+		public void SetStatusBarStyle(StatusBarStyle style)
+		{
+			if (Build.VERSION.SdkInt < BuildVersionCodes.M)
+				return;
+
+			switch (style)
+			{
+				case StatusBarStyle.Default:
+				case StatusBarStyle.LightContent:
+					RemoveBarAppearanceFlag(Activity, (StatusBarVisibility)SystemUiFlags.LightStatusBar);
```
InquisitorJax, line 63: is StatusBarVisibility deprecated for >= API 30?
https://medium.com/swlh/modifying-system-ui-visibility-in-android-11-e66a4128898b
https://stackoverflow.com/questions/63947770/how-can-i-make-the-status-bar-white-with-black-icons-android-11

maxkoshevoi: It is, but since XCT is targeting Android 10, we cannot use Android 11s SDK yet, also because we are targeting Android 10, this logic works on Android 11 devices (tested this).

We had a discussion about this with @pictos [here](https://github.com/xamarin/XamarinCommunityToolkit/pull/812#discussion_r575855668).

maxkoshevoi:
@pictos Did you by any chance had time to work on the platform specific API or test iOS implementation last week?

pictos:
@maxkoshevoi yes you can see the discussion and work here #949

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Window/PlatformWindowEffect.android.cs
maxkoshevoi: Added note about this in PR description

maxkoshevoi:
@jfversluis, @AndreiMisiukevich, @brminnick Could you please take a look at this?

- Everything is tested and ready for review, except iOS implementation is not tested, but it should work since it's ported from xamarin/Xamarin.Forms#8298. If someone has the ability to test sample page on iOS, could you please do this?
- Not sure where to put sample page, so it's under `Effects` for now. We may need to create new category for `StatusBar` and `NavigationBar` classes, like `API` or something.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/AndroidSpecific/NavigationBar.android.cs
```diff
@@ -0,0 +1,36 @@
...
+		{
+			if (Build.VERSION.SdkInt < BuildVersionCodes.M)
+				return;
+
+			switch (style)
+			{
+				case NavigationBarStyle.DarkContent:
+					BarStyle.AddBarAppearanceFlag((StatusBarVisibility)SystemUiFlags.LightNavigationBar);
+					break;
+				case NavigationBarStyle.LightContent:
```
pictos, line 29: We don't need this case here.
```suggestion
```

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/AndroidSpecific/NavigationBar.android.cs
```diff
@@ -0,0 +1,36 @@
...
+using Xamarin.Forms.Platform.Android;
+
+namespace Xamarin.CommunityToolkit.PlatformConfiguration.AndroidSpecific
+{
+	public static partial class NavigationBar
+	{
+		static partial void SetColor(Color color)
+		{
+			if (Build.VERSION.SdkInt < BuildVersionCodes.M)
+				return;
```
pictos, line 14: How about adding a log message that this feature isn't supported in older versions? I believe that will improve the dev xp.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.android.cs
```diff
@@ -0,0 +1,36 @@
...
+		{
+			if (Build.VERSION.SdkInt < BuildVersionCodes.M)
+				return;
+
+			switch (style)
+			{
+				case StatusBarStyle.DarkContent:
+					BarStyle.AddBarAppearanceFlag((StatusBarVisibility)SystemUiFlags.LightStatusBar);
+					break;
+				case StatusBarStyle.LightContent:
```
pictos, line 29:
```suggestion
```

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.android.cs
```diff
@@ -0,0 +1,36 @@
...
+using Xamarin.Forms.Platform.Android;
+
+namespace Xamarin.CommunityToolkit.PlatformConfiguration.Multiplatform
+{
+	public static partial class StatusBar
+	{
+		static partial void PlatformSetColor(Color color)
+		{
+			if (Build.VERSION.SdkInt < BuildVersionCodes.M)
+				return;
```
pictos, line 13: How about adding a log message that this feature isn't supported in older versions? I believe that will improve the dev xp.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.ios.cs
```diff
@@ -0,0 +1,64 @@
...
+	{
+		static partial void PlatformSetColor(Color color)
+		{
+			var uiColor = color.ToUIColor();
+
+			if (UIDevice.CurrentDevice.CheckSystemVersion(13, 0))
+			{
+				foreach (var window in UIApplication.SharedApplication.Windows)
+				{
+					const int statusBarTag = 38482;
```
pictos, line 19: can we move this const to be outside the `foreach`?

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.ios.cs
```diff
@@ -0,0 +1,64 @@
...
+
+			GetCurrentViewController().SetNeedsStatusBarAppearanceUpdate();
+		}
+
+		static UIViewController GetCurrentViewController()
+		{
+			var window = UIApplication.SharedApplication.KeyWindow;
+
+			var vc = window.RootViewController ?? throw new NullReferenceException();
+			while (vc.PresentedViewController != null)
```
pictos, line 57: can we be in a deadlock here? Just imagining some edge case

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.shared.cs
```diff
@@ -0,0 +1,53 @@
...
+		/// Sets the color of application's status bar
+		/// </summary>
+		public static readonly BindableProperty ColorProperty = BindableProperty.Create(
+			"Color", typeof(Color), typeof(StatusBar), Color.Default, propertyChanged: (_, _, newValue) => SetColor((Color)newValue));
+
+		/// <summary>
+		/// Sets style of application's status bar
+		/// </summary>
+		public static readonly BindableProperty StyleProperty = BindableProperty.Create(
+			"Style", typeof(StatusBarStyle), typeof(StatusBar), StatusBarStyle.Default, propertyChanged: (_, _, newValue) => SetStyle((StatusBarStyle)newValue));
```
pictos, line 18: We don't use an inline delegate, can you extract it to a method?

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.shared.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Xamarin.CommunityToolkit.PlatformConfiguration.Multiplatform
+{
+	public static partial class StatusBar
+	{
+		/// <summary>
+		/// Sets the color of application's status bar
+		/// </summary>
+		public static readonly BindableProperty ColorProperty = BindableProperty.Create(
+			"Color", typeof(Color), typeof(StatusBar), Color.Default, propertyChanged: (_, _, newValue) => SetColor((Color)newValue));
```
pictos, line 12: We don't use an inline delegate, can you extract it to a method?

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.shared.cs
```diff
@@ -0,0 +1,53 @@
...
+		/// </summary>
+		/// <param name="style">Style to set</param>
+		public static void SetStyle(StatusBarStyle style) => PlatformSetStyle(style);
+
+		static partial void PlatformSetColor(Color color);
+
+		static partial void PlatformSetStyle(StatusBarStyle style);
+	}
+
+	public enum StatusBarStyle
```
pictos, line 47: since this is public, what do you think to move this enum (and others) to a separated file?

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.uwp.cs
```diff
@@ -0,0 +1,54 @@
...
+				StatusBarStyle.DarkContent => Color.Black.ToWindowsColor(),
+				_ => Color.Default.ToWindowsColor(),
+			};
+
+			UpdateStatusBar(
+				sb => sb.ForegroundColor = foregroundColor,
+				tb => tb.ForegroundColor = foregroundColor);
+		}
+
+		static void UpdateStatusBar(Action<ViewManagement::StatusBar> updateStatusBar, Action<ApplicationViewTitleBar> updateTitleBar)
```
pictos, line 34: Do these arguments need to be an `Action`? It's very simple to add this overhead.

src/CommunityToolkit/Xamarin.CommunityToolkit/Startup/ToolkitPlatform.android.cs
```diff
@@ -17,8 +17,16 @@ internal static Context Context
...
-				return renderer.View.Context ?? throw new NullReferenceException($"{nameof(Context)} cannot be null");
+				if (page != null)
+				{
+					var renderer = page.GetRenderer();
+					return renderer.View.Context ?? throw new NullReferenceException($"{nameof(Context)} cannot be null");
+				}
+
+				// If MainPage isn't set yet, this is the only way to get Context
+#pragma warning disable CS0618 // Type or member is obsolete
+				return Forms.Forms.Context ?? throw new NullReferenceException($"{nameof(Context)} cannot be null");
```
pictos, line 28: Can you revert this code? We don't want to use the obsolete member. If we don't have a MainPage it's because something is wrong.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/RangeSlider/ThumbFrame.shared.cs
```diff
@@ -10,7 +9,7 @@ public ThumbFrame()
 			#region Required work-around to prevent linker from removing the platform-specific implementation
 #if __ANDROID__
 			if (System.DateTime.Now.Ticks < 0)
-				_ = new ThumbFrameRenderer(ToolkitPlatform.Context ?? throw new NullReferenceException());
+				_ = new ThumbFrameRenderer(ToolkitPlatform.Context);
```
pictos, line 12: can you revert this change? Please avoid doing changes that are not related to the issue.
OR is it a rebase issue?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/SideMenuView/SideMenuView.shared.cs
```diff
@@ -85,7 +85,7 @@ public SideMenuView()
 			#region Required work-around to prevent linker from removing the platform-specific implementation
 #if __ANDROID__
 			if (System.DateTime.Now.Ticks < 0)
-				_ = new Xamarin.CommunityToolkit.Android.UI.Views.SideMenuViewRenderer(ToolkitPlatform.Context ?? throw new NullReferenceException());
+				_ = new Xamarin.CommunityToolkit.Android.UI.Views.SideMenuViewRenderer(ToolkitPlatform.Context);
```
pictos, line 88: same here

src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj
```diff
@@ -97,6 +97,9 @@
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
     <PackageReference Include="System.Reflection.Emit.Lightweight" Version="4.7.0" />
     <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform" Version="6.2.10" />
+    <SDKReference Include="WindowsMobile, Version=10.0.19041.0">
```
pictos, line 100: Do we need this?

pictos (changes requested):
Thanks for this PR, I have some requests

brminnick:
Has this feature request been approved? 

#708 still has the `needs approval` tag.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.shared.cs
maxkoshevoi: I was using your PR as a guideline. Does you comment mean that `PopoverArrowDirection` also needs to be moved to a separate file?

https://github.com/xamarin/XamarinCommunityToolkit/pull/949/files#diff-9fdc661ce4c965ee9871e89532420732f4d0cf2c1d8be607a17d88ee35ce1aca

src/CommunityToolkit/Xamarin.CommunityToolkit/Startup/ToolkitPlatform.android.cs
maxkoshevoi: If I revert this code, our sample app will crash (this code is needed for this PR to work properly). 

That thing is that this line: `MainPage = new BaseNavigationPage(new WelcomePage());`
is not an atomic operations. First page is created, and only then it is set to `MainPage`. Issue is that if `NavigationBar` or `StatusBar` is applied to the `WelcomePage` (or to all pages through styles, like in the sample app), `MainPage` isn't set **yet** when they are evaluated.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/RangeSlider/ThumbFrame.shared.cs
maxkoshevoi: I've updated `ToolkitPlatform.Context` and noticed that some calls to it treat it as nullable, while it's not.
Reverted this change

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/SideMenuView/SideMenuView.shared.cs
maxkoshevoi: Reverted

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.uwp.cs
maxkoshevoi: Yes. They need to be actions since we update `StatusBar` and `ApplicationViewTitleBar` in different ways from different methods.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.shared.cs
maxkoshevoi: Is there some downsides of using lambdas here? I just don't want to create two private methods that are only used to call another method with one of the arguments, and which are only used once.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.android.cs
maxkoshevoi: Like `Debug.WriteLine`, or `Trace.TraceInformation`, or something different?

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.ios.cs
maxkoshevoi: Do you mean infinite loop? Not sure since I'm not familiar with iOS development.
I found this on SO: https://stackoverflow.com/a/61642583/4312132
From what I understand it `RootViewController` returns controller that is currently presenting something. Not sure, why do we need a loop there.

Could you please debug this code and see what does `RootViewController` return and if we go inside the loop at all?

src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj
maxkoshevoi: Yes. Without it, our project wouldn't compile. I think, it's needed to change status bar color on mobile devices in UWP applications.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.android.cs
maxkoshevoi: This is there for readability. It's not obvious, what values are covered by `default` case

src/CommunityToolkit/Xamarin.CommunityToolkit/Startup/ToolkitPlatform.android.cs
pictos: We'll need to find another solution then since we want to avoid using Obsolete members.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.android.cs
pictos: There's a Log class in `Xamarin.Forms.Internals` that we can use.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.android.cs
pictos: Any value that isn't `DarkContent`

src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj
pictos: But there's no Windows Mobile (RIP) we need to review it and make sure that we aren't missing anything

maxkoshevoi: Should I remove code related to Windows Mobile?

src/CommunityToolkit/Xamarin.CommunityToolkit/Startup/ToolkitPlatform.android.cs
maxkoshevoi: I didn't find any. Except this one, or adding `Init()` method to XCT

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.android.cs
maxkoshevoi: Searched through the project. We use `Debug.WriteLine` (`Trace` or `Log`) are never used. So added logs as `Debug.WriteLine`

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.android.cs
maxkoshevoi: Yes, but that's not future proof. What if we add `Auto` value, for example? `LightContent` logic will always be used for `LightContent` and `Default` values, but not for anything else.

src/CommunityToolkit/Xamarin.CommunityToolkit/Startup/ToolkitPlatform.android.cs
pictos: I need to dig into this, but we can listen for the "Renderer" property in Page control and do the action when this property changes

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.android.cs
pictos: If we need to add we can refactor this late, premature optimization isn't good

src/CommunityToolkit/Xamarin.CommunityToolkit/Startup/ToolkitPlatform.android.cs
maxkoshevoi: The thing is that we don't have the page, since we can do `StatusBar.Color = Color.Black;`. So the only page `StatusBar` can get reliable access to is the `MainPage`. 

We can try to fetch page from element to which bindable property is attached to, but as you can see, user can change the style without using bindable property an all.

src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj
maxkoshevoi: Here is type `WindowsMobile` SDK is needed for: https://docs.microsoft.com/en-us/uwp/api/windows.ui.viewmanagement.statusbar

src/CommunityToolkit/Xamarin.CommunityToolkit/Startup/ToolkitPlatform.android.cs
maxkoshevoi: Good thing is that this solution is here only for half a year. After XF merges with Xamarin.Essentials, we would be able to use `CurrentActivity` from there.

Or we can add initialization to XCT. XF and Xamarin.Essentials have it, why cannot we? 😄

AndreiMisiukevich: ```ToolkitPlatform.Context``` ?

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/BarStyle.android.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Xamarin.CommunityToolkit.PlatformConfiguration.Multiplatform
+{
+	static class BarStyle
+	{
+		internal static bool IsSupported()
+		{
+			if (Build.VERSION.SdkInt < BuildVersionCodes.M)
+			{
+				Debug.WriteLine($"This functionality is not available. Minimum supported API is {BuildVersionCodes.M}");
```
AndreiMisiukevich, line 15: Don't you mind removing this debug printing?

If I understand correctly, it won't be visible for devs-SDK-consumers anyway, cuz we deliver "release" build via nuget. So, all Debug writes will be ignored.

```internal static bool IsSupported() => Build.VERSION.SdkInt >= BuildVersionCodes.M;``` is sufficent from my perspective

File: src/CommunityToolkit/Xamarin.CommunityToolkit/Startup/ToolkitPlatform.android.cs (hunk 19 above)
Comment (maxkoshevoi) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Startup/ToolkitPlatform.android.cs:28]: @AndreiMisiukevich What do you mean? The issue is that 
- `ToolkitPlatform.Context` is null when `MainPage` is not set yet
- `MainPage` is created and only then it's set
- Methods in this PR can be used during `MainPage` initialization (like in its constructor or with style applied to all pages), and they use `ToolkitPlatform.Context` to get context

I updated `ToolkitPlatform.Context` to use obsolete `Forms.Forms.Context` in case if `MainPage` is null, and @pictos doesn't like that. I'm also not a fan of this approach, but it's the only way to get `Context` in this case that I found.

File: src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/BarStyle.android.cs (hunk 23 above)
Comment (maxkoshevoi) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/BarStyle.android.cs:15]: @pictos [asked](https://github.com/xamarin/XamarinCommunityToolkit/pull/812#discussion_r603644830) to add some logging when this functionality is not supported. Developer would want to know why they changed statusbar color, and nothing happened.

I can use `Trace` or Xamarin's `Log` instead. But they are not yet used anywhere in XCT.

Comment (pictos) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/BarStyle.android.cs:15]: I would say that the Xamarin's log is a better option

Comment (maxkoshevoi) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/BarStyle.android.cs:15]: @pictos I'm a little concerned about why it's in `Xamarin.Forms.Internals` namespace. Feels like things in `Internals` namespace are not intended to be used outside XF assemblies.

On the other hand, why it's public then?

Comment (pictos) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/BarStyle.android.cs:15]: For this situation I think that is ok to use it

Comment (maxkoshevoi) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/BarStyle.android.cs]: Replaced `Debug` with `Log`.
@AndreiMisiukevich Are you ok with this approach?

Comment (AndreiMisiukevich) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/BarStyle.android.cs]: @maxkoshevoi perfect!

File: src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/AndroidSpecific/NavigationBar.shared.cs [hunk 24]
```diff
@@ -0,0 +1,63 @@
...
+		/// <summary>
+		/// Sets the style of application's navigation bar
+		/// </summary>
+		public static readonly BindableProperty StyleProperty = BindableProperty.Create(
+			"Style", typeof(NavigationBarStyle), typeof(TargetElement), NavigationBarStyle.Default, propertyChanged: (_, _, newValue) => SetStyle((NavigationBarStyle)newValue));
+
+		/// <summary>
+		/// Needed for BindableProperty to work. Don't call this method.
+		/// </summary>
+		public static Color GetColor(BindableObject element) => throw new NotImplementedException();
```
pictos, line 25:
```suggestion
		[EditorBrowsable(EditorBrowsableState.Never)]
        public static Color GetColor(BindableObject element) => throw new NotImplementedException();
```

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/AndroidSpecific/NavigationBar.shared.cs
```diff
@@ -0,0 +1,63 @@
...
+
+		/// <summary>
+		/// Needed for BindableProperty to work. Don't call this method.
+		/// </summary>
+		public static NavigationBarStyle GetStyle(BindableObject element) => throw new NotImplementedException();
```
pictos, line 30:
```suggestion
		[EditorBrowsable(EditorBrowsableState.Never)]
        public static NavigationBarStyle GetStyle(BindableObject element) => throw new NotImplementedException();
```

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.android.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Xamarin.CommunityToolkit.PlatformConfiguration.Multiplatform
+{
+	public static partial class StatusBar
+	{
+		static partial void PlatformSetColor(Color color)
+		{
+			if (!BarStyle.IsSupported())
+				return;
+
+			var activity = (FormsAppCompatActivity)ToolkitPlatform.Context.GetActivity();
```
pictos, line 14: If the activity isn't a `FormsAppCompatActivity`? Can we use the `AppCompatActivity`

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.shared.cs
pictos: Yes that file should be moved to a separeted file, we missed that.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.shared.cs
pictos: A method is better to debug than a lambda, and the IDE will do the hard for you in the refactoring.

pictos (changes requested):
Why the StatusBar implementation isn't an `Effect`?  Since it's supported by 3 platforms we should move that to use the `Effect` API. that will prevent facing NRE about the `Android.Context` at least on `StatusBar` implementation.

src/CommunityToolkit/Xamarin.CommunityToolkit/Startup/ToolkitPlatform.android.cs
pictos: We should wait for the `MainPage` to be instantiate to perform those changes.

maxkoshevoi:
> Why the `StatusBar` implementation isn't an `Effect`?

@pictos It's not an effect so it can be set in C# code. For example, color can be changed when some specific event happens on the page, instead of right when the page is opened. Implementing is as `Effect` artificially limits how it can be used.

Second thing is that when page is closed bar style remains preserved, which might be quite a surprise to users who used to effects being tied to the page they are applied on.

Also making `StatusBar` an Effect doesn't solve the issue since `NavigationBar` also needs to know the `Context`.

pictos:
> > Why the `StatusBar` implementation isn't an `Effect`?
> 
> @pictos It's not an effect so it can be set in C# code. For example, color can be changed when some specific event happens on the page, instead of right when the page is opened. Implementing is as `Effect` artificially limits how it can be used.
>
> Second thing is that when page is closed bar style remains preserved, which might be quite a surprise to users who used to effects being tied to the page they are applied on.
> 
> Also making `StatusBar` an Effect doesn't solve the issue since `NavigationBar` also needs to know the `Context`.

Ok, I'm fine with not moving this to an effect, looking into the original code on [Forms repo](https://github.com/xamarin/Xamarin.Forms/pull/8298) the status bar color is controlled by BP on Page control, we can create attached properties to the Page control instead of creating a new CustomPage control.
The only reason that the Android implementation is falling is that the code tries to run too early. I implemented that code in an old project and works fine, said that I'm pretty sure that we can avoid using that obsolete prop from Forms.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/AndroidSpecific/NavigationBar.shared.cs
maxkoshevoi: Turns out this breaks IneliSense for `BindableProperty`. (the same as not including this method at all)

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/AndroidSpecific/NavigationBar.shared.cs
maxkoshevoi: Turns out this breaks IneliSense for `BindableProperty`. (the same as not including this method at all)

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.android.cs
maxkoshevoi: Unfortunately no. `SetStatusBarColor` method belongs directly to `FormsAppCompatActivity`

jfversluis:
Sample app gives an exception even before showing the first page

image


```
  at Xamarin.CommunityToolkit.PlatformConfiguration.Multiplatform.StatusBar.GetCurrentViewController () [0x0000c] in /Users/jfversluis/Projects/xamarin/XamarinCommunityToolkit/src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.ios.cs:56 
  at Xamarin.CommunityToolkit.PlatformConfiguration.Multiplatform.StatusBar.PlatformSetColor (Xamarin.Forms.Color color) [0x000d8] in /Users/jfversluis/Projects/xamarin/XamarinCommunityToolkit/src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.ios.cs:36 
  at Xamarin.CommunityToolkit.PlatformConfiguration.Multiplatform.StatusBar.SetColor (Xamarin.Forms.Color color) [0x00000] in /Users/jfversluis/Projects/xamarin/XamarinCommunityToolkit/src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.shared.cs:38 
  at Xamarin.CommunityToolkit.PlatformConfiguration.Multiplatform.StatusBar.ColorChanged (Xamarin.Forms.BindableObject bindable, System.Object oldValue, System.Object newValue) [0x00000] in /Users/jfversluis/Projects/xamarin/XamarinCommunityToolkit/src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.shared.cs:14 
  at Xamarin.Forms.BindableObject.SetValueActual (Xamarin.Forms.BindableProperty property, Xamarin.Forms.BindableObject+BindablePropertyContext context, System.Object value, System.Boolean currentlyApplying, Xamarin.Forms.Internals.SetValueFlags attributes, System.Boolean silent) [0x00120] in D:\a\1\s\Xamarin.Forms.Core\BindableObject.cs:512 
  at Xamarin.Forms.BindableObject.SetValueCore (Xamarin.Forms.BindableProperty property, System.Object value, Xamarin.Forms.Internals.SetValueFlags attributes, Xamarin.Forms.BindableObject+SetValuePrivateFlags privateAttributes) [0x00173] in D:\a\1\s\Xamarin.Forms.Core\BindableObject.cs:446 
  at Xamarin.Forms.BindableObject.SetValue (Xamarin.Forms.BindableProperty property, System.Object value, System.Boolean fromStyle, System.Boolean checkAccess) [0x0004d] in D:\a\1\s\Xamarin.Forms.Core\BindableObject.cs:374 
  at Xamarin.Forms.BindableObject.SetValue (Xamarin.Forms.BindableProperty property, System.Object value, System.Boolean fromStyle) [0x00000] in D:\a\1\s\Xamarin.Forms.Core\BindableObject.cs:359 
  at Xamarin.Forms.Setter.Apply (Xamarin.Forms.BindableObject target, System.Boolean fromStyle) [0x000fa] in D:\a\1\s\Xamarin.Forms.Core\Setter.cs:85 
  at Xamarin.Forms.Style.ApplyCore (Xamarin.Forms.BindableObject bindable, Xamarin.Forms.Style basedOn) [0x0001e] in D:\a\1\s\Xamarin.Forms.Core\Style.cs:129 
  at Xamarin.Forms.Style.Xamarin.Forms.IStyle.Apply (Xamarin.Forms.BindableObject bindable) [0x0002b] in D:\a\1\s\Xamarin.Forms.Core\Style.cs:91 
  at Xamarin.Forms.MergedStyle.SetStyle (Xamarin.Forms.IStyle implicitStyle, System.Collections.Generic.IList`1[T] classStyles, Xamarin.Forms.IStyle style) [0x000f8] in D:\a\1\s\Xamarin.Forms.Core\MergedStyle.cs:192 
  at Xamarin.Forms.MergedStyle.set_ImplicitStyle (Xamarin.Forms.IStyle value) [0x00000] in D:\a\1\s\Xamarin.Forms.Core\MergedStyle.cs:91 
  at Xamarin.Forms.MergedStyle.OnImplicitStyleChanged () [0x00038] in D:\a\1\s\Xamarin.Forms.Core\MergedStyle.cs:130 
  at Xamarin.Forms.MergedStyle.<RegisterImplicitStyles>b__30_0 (Xamarin.Forms.BindableObject bindable, System.Object oldvalue, System.Object newvalue) [0x00000] in D:\a\1\s\Xamarin.Forms.Core\MergedStyle.cs:146 
  at Xamarin.Forms.BindableObject.SetValueActual (Xamarin.Forms.BindableProperty property, Xamarin.Forms.BindableObject+BindablePropertyContext context, System.Object value, System.Boolean currentlyApplying, Xamarin.Forms.Internals.SetValueFlags attributes, System.Boolean silent) [0x00120] in D:\a\1\s\Xamarin.Forms.Core\BindableObject.cs:512 
  at Xamarin.Forms.BindableObject.SetValueCore (Xamarin.Forms.BindableProperty property, System.Object value, Xamarin.Forms.Internals.SetValueFlags attributes, Xamarin.Forms.BindableObject+SetValuePrivateFlags privateAttributes) [0x00173] in D:\a\1\s\Xamarin.Forms.Core\BindableObject.cs:446 
  at Xamarin.Forms.BindableObject.SetValueCore (Xamarin.Forms.BindableProperty property, System.Object value, Xamarin.Forms.Internals.SetValueFlags attributes) [0x00000] in D:\a\1\s\Xamarin.Forms.Core\BindableObject.cs:383 
  at Xamarin.Forms.Element.OnResourceChanged (Xamarin.Forms.BindableProperty property, System.Object value) [0x00000] in D:\a\1\s\Xamarin.Forms.Core\Element.cs:599 
  at Xamarin.Forms.Element.OnSetDynamicResource (Xamarin.Forms.BindableProperty property, System.String key) [0x00020] in D:\a\1\s\Xamarin.Forms.Core\Element.cs:472 
  at Xamarin.Forms.BindableObject.SetDynamicResource (Xamarin.Forms.BindableProperty property, System.String key, System.Boolean fromStyle) [0x00066] in D:\a\1\s\Xamarin.Forms.Core\BindableObject.cs:346 
  at Xamarin.Forms.BindableObject.SetDynamicResource (Xamarin.Forms.BindableProperty property, System.String key) [0x00000] in D:\a\1\s\Xamarin.Forms.Core\BindableObject.cs:327 
  at Xamarin.Forms.MergedStyle.RegisterImplicitStyles () [0x00044] in D:\a\1\s\Xamarin.Forms.Core\MergedStyle.cs:148 
  at Xamarin.Forms.MergedStyle..ctor (System.Type targetType, Xamarin.Forms.BindableObject target) [0x0001f] in D:\a\1\s\Xamarin.Forms.Core\MergedStyle.cs:31 
  at Xamarin.Forms.NavigableElement..ctor () [0x00011] in D:\a\1\s\Xamarin.Forms.Core\Shell\NavigableElement.cs:24 
  at Xamarin.Forms.VisualElement..ctor () [0x00059] in D:\a\1\s\Xamarin.Forms.Core\VisualElement.cs:367 
  at Xamarin.Forms.Page..ctor () [0x00016] in D:\a\1\s\Xamarin.Forms.Core\Page.cs:60 
  at Xamarin.Forms.TemplatedPage..ctor () <0x10ca7d340 + 0x0004a> in <978ec34c5c9c4c4eb3d73b6da958bcd6>:0 
  at Xamarin.Forms.ContentPage..ctor () <0x10ca7d110 + 0x0004a> in <978ec34c5c9c4c4eb3d73b6da958bcd6>:0 
  at Xamarin.CommunityToolkit.Sample.Pages.BasePage..ctor () [0x00000] in /Users/jfversluis/Projects/xamarin/XamarinCommunityToolkit/samples/XCT.Sample/Pages/Base/BasePage.cs:14 
  at Xamarin.CommunityToolkit.Sample.Pages.WelcomePage..ctor () [0x00000] in /Users/jfversluis/Projects/xamarin/XamarinCommunityToolkit/samples/XCT.Sample/Pages/WelcomePage.xaml.cs:7 
  at Xamarin.CommunityToolkit.Sample.App..ctor () [0x0006f] in /Users/jfversluis/Projects/xamarin/XamarinCommunityToolkit/samples/XCT.Sample/App.xaml.cs:21 
  at Xamarin.CommunityToolkit.Sample.iOS.AppDelegate.FinishedLaunching (UIKit.UIApplication app, Foundation.NSDictionary options) [0x0001b] in /Users/jfversluis/Projects/xamarin/XamarinCommunityToolkit/samples/XCT.Sample.iOS/AppDelegate.cs:21 
  at (wrapper managed-to-native) UIKit.UIApplication.UIApplicationMain(int,string[],intptr,intptr)
  at UIKit.UIApplication.Main (System.String[] args, System.IntPtr principal, System.IntPtr delegate) [0x00005] in /Users/builder/azdo/_work/1/s/xamarin-macios/src/UIKit/UIApplication.cs:86 
  at UIKit.UIApplication.Main (System.String[] args, System.String principalClassName, System.String delegateClassName) [0x0000e] in /Users/builder/azdo/_work/1/s/xamarin-macios/src/UIKit/UIApplication.cs:65 
  at Xamarin.CommunityToolkit.Sample.iOS.Application.Main (System.String[] args) [0x00001] in /Users/jfversluis/Projects/xamarin/XamarinCommunityToolkit/samples/XCT.Sample.iOS/Main.cs:12 
```

jfversluis (changes requested):
See comment above

pictos:
This null reference occurs because of the styles in App.xaml, since the `InitializeComponenents` method runs before the `MainPage = ... ` we don't have those there. I talked with @maxkoshevoi on Discord yesterday and I believe that we should change the API to be able to fix this scenario. Also, I'm not sure if we need to make this xaml friendly.

maxkoshevoi:
@jfversluis Issue was that iOS implementation used API that is deprecated in iOS 13. Updated iOS implementation thanks to @justadaniel. Everything should work now.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/StatusBar/StatusBar.ios.cs
maxkoshevoi: Here's why do we need a loop here: "There is no single "current" view controller in iOS. There could be multiple view controllers on the screen at once (either nested or presented as popovers or modals). If we have modal page open, we need to use it's controller instead of one from the `window`. We need to find the "most-presented" (for lack of a better term) view controller. "

There shouldn't be an infinite loop here since we would exit when top-most controller is found.

src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj
maxkoshevoi: @pictos Should I remove code related to Windows Phone / Windows Mobile?

src/CommunityToolkit/Xamarin.CommunityToolkit/Startup/ToolkitPlatform.android.cs
maxkoshevoi: Discussed offline. Converted them into effects

src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj
maxkoshevoi: Discussed offline. Dropped support for Windows Mobile

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/StatusBar/BarStyle.android.cs
```diff
@@ -0,0 +1,44 @@
...
+		internal static bool IsSupported()
+		{
+			if (Build.VERSION.SdkInt < BuildVersionCodes.M)
+			{
+				Log.Warning(string.Empty, $"This functionality is not available. Minimum supported API is {BuildVersionCodes.M}");
+				return false;
+			}
+
+			return true;
+		}
```
pictos, line 20: The SO version will not change while the app is running, so we can cache the result.

```suggestion
		static bool? isSupported = null;

		internal static bool IsSupported()
		{
			if (isSupported.HasValue)
			{
				ShowWarning();
				return isSupported.Value;
			}

			if (Build.VERSION.SdkInt < BuildVersionCodes.M)
			{
				ShowWarning();
				isSupported = false;
				return false;
			}

			isSupported = true;
			return true;

			static void ShowWarning() => Log.Warning(string.Empty, $"This functionality is not available. Minimum supported API is {BuildVersionCodes.M}");
		}
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/StatusBar/StatusBarEffect.shared.cs
```diff
@@ -0,0 +1,53 @@
...
+			var oldEffect = page.Effects.FirstOrDefault(e => e is StatusBarEffect);
+			if (oldEffect != null)
+				page.Effects.Remove(oldEffect);
+
+			page.Effects.Add(new StatusBarEffect());
+		}
+
+		public Color Color => GetColor(Element);
+
+		public StatusBarStyle Style => GetStyle(Element);
```
pictos, line 48: Let's remove those properties.
```suggestion
```

pictos (changes requested):
On UWP the right part of the status bar doesn't change the color, do you know if you can control that color?
image

I tested all code changes suggestions on my local machine.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/NavigationBar/PlatformNavigationBarEffect.android.cs
```diff
@@ -0,0 +1,73 @@
...
+
+		FormsAppCompatActivity Activity
+		{
+			get
+			{
+				if (Control != null)
+					return (FormsAppCompatActivity)(Control.Context ?? throw new NullReferenceException());
+				else
+					return (FormsAppCompatActivity)(Container.Context ?? throw new NullReferenceException());
+			}
```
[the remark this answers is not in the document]
maxkoshevoi: `FormsAppCompatActivity` is needed for SetColor method. It doesn't exists for regular `Activity`

samples/XCT.Sample.iOS/Info.plist
```diff
@@ -46,5 +46,7 @@
 	<array>
 		<string>audio</string>
 	</array>
+	<key>UIViewControllerBasedStatusBarAppearance</key>
+	<false/>
```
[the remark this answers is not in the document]
maxkoshevoi: I don't think there are docs for this feature yet

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/StatusBar/BarStyle.android.cs
maxkoshevoi: That check is free since `SdkInt` is constant, so it doesn't really matter if you're checking `isSupported.HasValue` or `Build.VERSION.SdkInt < BuildVersionCodes.M`. Caching just complicates the code here.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/StatusBar/StatusBarEffect.shared.cs
maxkoshevoi: This might break something. Let me check

pictos: Didn't break, as I said I tested all those changes on my end.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/NavigationBar/PlatformNavigationBarEffect.android.cs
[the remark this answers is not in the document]
pictos: Not on this one, just on `StatusBarEffect`

samples/XCT.Sample.iOS/Info.plist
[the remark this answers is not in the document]
pictos: I know, but just to remember us to add it there

maxkoshevoi:
> On UWP the right part of the status bar doesn't change the color

Yeah, UWP is a bit strange. Code executes just fine, UWP just doesn't update the text color sometimes. I've described this behavior a bit earlier: https://github.com/xamarin/XamarinCommunityToolkit/pull/812#issuecomment-779498900

pictos:
> > On UWP the right part of the status bar doesn't change the color
> 
> Yeah, UWP is a bit strange. Code executes just fine, UWP just doesn't update the text color sometimes. I've described this behavior a bit earlier: [#812 (comment)](https://github.com/xamarin/XamarinCommunityToolkit/pull/812#issuecomment-779498900)

I'm ok with that limitation, we can add this an enhancement, or maybe some one that knows UWP can help us :D

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/NavigationBar/PlatformNavigationBarEffect.android.cs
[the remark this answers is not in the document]
maxkoshevoi: Good point, here we just need `Activity`

pictos (dismissed):

maxkoshevoi:
@jfversluis @AndreiMisiukevich @brminnick Could you take a look at this please? PR is ready to be merged

jamesmontemagno:
What is the difference between statusbar and navigationbar? NavigationPage and Shell page already have background colors and text colors.

pictos:
@jamesmontemagno
> What is the difference between statusbar and navigationbar? NavigationPage and Shell page already have background colors and text colors.

NavigationBar is just for Android. You can see it in the image below.
image

BTW, feel free to suggest another name that can avoid confusion on devs.

jamesmontemagno:
I would say "BottomNavigationBar" I think "NavigationBar" is just very confusing in the context that that is usually the top of the page.

jamesmontemagno:
maybe it is fine idk, i guess that is the actual android api

maxkoshevoi:
> maybe it is fine idk, i guess that is the actual android api

Yes, that's how it's called in Android API: https://developer.android.com/reference/android/view/Window.html#setNavigationBarColor(int)

jfversluis:
I think I kind of agree with @jamesmontemagno on this one though. I can see how people are going to think this is about the top Xamarin.Forms bar even though that is the name for it on Android.

Since it's only supported by Android should we make that also more clear with: `AndroidNavigationBar`? Or indeed `BottomNavigationBar`? Something like that?

maxkoshevoi:
@jfversluis I vote for `AndroidNavigationBar`.

Cfun1:
If the "NavigationBar" we are talking about in this context is Android specific only, wouldn't that be obvious from the way it is consumed, like the platform specific namespace? if yes then I am with the name "BottomNavigationBar".

maxkoshevoi:
@Cfun1 It's not in the platform specific namespace, so it can be used by just using `xmlns:xct="http://xamarin.com/schemas/2020/toolkit"`.
XCT already has similar effects. For example `SafeArea` is iOS only.

maxkoshevoi:
> `On<Android>.SetNavigationBarColor()` something like that.

This was the way I implemented it initially, but [there was an issue with accessing Android's activity](https://github.com/xamarin/XamarinCommunityToolkit/pull/812#discussion_r603646978), so we decided to convert it to an Effect.

Cfun1:
@maxkoshevoi there was recently a pr related to Android activity but not sure whether it also solves the issue you mentioned or no https://github.com/xamarin/XamarinCommunityToolkit/pull/1351

maxkoshevoi:
@Cfun1 It doesn't. With current implementation, first time when `ToolkitPlatform.Context` can return any result is after `Application.Current.MainPage` is set. This effect needs to access it while `MainPage` is being created (so before it is set).

Cfun1:
Gotcha I understand now.

jfversluis (approved):

kerberosargos:
Hello I think there is a bug in iOS 12.4 and above. I have created report in here https://github.com/xamarin/XamarinCommunityToolkit/issues/1646

PR #815, opened 2021-01-30
CommandFactory fixes for existing overloads

### LINKED ISSUE #798: [Enhancement] Add CommandFactory class

maxkoshevoi:
### Description of Change ###

I found two edge case issues with existing overloads in `CommandFactory`:
1) 
image
This call is failing with "Ambiguous between `Task` and `ValueTask` error", since we don't have 
` IAsyncCommand<TExecute, TCanExecute> Create<TExecute, TCanExecute>(Func<TExecute, Task> executeTask)` overload.
Solution is to either add it or make `canExecute` parameter for `Create<TExecute, TCanExecute>` mandatory. It doesn't make sense to specify type for a parameter and not provide it. I decided to go with second option.
2) 
image
If user provides some argument except `execute` and `canExecute` to `Create` with `execute` being async lambda, call becomes ambiguous. Solution is to either create another 20 overloads to cover all possible parameter combinations or to rename `execute` parameter for `IAsyncCommand` and `IAsyncValueCommand` to have two different names, so that user could do this:
image

### Bugs Fixed ###

- Fixes #798

### API Changes ###

- Made `canExecute` parameter for `Create<TExecute, TCanExecute>` mandatory
- Renamed `execute` parameter for `IAsyncCommand` and `IAsyncValueCommand`

### Behavioral Changes ###

Less ambiguity warnings when calling `CommandFactory.Create`

### PR Checklist ###

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/CommandFactory.Command.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/CommandFactory.Command.shared.cs
@@ -20,6 +20,7 @@ namespace Xamarin.CommunityToolkit.ObjectModel
 		/// Initializes Xamarin.Forms.Command
 		/// </summary>
 		/// <param name="execute">The Function executed when Execute is called. This does not check canExecute before executing and will execute even if canExecute is false</param>
+		/// <param name="canExecute">The Function that verifies whether or not Command should execute.</param>
 		/// <returns>Xamarin.Forms.Command</returns>
 		public static Command Create(Action execute, Func<bool> canExecute) =>
 			new Command(execute, canExecute);
@@ -36,6 +37,7 @@ namespace Xamarin.CommunityToolkit.ObjectModel
 		/// Initializes Xamarin.Forms.Command
 		/// </summary>
 		/// <param name="execute">The Function executed when Execute is called. This does not check canExecute before executing and will execute even if canExecute is false</param>
+		/// <param name="canExecute">The Function that verifies whether or not Command should execute.</param>
 		/// <returns>Xamarin.Forms.Command</returns>
 		public static Command Create(Action<object> execute, Func<object, bool> canExecute) =>
 			new Command(execute, canExecute);
@@ -52,6 +54,7 @@ namespace Xamarin.CommunityToolkit.ObjectModel
 		/// Initializes Xamarin.Forms.Command<typeparamref name="T"/>
 		/// </summary>
 		/// <param name="execute">The Function executed when Execute is called. This does not check canExecute before executing and will execute even if canExecute is false</param>
+		/// <param name="canExecute">The Function that verifies whether or not Command should execute.</param>
 		/// <returns>Xamarin.Forms.Command<typeparamref name="T"/></returns>
 		public static Command<T> Create<T>(Action<T> execute, Func<T, bool> canExecute) =>
 			new Command<T>(execute, canExecute);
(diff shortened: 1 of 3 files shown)


pictos:
@maxkoshevoi first of all thank you so much for your contributions to this project!

I'm feeling that this feature `CommandFactory` is getting very complex. From my point of view, the tradeoff between the complexity and the benefit of this feature doesn't pay. And I'm thinking in a long term as well, with the maintenance of this feature.

```csharp
// using ctor
MyAsyncCommand = new AsyncCommand(MyTask, MyCanExecute);

// with the factory
MyAsyncCommand = CommandFactory.Create(MyTask, MyCanExecute);
```
Here are some points that I want to highlight.

1. If the dev uses the ctor approach he will type less.
2. On Factory implementation there's no additional check or any _plus_ besides using the ctor directly.

Said that I think that we can drop the `CommandFactory` for now. Maybe in the future versions of C#, we can implement this more easily.

maxkoshevoi:
@pictos Happy to help =)

Initially, `CommandFactory` was created with a goal to autoresolve needed command type (more detailed description of this benefit here: #798, #797), but this required to modify `Command` constructors to accept the same parameters as `AsyncCommand` does (except `execute` shouldn't return anything). @brminnick was strongly against it ([here is the reasoning](https://github.com/xamarin/XamarinCommunityToolkit/pull/811#issuecomment-769999331)), so `CommamdFactory` was merged with parameters in `Create` overloads for `Command` matching ones in `Command` constructors, and parameters in `Create` overloads for `AsyncCommand` matching ones in `AsyncCommand` constructors.

This implementation still provides automatic command type resolution, just in less cases:
- If just `execute` argument is provided (which should cover majority of cases) command type would be resolved in pretty much all the cases
- If `canExecute` also provided, `Create` call might look a bit different, but should be the same for most cases

Regarding your comments:
1. Yes, with constructor you type less, but you'll need to type different things depending on `MyTask` return type, and retype command creation if return type changes (from `void` to `Task` or vice versa). With `CommandFactory` you type the same thing once for every command creation and forget about it.
2. Benefits of `CommandFactory` is not in some logic inside it, but overload resolution logic build into C# compiler 

My take
- I would love to have `CommandFactory` with implementation that was originally proposed (that accepts the same arguments (excluding `execute` return type) regardless of command type). But since @brminnick against modifying `Command` constructor parameters, I don't see this happening.. But current implementation should do the job pretty good too.
- **Complexity**: Could you please elaborate on that? Currently `CommandFactory` is implemented as 1 to 1 mapping to command constructors. Not much complexity there. The only issue we faced is when async lambda is passed as `execute` argument, compiler doesn't know whether to cast it to `Task` or `ValueTask`, so we added `Create` overloads without optional parameters to resolve the ambiguity. In this PR I also want to change `execute` parameter name to provide another way to resolve it if needed overload is not found.
- **Maintenance**: This class would need to be modified only if `Command`, `AsyncCommand` or `AsyncValueCommand` add, delete or update some of its constructors. This shouldn't happen often.

brminnick:
Hi Maksym! 

I agree however that you've identified an Issue that a future dev may face when using `CommandFactory`, however, changing the parameter names is not the best solution for the issue you identified. And, there is an easy workaround for the dev to implement, meaning that this Issue is not a blocker to use `CommandFactory`.

I apologize for closing another PR, however, had we discussed your solution before you implemented it, we could have landed upon a better implementation.

Here is the proper way for solving issues like this that will save you time in the future:
1. Open an `Issue` identifying the problem and providing a valid reproduction 
2. Recommend a solution 
3. Allow the community to review the Issue
4. Allow the community to contribute alternative solutions
5. Open a PR implementing the consensus solution

I also agree with @pictos that `CommandFactory` doesn't add much benefit over using the `AsyncCommand` and `AsyncValueCommand` constructors. That being said, we did agree to merge your original PR implementing the `CommandFactory` in Xamarin.CommunityToolkit. That PR introduced thousands of lines of code to our codebase that we will have to maintain going forward. This is what Pedro is referring to when he says that it adds complexity + maintenance to the codebase without adding much benefit over using the `AsyncCommand` and `AsyncValueCommand` constructors.

### Next Steps

@maxkoshevoi - please open an Issue reporting the problem you identified here. 

I will comment on that Issue with the recommended workaround for developers who hit the same Issue, ensuring that it will not be a blocker for them. 

Then, as a community, we can provide recommendations on how to fix it; we may also recommend not fixing it, as the fix will introduce more code, and add more complexity to the code base, and require more maintenance, without adding much benefit.

PR #836, opened 2021-02-04
ValidationBehavior: Added IsNotValid property

AndreiMisiukevich:
### Description of Change ###
Very often it's needed to have the "inversed" result value of validation. (Especially in binding).
For example, if I wish to show a label with an error message if the validator failed.
Currently, I have to use the "inverse bool" converter. Yes, it's a working solution, but why we cannot simplify it? All the more so it's a common case.

### API Changes ###
Added:  
- `bool ValidationBehavior.IsNotvalid { get; set; }`

### Behavioral Changes ###
None

### PR Checklist ###


pictos (commented):
Can we update our sample to add this new feature?

AndreiMisiukevich:
@pictos done

pictos (approved):

PR #846, opened 2021-02-06
Add possibility to anchor SnackBar position above a View

### LINKED ISSUE #476: [Enhancement] SnackBar dismissal enhancements
Hi.

I was really looking forward to the snackbar component as I've tried many of them and they all are lacking in some fundamental aspect.

My use case is really simple; I want to display chat messages as some kind of toasts/snackbars/notifications/whatever to the user if he is currently not on the chat page. The toast position should preferably be configurable (top or bottom) so as not to obscure UI navigation elements and/or the notifications should support 'swipe to dismiss' (or other means of closing them).

Unfortunately the snackbar in XCT can't be dismissed, so I thought I'd add two actions to it: 'Hide' and 'Open'. Unfortunately it seems as XCT's snackbar can display one action only (not sure if it's a bug as it accepts an enumeration of actions).

So is there any way this could be addressed in some way? I'd really like to use it, but unfortunately I can't as the user experience would be bad as the user would have to wait out all of the toasts that currently block him from using the app if I were to use it.

Suggestions:
1) Make the position configurable (top or bottom)
2) Add swipe to dismiss, an 'X' close button or something, to allow hiding them prematurely
3) ACT.UserDialogs animates the toasts in/out via a slide instead of a fade, which I strongly prefer. It would also be more natural to swipe away sliding notifications. Any chance of changing this or at least making it configurable?

Thanks.

Cfun1:
### Description of Change ###
Add possibility to anchor SnackBar position above a View using `SetAnchorView()`.

### Bugs Fixed ###
- Fixes #476 for Android platform.

### Usage ###
Default position t the bottom, no anchor, to be called from Page:
When called from a View it will set that View as a bottom anchor (position above it).
```
await button.DisplaySnackBarAsync(options);
```

### API Changes ###
Added: 
 
- ViewExtension Added.
- All platforms change decoration: 
from `internal void Show(Page sender, SnackBarOptions arguments) `
To  	`internal void Show(VisualElement sender, SnackBarOptions arguments) `
### Platforms implemented
- [x] Android
- [ ] iOS
- [ ] uwp

### PR Checklist ###

- [ ] Updated [documentation](https://github.com/MicrosoftDocs/xamarin-communitytoolkit) (Not yet available for SnackBar))


Cfun1:
Not sure about the other platforms, or should we add `if __ANDROID__`?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.android.cs
```diff
@@ -18,6 +18,11 @@ internal void Show(Page sender, SnackBarOptions arguments)
 		{
 			var view = Platform.GetRenderer(sender).View;
 			var snackBar = AndroidSnackBar.Make(view, arguments.MessageOptions.Message, (int)arguments.Duration.TotalMilliseconds);
+			var anchorView = arguments.AnchorView?.GetRenderer()?.View;
+			if (!(anchorView is null))
```
pictos, line 22: We don't use pattern matching for null check
```suggestion
			if (anchorView != null)
```

Cfun1: Out of curiosity is it a convention or for other reason ? ie: is there a real difference or just the syntax ?

pictos: Well, the Xamarin has this pattern forever. I don't know why they don't change to use the pattern matching

pictos:
@Cfun1 other platforms already have this behavior?

Cfun1:
@pictos I don't know for now how SnackBar is implemented in other platforms need to investigate, https://github.com/xamarin/XamarinCommunityToolkit/issues/476#issuecomment-714466648

Cfun1:
I believe these piece of code from Popup control will be helpful:

https://github.com/xamarin/XamarinCommunityToolkit/blob/c9b606ac6109353db1c0d298b32bdbda2f9fd0a3/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Popup/UWP/PopupRenderer.uwp.cs#L172-L177

https://github.com/xamarin/XamarinCommunityToolkit/blob/c9b606ac6109353db1c0d298b32bdbda2f9fd0a3/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Popup/iOS/PopupRenderer.ios.cs#L167-L169

pictos:
@VladislavAntonyuk any thoughts on this one?

VladislavAntonyuk:
I like the idea and PR, just make sure you implement it for iOS as well.
As for the implementation We need to decide if we want extension method or use AnchorView from options.
I created the extension method for page, because XF had similar approach for prompts and alerts.
As for me I prefer extending Options.

VladislavAntonyuk:
For the other hand we can leave Snackbar as it. And use Popup with text and button.

Cfun1:
@AndreiMisiukevich What do you think ?
For me I prefer also the extension method way, in case we go this way instead of creating another View extension class, we can drop the extension method from Page and add it to VisualElement that way both Page and View types will have the method defined.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.android.cs
```diff
@@ -18,6 +18,11 @@ internal void Show(Page sender, SnackBarOptions arguments)
 		{
 			var view = Platform.GetRenderer(sender).View;
 			var snackBar = AndroidSnackBar.Make(view, arguments.MessageOptions.Message, (int)arguments.Duration.TotalMilliseconds);
+			var anchorView = arguments.AnchorView?.GetRenderer()?.View;
+			if (anchorView != null)
+			{
+				snackBar.SetAnchorView(anchorView);
+			}
```
AndreiMisiukevich, line 25:
```suggestion
				snackBar.SetAnchorView(anchorView);
```

AndreiMisiukevich:
@Cfun1 does it work for iOS as well as @VladislavAntonyuk mentioned? :)

Cfun1:
@AndreiMisiukevich I didn't make the related commit yet as was waiting for community to approve the extension method way, if approved I will move forward and implement it for type View only when targeting iOS, and i will update Sample as well.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/Options/ToastOptions.shared.cs
```diff
@@ -18,6 +18,11 @@ public class ToastOptions
 
 		public static MessageOptions DefaultMessageOptions { get; set; } = new MessageOptions();
 
+		/// <summary>
+		/// Show the SnackBar above AnchorView.
+		/// </summary>
+		public View AnchorView { get; set; }
+
```
Cfun1, line 25:
```suggestion
#if __ANDROID__
		/// <summary>
		/// Show the SnackBar above AnchorView.
		/// </summary>
		public View AnchorView { get; set; }
#endif
```
Tried this but failed (in line AnchorView = Fab,) even that I was targeting Android, any idea why?

VladislavAntonyuk: Because that is shared code. And it doesn't target Android (it targets netstandard).
If you decided to implement extension method you shouldn't add this property at all.

Cfun1:
Could someone test if anchor is working on iOS? and that nothing got broken on other platforms?
Tested on Android it is working.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.android.cs
jfversluis: Hey @AndreiMisiukevich maybe I've been doing it somewhere too, but let's prefer braces for single line ifs as well. See the link in #884 this is also suggested by our current editorconfig file. This was something that I recently learned and I think it makes sense. If you don't agree let's discuss, thanks!

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/Options/ToastOptions.shared.cs
Cfun1: @VladislavAntonyuk  it is not important now because I don't need it, but got me confused when saw it being used in https://github.com/xamarin/XamarinCommunityToolkit/blob/b0ca201f766da86a3d53fdddeccd6ced79319592/src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/TouchEffect.shared.cs#L470 it is shared code also, isn't ?

src/CommunityToolkit/Xamarin.CommunityToolkit/Extensions/ViewExtension .shared.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.Threading.Tasks;
+using Xamarin.CommunityToolkit.UI.Views;
+using Xamarin.CommunityToolkit.UI.Views.Options;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Extensions
+{
+	public static class ViewExtension
```
pictos: Looks like there is a space on the FileName, can you check it?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/Options/ToastOptions.shared.cs
```diff
@@ -18,11 +18,6 @@ public class ToastOptions
 
 		public static MessageOptions DefaultMessageOptions { get; set; } = new MessageOptions();
 
-		/// <summary>
-		/// Show the SnackBar above AnchorView.
-		/// </summary>
-		public View AnchorView { get; set; }
```
pictos: This will be a breaking change, can you mark this as Obsolete for now?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.shared.cs
```diff
@@ -7,7 +7,7 @@ namespace Xamarin.CommunityToolkit.UI.Views
 {
 	class SnackBar
```
pictos, line 8: instead of shared, can you rename this file to `SnackBar.netstandard.cs`?

pictos (changes requested):
Hey there, just a fell comments.

src/CommunityToolkit/Xamarin.CommunityToolkit/Extensions/ViewExtension .shared.cs
Cfun1: Yes, but it was fixed in commit https://github.com/xamarin/XamarinCommunityToolkit/pull/846/commits/6ba8d7f6984e593b671afde4d64b82ea107ef694

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.shared.cs
Cfun1: @pictos if we go this way, then I believe we should rename all the other files (Effects, Views...) with a *.share.cs name, right ?

pictos: If they throw exceptions yes. If that is valid code should have the suffix `shared`

Cfun1: Fixed in https://github.com/xamarin/XamarinCommunityToolkit/pull/846/commits/7ddc949e8bb97cbf0b8265322df42ebd3265ff1c

pictos:
@Cfun1 can you check the build falling our CI, looks like a test issue

Cfun1:
@pictos it was a previous issue in the ci pipeline

pictos:
locally that test passes?

Cfun1:
> locally that test passes? 

@pictos didn't run locally but check my dumb commit succeed when triggered ci again

pictos (approved):
LGTM!

pictos (changes requested):
Just checked on iOS and didn't work
image

Cfun1:
@VladislavAntonyuk I have implemented the Android part successfully, but as @pictos mentioned above iOS and UWP are having issues, I cannot do the work for those platforms (as I don't have the environment setup), if you want you can take this to finish. How to proceed ?

VladislavAntonyuk:
I can try to take a look on weekends.
I will create a new pr, after that this one can be closed.

VladislavAntonyuk:
@pictos As for UWP it is weird. I will check. Probably some changes in Xamarin Forms

pictos:
@VladislavAntonyuk if you want I can merge this PR in another branch and you can work from there.

VladislavAntonyuk:
> @VladislavAntonyuk if you want I can merge this PR in another branch and you can work from there.

It would be great, thank you

pictos:
Merging this branch into the `feature/snackbar-above-view` branch, that way @VladislavAntonyuk can move forward with this implementation. Thanks @Cfun1 for the Android implementation!

pictos (approved):

PR #874, opened 2021-02-09
Make ShieldView more customizable

### LINKED ISSUE #862: [Enhancement] Make shield Subject part BackGroundColor customizable
## Summary

Right now there is no property to set the `BackGroundColor` for the subject part of `ShieldView`. We probably need to customize TextColor separately also. What about font, size, attribues and family? maybe we want to put the subject in Bold 😉.

Idea from @jfversluis https://youtu.be/-02fFAyhRoY?t=193
### Added

<div align="left">

| Property| Description |
| :------------ | :------------ |
| SubjectBackgroundColor| Set the BackgroundColor of the Status part  |
| SubjectTextColor| Set the TextColor of the Status part |
| OnSubjectTextColorChanged()| - |
| OnSubjectBackgroundColorChanged()| - |
| UpdateSubjectTextColor()| - |
| UpdateSubjectBackgroundColor()| - |
</div>

**Renaming**  To be more consistent and avoid devs confusion

<Table>
 
<tr><th>🚨 To be removed in a later version  </th><th>To be removed in a later version</th></tr>
<tr><td>

| Obsolete| Replaced by|
| OnColorChanged()|  OnStatusBackgroundColorChanged()|
| UpdateColor()|  UpdateStatusBackgroundColor()|
| UpdateTextColor()|  UpdateStatusTextColor()| 
| OnTextColorChanged()|  OnStatusTextColorChanged()|
</td><td>

| Obsolete| Replaced by|
| Color|  StatusBackgroundColor|
| TextColor|  StatusTextColor| 
</Table>

## Intended Use Case
Be able to customize Subject and Status parts independently if needed

## Who Will Do The Work?

Cfun1:
### Description of Change ###
Make Shield subject more customizable, details in issue.

### Bugs Fixed ###

closes #862

### API Changes ###
Please refer to linked issue.

### PR Checklist ###

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Shield.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Shield.shared.cs
@@ -48,7 +48,7 @@ namespace Xamarin.CommunityToolkit.UI.Views
 			set => SetValue(SubjectBackgroundColorProperty, value);
 		}
 
-		[Obsolete("Use StatusTextColor instead")]
+		[Obsolete("TextColor is obsolete. Please use StatusTextColor instead")]
 		[EditorBrowsable(EditorBrowsableState.Never)]
 		public static readonly BindableProperty TextColorProperty =
 			BindableProperty.Create(nameof(TextColor), typeof(Color), typeof(Shield), Color.Default,
@@ -98,7 +98,7 @@ namespace Xamarin.CommunityToolkit.UI.Views
 			set => SetValue(StatusProperty, value);
 		}
 
-		[Obsolete("Use StatusBackgroundColor instead")]
+		[Obsolete("Color is obsolete. Please use StatusBackgroundColor instead")]
 		[EditorBrowsable(EditorBrowsableState.Never)]
 		public static readonly BindableProperty ColorProperty =
 			BindableProperty.Create(nameof(Color), typeof(Color), typeof(Shield), Color.Default,
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Shield.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Shield.shared.cs
@@ -40,7 +40,7 @@ namespace Xamarin.CommunityToolkit.UI.Views
 		static void OnSubjectBackgroundColorChanged(BindableObject bindable, object oldValue, object newValue) => ((Shield)bindable).UpdateSubjectColor();
 
 		/// <summary>
-		/// Background <see cref="Forms.Color" /> of the right side of the <see cref="Shield" />. This is a bindable property.
+		/// Background <see cref="Forms.Color" /> of the left side of the <see cref="Shield" />. This is a bindable property.
 		/// </summary>
 		public Color SubjectBackgroundColor
 		{
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Shield.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Shield.shared.cs
@@ -60,7 +60,7 @@ namespace Xamarin.CommunityToolkit.UI.Views
 			set => SetValue(TextColorProperty, value);
 		}
 
-		static void OnTextColorChanged(BindableObject bindable, object oldValue, object newValue) => ((Shield)bindable).UpdateTextColor();
+		static void OnTextColorChanged(BindableObject bindable, object oldValue, object newValue) => ((Shield)bindable).StatusTextColor = (Color)newValue;
 
 		/// <summary>
 		/// Backing BindableProperty for the <see cref="SubjectTextColor"/> property.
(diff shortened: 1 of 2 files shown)


Cfun1:
Remains whether we add the possibility to customize the fond properties separately Subject and Status or no.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Shield.shared.cs
```diff
@@ -29,6 +30,56 @@ public string Subject
...
+		/// Backing BindableProperty for the <see cref="SubjectBackgroundColor"/> property.
+		/// </summary>
+		public static readonly BindableProperty SubjectBackgroundColorProperty =
+			BindableProperty.Create(nameof(SubjectBackgroundColor), typeof(Color), typeof(Shield), Color.Default,
+				propertyChanged: OnSubjectBackgroundColorChanged);
+
+		static void OnSubjectBackgroundColorChanged(BindableObject bindable, object oldValue, object newValue) => ((Shield)bindable).UpdateSubjectColor();
+
+		/// <summary>
+		/// Background <see cref="Forms.Color" /> of the right side of the <see cref="Shield" />. This is a bindable property.
```
jfversluis, line 43:
```suggestion
		/// Background <see cref="Forms.Color" /> of the left side of the <see cref="Shield" />. This is a bindable property.
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Shield.shared.cs
```diff
@@ -92,7 +160,7 @@ public Color TextColor
 		/// <summary>
 		/// Font size of all the text on the <see cref="Shield" />. <see cref="NamedSize" /> values can be used. This is a bindable preoprty.
 		/// </summary>
-		[TypeConverter(typeof(FontSizeConverter))]
+		[Xamarin.Forms.TypeConverter(typeof(FontSizeConverter))]
```
jfversluis, line 163: Why is this needed?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Shield.shared.cs
```diff
@@ -240,10 +308,18 @@ protected override void OnPropertyChanged([CallerMemberName] string propertyName
...
 
+		void UpdateSubjectColor() => ShieldSubjectContainer.BackgroundColor = SubjectBackgroundColor;
+
 		void UpdateColor() => ShieldStatusContainer.BackgroundColor = Color;
 
+		void UpdateStatusBackgroundColor() => ShieldStatusContainer.BackgroundColor = StatusBackgroundColor;
+
+		void UpdateSubjectTextColor() => ShieldSubject.TextColor = SubjectTextColor;
+
 		void UpdateTextColor() => ShieldStatus.TextColor = TextColor;
```
jfversluis, line 319: We can already remove `UpdateTextColor` and let the old `TextColor` property update through `UpdateStatusTextColor`

jfversluis (changes requested):

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Shield.shared.cs
Cfun1: Because without it, there is a namespace ambiguity, (with  System.Runtime.CompilerServices;  if I remember well()

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Shield.shared.cs
Cfun1: I preferred to leave it for now so that the later deletion won't be confusing.
How would `UpdateStatusTextColor()` knows which property has changed (StatusTextColor or TextColor) in order to assign it to ShieldStatus.TextColor ?

jfversluis: > I preferred to leave it for now so that the later deletion won't be confusing.

I think the later deletion will be easier _and_ the code paths doing the same thing will be minimized. What happens when someone now updates the obsolete method but not the other one? We suddenly have a bug in code that we want to delete anyway.

> How would UpdateStatusTextColor() knows which property has changed (StatusTextColor or TextColor) in order to assign it to ShieldStatus.TextColor ?

This is not an issue, right? TextColor only updated the ShieldStatus, so we should just update that one and the behavior will be exactly the same as it is now.

Cfun1: @jfversluis for 2nd point i still see it as an issue since we support both TextColor and StatusTextColor properties could be used by dev to change the same property, unless we change the method definition to take a parameter.

jfversluis: In the [code that is implemented today](https://github.com/xamarin/XamarinCommunityToolkit/blob/ecd677a47678e3502aa4b892fd101e9f6b44bd11/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Shield.shared.cs#L245), whenever you update `TextColor` it only updates `ShieldStatus.TextColor`. We don't want to break that behavior. To deduplicate code-paths we can remove `UpdateTextColor` and let that point to `UpdateStatusTextColor` that you are implementing here. The functionality will be exactly the same as it was before, but we don't need to worry about another codepath.

On the outside everything still looks the same, but on the inside it's easier to maintain for the time that we still support the old API

jfversluis (approved):

jfversluis:
Would you also be able to update the Docs? :)

Cfun1:
@jfversluis sure it is in my mind I will asap.

developer9969:
thank you so much for this one.I really needed the ability to style the subject too, it now makes fully styleable

PR #884, opened 2021-02-11
Code formating preference

Cfun1:
Enable prefer one statement if block, without braces suggestion.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -107,3 +107,6 @@ csharp_new_line_before_catch = true
 csharp_new_line_before_finally = true
 csharp_new_line_before_members_in_object_initializers = true
 csharp_new_line_before_members_in_anonymous_types = true
+
+# Code-block preferences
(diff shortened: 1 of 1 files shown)


pictos (approved):

jfversluis:
Actually I do prefer brackets even for a single line https://stackoverflow.com/questions/359732/why-is-it-considered-a-bad-practice-to-omit-curly-braces

That is what this disables right?

Cfun1:
@jfversluis I do too but it was motivated after my code has been revewed to remove brackets, I am fine to close it.

jfversluis:
Who mentioned that? Then I'll discuss that :)

Thanks for paying attention and then also updating these rules!

Cfun1:
@jfversluis https://github.com/xamarin/XamarinCommunityToolkit/pull/846#discussion_r574033996

Issue #751, opened 2021-01-12
[Bug] [CameraView] Camera media capture (photo) orientation incorrect on Android


UKDeveloper99:
### Description

From what I can tell this looks to be a known issue, I've seen the same code used in other Camera2 implementations. The capture output orientation is incorrect from what I can tell on some devices. You can reproduce this with your camera view Android sample. The media capture is rotated 90 degrees to the right.

I found this stack overflow post that could be helpful. The guy posting it uses the same code as yourselves for GetJpegOrientation in your camera fragment class.

https://stackoverflow.com/questions/38282076/how-to-save-images-with-correct-orientation-using-camera2 

### Steps to Reproduce

1. Capture photo in Android

### Expected Behavior

Captured media orientation should match sensor orientation

### Basic Information

- Version with issue: 1.0.0
- IDE:
- Platform Target Frameworks: 
  - Android: 11
- Affected Devices: Oppo Find X2 Pro

Gelo-V:
This happened in my app too in version 1.0.1 , running on Android 10 device.

AndreiMisiukevich:
Hey folks, thank you very much for the report.
Please, feel free to submit a PRs resolving this issue, if you know what exactly should be changed.

Gelo-V:
Hi @AndreiMisiukevich, cool I'll give it a look during this Sunday by downloading the repo and if I'm able to solve it I'll submit the PR. I have never made any contributions before (but I'm wanting to) so I'll explore it first.

pictos:
@Gelo-V if you need any help, please let us know.

Gelo-V:
Hey @pictos thanks for offering your help (I'll need it jajaja)

I got familiar with the solution and tried to dig into this issue, here it is what I found and concluded:

- In the function `TakePhoto()` of the Camera Fragment script the line `sessionBuilder.Set(CaptureRequest.JpegOrientation, GetJpegOrientation());` had no effect on the orientation of the image in my device. If I understood correctly this is what should rotate the preview image.

-  According to the Android [docs](https://developer.android.com/reference/android/hardware/camera2/CaptureRequest.html#JPEG_ORIENTATION): 

>"For EXTERNAL cameras the sensor orientation will always be set to 0 and the facing will also be set to EXTERNAL. The above code is not relevant in such case.              
>This tag is also used to describe the orientation of the HEIC image capture, in which case the rotation is reflected by EXIF orientation flag, and not by rotating the image data itself."

[...]

PR #886, opened 2021-02-11
Fix camera media rotation android issue

UKDeveloper99:
### Description of Change ###

This should fix issue with incorrectly rotated media capture on some Android devices. As [this](https://medium.com/@kenodoggy/solving-image-rotation-on-android-using-camera2-api-7b3ed3518ab6) article implies. The JPEG_ORIENTATION flag in CaptureRequest is effectively ignored in some devices and the Android docs also imply it will have varying effects across devices. 

See [Android docs](https://developer.android.com/reference/android/hardware/camera2/CaptureRequest.html#JPEG_ORIENTATION)
There are two points about the implementation I would like to elaborate on. Firstly the removal or the JPEG_ORIENTATION flag. I wasn't able to reliably determine (in code) across various devices (some with rotation issue some not) if the addition of the flag had rotated the capture data or modified the Exif info. I don't believe this flag is needed at all now. As we now calculate the correct rotation upon capture. If the JPEG_ORIENTATION flag has modified the capture data by rotation. Then we will be applying another rotation on top of that. If it hasn't then only our rotation will be applied. Either way I can't see an easy way to determine that. Setting any rotation value for this flag on certain devices will have **no effect**.
So with those assumptions, if the only purpose now of the JPEG_ORIENTATION flag is to modify the Exif orientation info. Then we can remove it and set the Exif information using the ExifInterface class when the SaveToFile implementation has been completed.

Secondly, the correct point to apply the capture rotation. I tried several methods of applying the rotation to the capture data (byte array) all with sub optimal performance. E.g. ByteArray -> Bitmap -> CreateBitmapRotateWithMatrix -> ByteArray. Also copying the pixel data to a ByteBuffer and looking at methods to rotate a Jpeg byte array. With the Forms side needing the data in ByteArray format I believe I have provided the best solution by passing the rotation in degrees with the MediaCaptureEventArgs. This allows the ImageView to handle the rotation in a performant way. Citing a post from StackOverflow. It is not the job of the camera view to change the orientation of the capture data (camera2 api) and I believe this to be true.

No samples or test changes required.

### Bugs Fixed ###
https://github.com/xamarin/XamarinCommunityToolkit/issues/751

### API Changes ###

None

Added: 
 
- `double MediaCapturedEventArgs.Rotation { get; }`
- `int CameraFragment.GetRotationCompensation();`

Changed:

 - `void CameraFragment.OnPhoto(object sender, Tuple<string, byte[]> tuple)` => `void CameraFragment.OnPhoto(object sender, Tuple<string, byte[], int> tuple)`

### PR Checklist ###

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs
@@ -343,7 +343,7 @@ namespace Xamarin.CommunityToolkit.UI.Views
 					repeatingIsRunning = false;
 					sessionBuilder.AddTarget(photoReader.Surface);
 					sessionBuilder.Set(CaptureRequest.FlashMode, (int)flashMode);
-					sessionBuilder.Set(CaptureRequest.JpegOrientation, GetJpegOrientation());
+					/*sessionBuilder.Set(CaptureRequest.JpegOrientation, GetJpegOrientation());*/
 					session.Capture(sessionBuilder.Build(), null, null);
 					sessionBuilder.RemoveTarget(photoReader.Surface);
 					UpdateRepeatingRequest();
@@ -355,9 +355,9 @@ namespace Xamarin.CommunityToolkit.UI.Views
 			}
 		}
 
-		void OnPhoto(object sender, Tuple<string, byte[]> tuple) =>
+		void OnPhoto(object sender, Tuple<string, byte[], int> tuple) =>
 			Device.BeginInvokeOnMainThread(() =>
-				Element?.RaiseMediaCaptured(new MediaCapturedEventArgs(tuple.Item1, tuple.Item2)));
+				Element?.RaiseMediaCaptured(new MediaCapturedEventArgs(tuple.Item1, tuple.Item2, tuple.Item3)));
 
 		void OnVideo(object sender, string path) =>
 			Device.BeginInvokeOnMainThread(() =>
@@ -374,7 +374,11 @@ namespace Xamarin.CommunityToolkit.UI.Views
 			{
 				string filePath = null;
 
+				// Calculate image rotation based on sensor and device orientation
+				var rotation = GetRotationCompensation();
+
 				// See TODO on CameraView.SavePhotoToFile
+				// Insert Exif information to jpeg file
 				/*if (Element.SavePhotoToFile)
 				{
 					filePath = ConstructMediaFilename(null, extension: "jpg");
@@ -384,10 +388,30 @@ namespace Xamarin.CommunityToolkit.UI.Views
 				OnPhoto(this, new Tuple<string, byte[]>(filePath, Element.SavePhotoToFile ? null : bytes));*/
 
 				Sound(MediaActionSoundType.ShutterClick);
-				OnPhoto(this, new Tuple<string, byte[]>(filePath, bytes));
+				OnPhoto(this, new Tuple<string, byte[], int>(filePath, bytes, rotation));
 			};
 
 			photoReader.SetOnImageAvailableListener(readerListener, backgroundHandler);
+		}
+
+		private int GetRotationCompensation()
+		{
+			var rotationCompensation = GetDisplayRotationDegrees();
+			var c = Manager.GetCameraCharacteristics(cameraId);
+			// Get the device's sensor orientation.
+			var sensorOrientation = (int)c.Get(CameraCharacteristics.SensorOrientation);
+
+			var facingFront = ((Integer)c.Get(CameraCharacteristics.LensFacing)).IntValue() == (int)LensFacing.Front;
+			if (facingFront)
+			{
+				rotationCompensation = (sensorOrientation + rotationCompensation) % 360;
+			}
+			else
+			{
+				rotationCompensation = (sensorOrientation - rotationCompensation + 360) % 360;
+			}
+
+			return rotationCompensation;
 		}
 
 		void SetupMediaRecorder(Surface previewSurface)
--- a/samples/XCT.Sample/Pages/Views/CameraViewPage.xaml.cs
+++ b/samples/XCT.Sample/Pages/Views/CameraViewPage.xaml.cs
@@ -75,6 +75,7 @@ namespace Xamarin.CommunityToolkit.Sample.Pages.Views
 				case CameraCaptureMode.Default:
 				case CameraCaptureMode.Photo:
 					previewPicture.IsVisible = true;
+					previewPicture.Rotation = e.Rotation;
 					previewPicture.Source = e.Image;
 					doCameraThings.Text = "Snap Picture";
 					break;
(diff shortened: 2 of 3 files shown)


jfversluis:
Wowza! Thanks for this @UKDeveloper99 !

Let me try to understand what this does... The only thing this does is report back what the rotation is, right? So then you should rotate yourself?

UKDeveloper99:
@jfversluis no problem, don't thank me too much, I'll end up getting carried away and adding more features lol.
Yes so it uses a Google recommended (recent/current) func to get the correct rotation for sensor and device orientation then passes that along with the Byte data which the user can use to set a rotation on their image view. Shouldn't effect iOS. It seemed not feasible to rotate the image data in the camera fragment.

jfversluis:
I think we can use some help in the CameraView area so no worries, just make sure to coordinate before taking on too much! :)

I'll have a more in-depth look at this as soon as I have the time. 

Appreciate the effort!

UKDeveloper99:
@jfversluis well I'm relying on this fix in my current app so I thought I'd lend a hand. Glad I could be helpful. How long do you think that might be?

Sure, how would I go about coordinating, do you have a Slack of some sorts or something?

jfversluis:
How would one go about using this to show the image correctly? I don't think we have anything for that in Forms today, right? 

We have a Discord! The DotNetEvolution one, you should be able to join here https://aka.ms/dotnet-discord and find us in the XamarinCommunityToolkit channel :)

UKDeveloper99:
@jfversluis Could you elaborate what you mean by that sorry? I added the usage in to the camera sample page.
Thanks I'll join up now.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs
```diff
@@ -343,7 +343,7 @@ public void TakePhoto()
 					repeatingIsRunning = false;
 					sessionBuilder.AddTarget(photoReader.Surface);
 					sessionBuilder.Set(CaptureRequest.FlashMode, (int)flashMode);
-					sessionBuilder.Set(CaptureRequest.JpegOrientation, GetJpegOrientation());
+					/*sessionBuilder.Set(CaptureRequest.JpegOrientation, GetJpegOrientation());*/
```
jfversluis, line 346: Why keep this around? Let's get it out if it doesn't work!

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs
```diff
@@ -374,7 +374,11 @@ void SetupImageReader()
 			{
 				string filePath = null;
 
+				// Calculate image rotation based on sensor and device orientation
+				var rotation = GetRotationCompensation();
+
 				// See TODO on CameraView.SavePhotoToFile
+				// Insert Exif information to jpeg file
```
jfversluis, line 381: Is this a TODO? Do we know how?

jfversluis (approved):
Couple of small nits. Overall I think we should give this a try! Again, thanks so much!

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs
UKDeveloper99: Well someone might have an idea on how to determine if the capture data has been rotated or just applied to the Exif data. I tried to extract the Exif information without success. But yea I don't think it particularly matters. Hopefully it resolves the issue for the people who have run in to it. Happy I could assist!

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs
UKDeveloper99: Yes it's a TODO - it should be fairly simple using the ExifInterface class, which you just provide a File path to and then you can set the Exif data accordingly as far as I'm aware. E.g. pass the calculated orientation to the Orientation tag. I might have a look at doing something with the SavePhotoToFile functionality if no one is looking at it. Coordinating with yourselves of course.

AmirMahdiNassiri:
Hi guys. Thanks for all your efforts!

Is there a reason preventing us from using the new `GetRotationCompensation` function for `mediaRecorder.SetOrientationHint` as well?

I mean, my front camera on a Samsung Note 10 was capturing videos in this direction ← and I'm aiming for a vertical portrait by the front camera. But when I use the resulting value of the new `GetRotationCompensation` for `mediaRecorder.SetOrientationHint` as well, everything looks fine!

matteovelati:
> @jfversluis no problem, don't thank me too much, I'll end up getting carried away and adding more features lol.
> Yes so it uses a Google recommended (recent/current) func to get the correct rotation for sensor and device orientation then passes that along with the Byte data which the user can use to set a rotation on their image view. Shouldn't effect iOS. It seemed not feasible to rotate the image data in the camera fragment.
> 
> I've tested it with all the devices I have here some with bad rotation some without and it corrects the rotation on all of them. For front and rear cam.

How do you rotate the byte array?

UKDeveloper99:
> 
> 
> > @jfversluis no problem, don't thank me too much, I'll end up getting carried away and adding more features lol.
> > Yes so it uses a Google recommended (recent/current) func to get the correct rotation for sensor and device orientation then passes that along with the Byte data which the user can use to set a rotation on their image view. Shouldn't effect iOS. It seemed not feasible to rotate the image data in the camera fragment.
> > I've tested it with all the devices I have here some with bad rotation some without and it corrects the rotation on all of them. For front and rear cam.
> 
> How do you rotate the byte array?

Load the byte array in to a UI component such as UIImageView or ImageView and supply the rotation to the UI component.

matteovelati:
> > > @jfversluis no problem, don't thank me too much, I'll end up getting carried away and adding more features lol.
> > > Yes so it uses a Google recommended (recent/current) func to get the correct rotation for sensor and device orientation then passes that along with the Byte data which the user can use to set a rotation on their image view. Shouldn't effect iOS. It seemed not feasible to rotate the image data in the camera fragment.
> > > I've tested it with all the devices I have here some with bad rotation some without and it corrects the rotation on all of them. For front and rear cam.
> > 
> > 
> > How do you rotate the byte array?
> 
> Load the byte array in to a UI component such as UIImageView or ImageView and supply the rotation to the UI component.

Yeah but that's only a UI rotation. I'm interested in rotating the image by 90 degrees in order to store it in my online database. Everything I found on the web (SO) is outdated or not working with Xamarin.Forms. Do you have any suggestions?
Thanks

UKDeveloper99:
Can't you just store it in your database as it is with the rotation and then when you come to display it then rotate it in the UI. I did have a look down that path but it also depends on image format (jpg etc). And a few other factors. It's really complex I think you should avoid doing that if you can. What's the purpose of rotating it before it goes in to the online database?

matteovelati:
I need to perform OCR on server side for all the images coming from my cross-platform application, and the mismatch in rotation between pictures taken with Android or iOS phones is a problem for me.

Actually it is a bug that the same framework behaves differently on different platforms, isn't it?

PR #912, opened 2021-02-16
[Breaking Changes] ValidationBehavior Async checks support

### LINKED ISSUE #857: [Enhancement] ValidationBehavior should support async mode.
## API Changes
I propose to replace **protected** method **bool Validate(object value)** by **Task<bool> ValidateAsync(object value)**

## Intended Use Case
Allow to create custom validators that work in async mode (e.g. send requests to the server)

## Who Will Do The Work?

- [X] I am willing to take this on myself

AndreiMisiukevich:
### Description of Change ###
We are eager to allow using ValidationBehavior in async validations. (For example, make a request to the Server Side).

### Bugs Fixed ###
- Fixes #857

### API Changes ###

Added:
- ```public bool ValidationBehavior.IsRunning { get; set; }```
- ```protected Task<bool> ValidateAsync(object value, CancellationToken token);```
- ```protected object Decorate (object value);```

Removed:
- ```protected bool Validate (object value);```
- ```protected object DecorateValue ();```

### Behavioral Changes ###
None

### PR Checklist ###

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/Validators/CharactersValidationBehavior.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/Validators/CharactersValidationBehavior.shared.cs
@@ -65,7 +65,7 @@ namespace Xamarin.CommunityToolkit.Behaviors
 			set => SetValue(MaximumCharacterCountProperty, value);
 		}
 
-		protected override async Task<bool> ValidateAsync(object value, CancellationToken token)
+		protected override async ValueTask<bool> ValidateAsync(object value, CancellationToken token)
 			=> await base.ValidateAsync(value, token) && Validate(value?.ToString());
 
 		static void OnCharacterTypePropertyChanged(BindableObject bindable, object oldValue, object newValue)
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/Validators/CharactersValidationBehavior.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/Validators/CharactersValidationBehavior.shared.cs
@@ -66,7 +66,8 @@ namespace Xamarin.CommunityToolkit.Behaviors
 		}
 
 		protected override async ValueTask<bool> ValidateAsync(object value, CancellationToken token)
-			=> await base.ValidateAsync(value, token) && Validate(value?.ToString());
+			=> await base.ValidateAsync(value, token).ConfigureAwait(false)
+				&& Validate(value?.ToString());
 
 		static void OnCharacterTypePropertyChanged(BindableObject bindable, object oldValue, object newValue)
 		{
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/Validators/CharactersValidationBehavior.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/Validators/CharactersValidationBehavior.shared.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using Xamarin.Forms;
 
 namespace Xamarin.CommunityToolkit.Behaviors
@@ -39,8 +41,8 @@ namespace Xamarin.CommunityToolkit.Behaviors
 			set => SetValue(MaximumCharacterCountProperty, value);
 		}
 
-		protected override bool Validate(object value)
-			=> base.Validate(value) && Validate(value?.ToString());
+		protected override async Task<bool> ValidateAsync(object value, CancellationToken token)
+			=> await base.ValidateAsync(value, token) && Validate(value?.ToString());
 
 		static void OnCharacterTypePropertyChanged(BindableObject bindable, object oldValue, object newValue)
 		{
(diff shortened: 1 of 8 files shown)


src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/Validators/UriValidationBehavior.shared.cs
```diff
@@ -23,8 +25,8 @@ public UriKind UriKind
 			set => SetValue(UriKindProperty, value);
 		}
 
-		protected override bool Validate(object value)
-			=> base.Validate(value)
+		protected override async Task<bool> ValidateAsync(object value, CancellationToken token)
+			=> await base.ValidateAsync(value, token)
```
pictos, line 29: can we just return a `Task.FromResult` here

pictos (commented):
I saw a lot of places that we don't wait, we just return the `Task.FromResult`, said that what do you think to change the return type to `ValueTask`?

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/Validators/CharactersValidationBehavior.shared.cs
```diff
@@ -63,8 +65,8 @@ public int MaximumCharacterCount
 			set => SetValue(MaximumCharacterCountProperty, value);
 		}
 
-		protected override bool Validate(object value)
-			=> base.Validate(value) && Validate(value?.ToString());
+		protected override async Task<bool> ValidateAsync(object value, CancellationToken token)
+			=> await base.ValidateAsync(value, token) && Validate(value?.ToString());
```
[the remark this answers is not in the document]
AndreiMisiukevich: I think, no, cuz we should await the result from the base class.

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/Validators/UriValidationBehavior.shared.cs
AndreiMisiukevich: I think, no, cuz we should await the result from the base class.

pictos: But we await the `ValidateAsync` somewhere, no?

AndreiMisiukevich: Yes, but I don't think we can rewrite it without awaiting
```csharp
await base.ValidateAsync(value, token) && Uri.IsWellFormedUriString(value?.ToString(), UriKind);
```

we need both values (one from the base class and another from here)

pictos: Yes, the code where we have this snippet that you paste here, needs to be async this ValidadeAsync doesn't need. Or is this snippet just to illustrate the idea and doesn't exist on our code.

AndreiMisiukevich: @pictos now every developer can write custom class inherited from ValidationBehavior and define custom logic of validation that can be async.

As for our validators, they all don't need to be async

pictos: So far you don't show me why we need this to be async. The method that should be async is the method that calls this `ValidateAsync` method. But maybe I'm missing something... Since this is one line of code, I believe that we can agree that we can add the ConfigureAwait(false) at least. The same for the previous one.

Since you said that "(...) they all don't need to be async" we change the return type to `ValueTask` them when we don't need to wait we don't pay the price for it.

AndreiMisiukevich: @pictos can you please show me what you mean. How can we avoid "async" in this particular case?

AndreiMisiukevich: I don't think we can put configureAwait(false) here, because we want to return back to UI context as soon as validation is ended (for updating styles)

pictos: I'll check out it here and play with it to have a better vision of the code

AndreiMisiukevich: ValueTask sounds good

pictos: @AndreiMisiukevich I see now about your snippet, I missed the second line on my review here on Github, in the IDE I could see it. Shame on me.

Yeah, we can't avoid `async` here. I tried with `ConfigureAwait(false)` and works. If you think that is valuable...

```csharp
protected override async Task<bool> ValidateAsync(object value, CancellationToken token)
			=> (await base.ValidateAsync(value, token).ConfigureAwait(false))
				&& Uri.IsWellFormedUriString(value?.ToString(), UriKind);
```

AndreiMisiukevich: updated)

pictos (approved):
LGTM!

Issue #832, opened 2021-02-04
[Enhancement] Localization manager with more than one resource manager


WebDucer:
## Summary

Currently the localization manager (singleton) can be initialized only with one resource manager. This is often not enough.

We ceurrently use own implementation for this purpose, that accept a list of resource managers (form common library, different app modules).

## API Changes

Non breaking changes:

https://github.com/xamarin/XamarinCommunityToolkit/blob/f62c078146a05ef92e548f71d6cbb22cea89ff8c/src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs#L16

```csharp
ResourceManager[] resourceManagers = new ResourceManager[0];
```

https://github.com/xamarin/XamarinCommunityToolkit/blob/f62c078146a05ef92e548f71d6cbb22cea89ff8c/src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs#L25-L32

```csharp
public void Init(params ResourceManager[] resources) =>
    resourceManagers = resources;

public void Init(ResourceManager resource, CultureInfo initialCulture)=>
    Init(initialCulture, resource);

public void Init(CultureInfo initialCulture, params ResourceManager[] resources)
{
    resourceManagers = resources;
    currentCulture = initialCulture;
}
```

https://github.com/xamarin/XamarinCommunityToolkit/blob/f62c078146a05ef92e548f71d6cbb22cea89ff8c/src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/LocalizationResourceManager.shared.cs#L34-L35

```csharp
public string GetValue(string text)
{
    string translated = null;

    foreach (var rm in resourceManagers)
    {
        translated = rm.GetString(text, CurrentCulture);
        if (translated != null)
        {
            break;
        }
    }

    return translated;
    // Or as LINQ
    // return resourceManagers.Select(s => s.GetString(text, CurrentCulture))
    //    .FirstOrDefault(f => f != null);
}
```

## Intended Use Case

Usage with more than one resource manager.

## Who Will Do The Work?

- [x] I am willing to take this on myself
- [ ] Just putting this out there for someone to pick up

AmrAlSayed0:
[...]

PR #919, opened 2021-02-17
Fix Singleton Pattern for LocalizationResourceManager

brminnick:
### Description of Change ###

The `LocalizationResourceManager` implements the Singleton Pattern by implementing `public static LocalizationResourceManager Current { get; }`, but still had a public constructor. 

To prevent users from creating a new instance of `LocalizationResourceManager`, and to follow [best practices for implementing the Singleton Pattern in C#](https://csharpindepth.com/articles/singleton), this Pull Request does the following:
- Removes Public Constructor, replacing it with Private Constructor
- Provides performance improvement and thread safety by implementing `System.Lazy<T>`
- Confirmed [`LocalizationResourceManagerTests` unit tests](https://github.com/xamarin/XamarinCommunityToolkit/blob/12ee0d01d66adf2c97f96d977773832edec88002/src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Helpers/LocalizationResourceManagerTests/LocalizationResourceManagerTests.cs#L7) pass
- Confirmed [`SettingsViewModel.ChangeLanguageCommand`](https://github.com/xamarin/XamarinCommunityToolkit/blob/12ee0d01d66adf2c97f96d977773832edec88002/samples/XCT.Sample/ViewModels/SettingViewModel.cs#L34-L38) still works in the sample app

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


pictos:
@brminnick can you target the `develop` branch? Since this will be a breaking change.
Also, thanks for moving this forward, I started this change [here](https://github.com/xamarin/XamarinCommunityToolkit/pull/770#issuecomment-761709473) but we move back to be able to release this on `main`. Also, during this change, some devs asked to not do it because that way they will not be able to create different instances with different resources when you have different assemblies in the project, as you can see [here](https://github.com/xamarin/XamarinCommunityToolkit/issues/769#issuecomment-761703648).

I'm open to suggestions on how to solve this second issue if you have one.

pictos (approved):

brminnick:
No problem! 

I opened it on `main` since the public constructor, `public LocalizationResourceManager()`, was marked `[Obsolete]`, meaning it _shouldn't_ be a breaking change. 

But I understand that there is a chance of a dev out there initializing the class 👍

brminnick:
Regarding https://github.com/xamarin/XamarinCommunityToolkit/issues/769#issuecomment-761703648, I would argue that `LocalizationResourceManager.Init` and `LocalizationResourceManager.Get` are not necessary.

With RESX files in .NET, we can just reference the `string` constant in the RESX file, and .NET will automatically chose the correct translation. 

A good example is my app GitTrends, which is translated into 9 languages. To reference the translated string for the AboutPage title, for example, [I just call `PageTitles.AboutPage`](https://github.com/brminnick/GitTrends/blob/b643b9b5c787998b4b778bd3a9703b884c3e5e0c/GitTrends/Pages/AboutPage.cs#L30) from `PageTitles.resx`and `AboutPage` is treated like a string constant. Since the [`CultureInfoDefaultThreadCulture` and the `CultureInfo.DefaultThreadUICulture` has already been set to the preferred language](https://github.com/brminnick/GitTrends/blob/b643b9b5c787998b4b778bd3a9703b884c3e5e0c/GitTrends/Services/LanguageService.cs#L49-L68), the RESX file returns the translated value for `PageTitles.AboutPage`.

brminnick:
The other option is to open a feature request that would add support to `LocalizationResourceManager` for multiple `ResourceManager`s.

maxkoshevoi:
> `LocalizationResourceManager.Get` are not necessary

In the C# code yes, but it's needed for our `TranslateExtension` (more specifically indexer is used there, but `Get` and indexer do the same thing). We need to use it there since `LocalizationResourceManager` can report when culture is changed 

> Since the [`CultureInfoDefaultThreadCulture` and the `CultureInfo.DefaultThreadUICulture` has already been set to the preferred language](https://github.com/brminnick/GitTrends/blob/b643b9b5c787998b4b778bd3a9703b884c3e5e0c/GitTrends/Services/LanguageService.cs#L49-L68), the RESX file returns the translated value for `PageTitles.AboutPage`.

Sorry, I got a little confused here. Shouldn't [PageTitles.Culture](https://github.com/brminnick/GitTrends/blob/b643b9b5c787998b4b778bd3a9703b884c3e5e0c/GitTrends.Mobile.Common/Constants/PageTitles.Designer.cs#L41) be changed in order for return value of `AboutPage` to change?

WebDucer:
> The other option is to open a feature request that would add support to `LocalizationResourceManager` for multiple `ResourceManager`s.

Already suggested this here: https://github.com/xamarin/XamarinCommunityToolkit/issues/832

PR #920, opened 2021-02-17
Add `#region` to specify Linker Work-Around

### LINKED ISSUE #910: [Housekeeping] Wrap Linker-Specific Code in a `#region`, adding explaination
## Problem

The Xamarin.CommunityToolkit contains a few work-arounds to prevent the Linker from accidentally removing platform-specific code. One example can be found here: https://github.com/xamarin/XamarinCommunityToolkit/pull/674/files/492d4f0908e84b9141deeb05dc19e3e5878ea105#r576375815

Typically, this code looks like it can be removed because it is "not used", but it is indeed required to tell the linker to keep the code in the compiled binary.

The concern is that future devs may remove this "unused code", accidentally causing a breaking change.

## Solution

As [discussed here](https://github.com/xamarin/XamarinCommunityToolkit/pull/674#discussion_r576380021), let's wrap this linker-specific code in a `#region` and a comment that alerts future devs to its purpose, like so:

```csharp
#region Required work-around to prevent linker from removing the platform-specific implementation 

#if __ANDROID__
				_ = new Xamarin.CommunityToolkit.Android.Effects.LifeCycleEffectRouter();
#elif __IOS__
				_ = new Xamarin.CommunityToolkit.iOS.Effects.LifeCycleEffectRouter();
#elif UWP
				_ = new Xamarin.CommunityToolkit.UWP.Effects.LifeCycleEffectRouter();
#endif

#endregion
```

## Who Will Do The Work?

brminnick:
## Problem

The Xamarin.CommunityToolkit contains a few work-arounds to prevent the Linker from accidentally removing platform-specific code. One example can be found here: https://github.com/xamarin/XamarinCommunityToolkit/pull/674/files/492d4f0908e84b9141deeb05dc19e3e5878ea105#r576375815

Typically, this code looks like it can be removed because it is "not used", but it is indeed required to tell the linker to keep the code in the compiled binary.

The concern is that future devs may remove this "unused code", accidentally causing a breaking change.

## Solution

As [discussed here](https://github.com/xamarin/XamarinCommunityToolkit/pull/674#discussion_r576380021), let's wrap this linker-specific code in a `#region` and a comment that alerts future devs to its purpose, like so:

```csharp
#region Required work-around to prevent linker from removing the platform-specific implementation 

#if __ANDROID__
				_ = new Xamarin.CommunityToolkit.Android.Effects.LifeCycleEffectRouter();
#elif __IOS__
				_ = new Xamarin.CommunityToolkit.iOS.Effects.LifeCycleEffectRouter();
#elif UWP
				_ = new Xamarin.CommunityToolkit.UWP.Effects.LifeCycleEffectRouter();
#endif

#endregion
```

### Bugs Fixed ###

- Fixes #910 

### API Changes ###

None

### Behavioral Changes ###

None


pictos (dismissed):

maxkoshevoi:
There is a warning `SA1123	Region should not be located within a code element.` when using regions like that.
Should it be suppressed around those regions or should its severity be reduced?

jfversluis:
> There is a warning `SA1123 Region should not be located within a code element.` when using regions like that.
> Should it be suppressed around those regions or should its severity be reduced?

In order to try and minimize our warnings this seems like a good idea. @brminnick thoughts?

jfversluis (approved):

PR #931, opened 2021-02-18
Cherry-pick Sourcelink fix to main from dev

jfversluis:


pictos (approved):

pictos (changes requested):
Don't merge this yet, I just discovered that - for some reason - this breaks debug on Android. At least on VSWindows, the breakpoints aren't respected. Tomorrow I'll try to find why this happens.

jfversluis:
Could it be because we now also include the nuget in debug? I don't see why that would be necessary?

pictos:
Could be, TBH not sure how Mono handles that, because for UWP everything works as expected in debug area, but for android don't. I give some tries and see if I can find something.

jfversluis:
I'm going to skip this. We'll release 1.1 soon which has all the right fixes for this

PR #932, opened 2021-02-18
Generate Mdoc documentation files

### LINKED ISSUE #76: Integrate MDoc for API documentation
See: https://github.com/xamarin/Caboodle/tree/master/Caboodle

jfversluis:
This doesn't really do anything for the automatic builds etc. but adds some targets that I use to generate the API documentation. For now I do that manually/locally, hopefully in the future more automated.

Fixes #76 

Instructions for the manual generation here: https://github.com/xamarin/XamarinCommunityToolkit/wiki/Generate-API-documentation-with-mdoc


jfversluis:
@brminnick @pictos @AndreiMisiukevich could you double-check if I'm not doing anything dumb here?

I think some things can be rearranged to make it a bit more cleaned up and organized, but for now this works and shouldn't interfere with what's there today.

brminnick:
Oh cool - I hadn't heard of MDoc before, and I love how it cleaned up all of the API documentation!

Is there an example repo that also uses MDoc that I can reference to review this PR? I can't seem to find the docs anywhere for MDoc.

pictos:
@brminnick you can look into Xamarin.Essentials, they use this feature.

docs/frameworks.xml
```diff
@@ -0,0 +1,9 @@
+<Frameworks>
+  <Framework Name="xamarin-community-toolkit" Source="xamarin-community-toolkit-sdk" />
+  <!-- <Framework Name="xamarin-community-toolkit-android" Source="xamarin-community-toolkit-android" />
```
pictos, line 3: can we remove those?

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/StateToBooleanConverter.shared.cs
```diff
@@ -6,7 +6,7 @@
 namespace Xamarin.CommunityToolkit.Converters
 {
 	/// <summary>
-	/// This converter can be used with <see cref="StateLayout"/> to determine if a certain state is visible. This can be useful, for instance, in scenarios where you want to show/hide certain elements based on the current <see cref="StateLayout.LayoutState"/>. Additionally a <see cref="StateLayout"/> can be supplied in the parameter of the Convert method to compare against that.
+	/// This converter can be used with <see cref="StateLayout"/> to determine if a certain state is visible. This can be useful, for instance, in scenarios where you want to show/hide certain elements based on the current <see cref="StateLayout"/>.CurrentState. Additionally a <see cref="StateLayout"/> can be supplied in the parameter of the Convert method to compare against that.
```
pictos, line 9: I'm not sure about it but the CurrentState shouldn't be inside the cref?

src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj
```diff
@@ -32,6 +32,8 @@
     <Configurations>Debug;Release</Configurations>
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
     <NoWarn>SA1123</NoWarn>
+    <MDocDocumentationDirectory>$(MSBuildThisFileDirectory)..\docs\en</MDocDocumentationDirectory>
```
pictos, line 35: we can move all those changes to [here](https://github.com/xamarin/XamarinCommunityToolkit/blob/develop/src/Directory.Build.props) and avoid duplication between Markup and XCT projects

src/mdoc.targets
```diff
@@ -0,0 +1,269 @@
...
+    compiled assemblies.
+    This runs as a separate task.
+    ============================================================
+  -->
+  <Target Name="MDocUpdateDocs"
+          DependsOnTargets="_MDocSetup">
+    <!-- bits for resolving references on Windows -->
+    <PropertyGroup Condition=" '$(OS)' == 'Windows_NT' ">
+      <FrameworkReferenceAssemblyPath>$(VsInstallRoot)\Common7\IDE\ReferenceAssemblies\Microsoft\Framework</FrameworkReferenceAssemblyPath>
+      <WindowsKitsReferenceAssemblyPath>$(MSBuildProgramFiles32)\Windows Kits\10\References\10.0.17763.0</WindowsKitsReferenceAssemblyPath>
```
pictos, line 74: I believe that our WindowsSDK version is different.

pictos (changes requested):
I didn't test it on my end yet, but it looks right. With this I believe that we can remove the Summary comments from the code, right?

jfversluis:
> Oh cool - I hadn't heard of MDoc before, and I love how it cleaned up all of the API documentation!
> 
> Is there an example repo that also uses MDoc that I can reference to review this PR? I can't seem to find the docs anywhere for MDoc.

SkiaSharp, Essentials and Forms are using it, _but_. They do it all manually, generate the structure and update the XML values manually. As I've understood they are going to pivot (eventually) to doing these inline XML comments and generate it from that as well.

So, what happens with this is; we comment all public stuff, export the Visual Studio XML for both the IntelliSense in the NuGet and when we release a new version I will manually generate the new Docs so that will be updated too. The latter will happen by taking that same VS XML and transform it into the MDoc format which is used for the Docs engine.

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/StateToBooleanConverter.shared.cs
jfversluis: The problem is that `CurrentState` is an attached property so it won't recognize it and will throw errors in the Docs generation

src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj
jfversluis: Haha yeah that was what I meant by more optimization. But ok, ok.... You're so tough! I will do it right now which is better anyway :D

src/mdoc.targets
jfversluis: I took it from here: https://github.com/xamarin/XamarinCommunityToolkit/blob/main/src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj#L5

jfversluis:
Only thing I still want to check is to see if that mdoc package now shows up as a dependency to our nuget, then I should fix that, other than that this is good to go

jfversluis:
> Only thing I still want to check is to see if that mdoc package now shows up as a dependency to our nuget, then I should fix that, other than that this is good to go

It does :(

@brminnick do you happen to know a way to _not_ include a nuget package as a reference for when packing the Toolkit to a Nuget? If that makes sense? :D

I think the only way to do that is to do it through a nuspec file..

brminnick:
> do you happen to know a way to not include a nuget package as a reference for when packing the Toolkit to a Nuget

@jfversluis Yup! In the CSPROJ, append `PrivateAssests="all"`:

```xml
<PackageReference Include="mdoc" Version="5.8.0" PrivateAssets="all" />


[Here's an example of where we're actually doing that already for the NETStandard.Library](https://github.com/xamarin/XamarinCommunityToolkit/blob/aba6205274a7fb8c8a55eaedb8001e3a693ed5f1/src/CommunityToolkit/Xamarin.CommunityToolkit/Xamarin.CommunityToolkit.csproj#L82) in the Toolkit!

jfversluis:
Awesome Brandon, thank you!

Using my superpowers to merge this as this doesn't impact anything but is used now for generating the API documentation

PR #935, opened 2021-02-18
Refactoring: Remove redundant logic from "Not" converters

### LINKED ISSUE #936: [Refactoring] Remove redundant logic from "Not" converters
### Description

We have several pairs of convertors like `IsNullOrEmptyConverter` and `IsNotNullOrEmptyConverter`. One is simply inverted version of another. But both convertors contain duplicate version of the same logic.

maxkoshevoi:
### Description of Change ###

We have several pairs of convertors like `IsNullOrEmptyConverter` and `IsNotNullOrEmptyConverter`. One is simply inverted version of another. But both convertors contain duplicate version of the same logic.

Removed all logic from "Not" convertors and replaced it with the call to regular converter with inverted result.

### Bugs Fixed ###

- Fixes #936

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


jfversluis:
Makes total sense, thanks! 

Why did you make the methods virtual though? :)

maxkoshevoi:
> Why did you make the methods virtual though? :)

`Convert` method is `virtual` in "main" convertor to be overridden in "not" convertor

pictos (changes requested):
TBH I didn't like this approach, to mark the converters methods as virtual, to me our converters could be `sealed`. Instead of doing this, why not create static methods to do the logic?

maxkoshevoi:
@pictos This could work too. Why our converters should be `sealed` though?

jfversluis:
> > Why did you make the methods virtual though? :)
> 
> `Convert` method is `virtual` in "main" convertor to be overridden in "not" convertor

🤦  it must be Monday... 

I'll let you discuss the other options :)

pictos:
@maxkoshevoi I'm sorry if I sounded rude. That isn't the intention. Also, I need to comment fast because it was about to start my work hours. So let me rephrase that.

Inherit is dangerous and could lead to a bug and unexpected behaviors so we should use it in places that are designed for it. I can't have the reference that I read a while ago but if you google it you will find a lot of references about it. So because of that I suggest using another approach

maxkoshevoi:
@pictos No worries =)
Thanks for the explanation. I already updated PR to use static methods instead of inheritance.

Does this version look good to you?

pictos:
I'll update the branch here, and take a look in a bit (;

AndreiMisiukevich:
Hm, unit tests fail for some reason

maxkoshevoi:
> Hm, unit tests fail for some reason

Looks like https://github.com/xamarin/XamarinCommunityToolkit/pull/965 broke something (but tests passed for it, strange).
Tests are now failing for this PR too: https://github.com/xamarin/XamarinCommunityToolkit/pull/963

@AndreiMisiukevich Could you please run CI pipeline against `main` to double check?

AndreiMisiukevich:
@brminnick can you please check if #965 broke unit tests? :)

maxkoshevoi:
> @brminnick can you please check if #965 broke unit tests? :)

Or, maybe it was this one.. #954

image

brminnick:
image
When I run the Unit Tests for this PR locally in VS for Mac and Visual Studio on PC, the only test that fails is this one:
```bash
Xamarin.CommunityToolkit.UnitTests.Converters.DateTimeOffsetConverter_Tests.DateTimeOffsetConverterBack(value: 2020-08-25T13:37:00.0000000-07:00, expectedResult: 2020-08-25T13:37:00.0000000-08:00
```

I'll re-run the GitHub Actions to see if the failures were an anomaly.

brminnick:
Yup - after re-running the GitHub Actions, all of the Unit Tests are passing ✅

pictos (commented):
@maxkoshevoi I did some changes here, please let me know what do you think about it

maxkoshevoi:
@pictos I'm not sure about this one. Why is it better to move convertor logic outside of convertor class? Logic for conversion should be encapsulated inside the class for respective convertor.
Now logic from some, but not all convertors is in separate `ConverterMethods` class
1) It's not obvious that it's located there and when new convertor is created if logic for it should also be located there
2) It contains unrelated methods (violating single responsibility principle)
3) It potentially will just keep growing into a lot of methods that are only intended to be used by the two classes with names that match method name (with or without "not") each.

Anyway, just wanted to hear your reasoning for this approach.

pictos:
@maxkoshevoi 
> I'm not sure about this one. Why is it better to move convertor logic outside of convertor class? Logic for conversion should be encapsulated inside the class for respective convertor.
Now logic from some, but not all convertors is in separate ConverterMethods class

Well, your PR is to reuse code, so when you did it first you remove part of the logic from the convert and reuse some other implementation. The final user will not see it because is `internal` and for us (maintainers) is better to have all of the shared convert logic in the same place to make it easier to find and maintain it.

> 1. It's not obvious that it's located there and when new convertor is created if logic for it should also be located there

That class should be use if there's some method that we can reuse, like in the case of those 3, maybe change the class name would make it more clear.

> 2. It contains unrelated methods (violating single responsibility principle)

I believe that it doesn't apply in that case, that is a `helper` static class to hold all logic that can be reused between converters

> 3. It potentially will just keep growing into a lot of methods that are only intended to be used by the two classes with names that match method name (with or without "not") each.

I don't believe that will grow up that much, we have 23 converters there and just 3 reused methods. Probably we will have more methods that can be reused in the future but will be a feel

maxkoshevoi:
@pictos **Disclaimer**. Everything below is purely my opinion. In my practice I saw a lot of `helper` or `utility` classes that just contained a lot of vaguely related methods, were impossible to read and figure out what is called from where, and at the end needed to be refactored into OOP classes with single SOLID in mind. So I just don't like those sorts of static `helper` classes with unrelated methods in general. 

This repository isn't mine to maintain, so please feel free to ignore everything below, and implement code in a way that will be easier for you and other members to maintain. I'm just happy that this code isn't being duplicated anymore =)

> for us (maintainers) is better to have all of the shared convert logic in the same place to make it easier to find and maintain it.

I think it's easy to find "IsNullOrEmpty" converting logic inside "IsNullOrEmpty" convertor class. If we are searching for "IsNotNullOrEmpty" converting logic, if it would say `!IsNullOrEmptyConvertor.ConvertInternal(..)`, 
1) it should be pretty obvious that this convertor returns inverted result of "IsNullOrEmptyConvertor"
2) You are one F12 away from looking at converting logic itself

I don't see how it's not easy to maintain. Main convertor contains main logic, and some other convertor uses it and modifies the result in some way. 
Does this repository has some other "helper" classes like this?

> I believe that it doesn't apply in that case, that is a `helper` static class to hold all logic that can be reused between converters

It should also apply to `helper` classes. What I'm trying to say here is when you will modify one method from this class, it will have nothing to do with all the another methods. When you'll be reading this class you'll have to constantly switch context because methods are not related in any way (except that they convert something to something), work with different input types, and intended to check for completely different things. What I'm saying is: having those methods in the same place won't help maintain them.

pictos:
@maxkoshevoi nether my. I see your point. So I'll revert my changes and move it to the previous state. Thanks for your inputs (;

pictos:
@maxkoshevoi can you double-check the files?

maxkoshevoi:
@pictos looks good

pictos (approved):

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/EqualConverter.shared.cs
```diff
@@ -18,8 +18,10 @@ public class EqualConverter : ValueConverterExtension, IValueConverter
 		/// <param name="parameter">The second object to compare.</param>
 		/// <param name="culture">The culture to use in the converter. This is not implemented.</param>
 		/// <returns>True if <paramref name="value"/> and <paramref name="parameter"/> are equal, False if they are not equal.</returns>
-		public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
-			=> (value != null && value.Equals(parameter)) || (value == null && parameter == null);
+		public object Convert(object value, Type targetType, object parameter, CultureInfo culture) => ConvertInternal(value, parameter);
+
+		internal static bool ConvertInternal(object value, object parameter) =>
```
AndreiMisiukevich, line 23: Why is it static?

It can be protected and instace

maxkoshevoi: Initial solution was [exactly like you describe](https://github.com/xamarin/XamarinCommunityToolkit/pull/935/commits/62c43e1a34699381486ea4e9ea477b337b0dd019), but @pictos wants convertors to be sealed to  avoid issue with inheritance. I think that discussion got deleted, I cannot find it. So we changed fix to use static methods.

pictos: @maxkoshevoi also with inheritance didn't work, right? Or am I mix with another issue?

maxkoshevoi: @pictos It did. 
I just tried to do it without `virtual` using `new` instead of `override` after your request to not use `virtual` (just wanted to see if it would work), and that didn't work. But `virtual` - `override` worked as expected.

pictos: Got ya. And yeah, I want to avoid inheritance, since isn't a good practice in places that we don't expect it.

Issue #889, opened 2021-02-11
[Bug] Popup on UWP throws an exception when accessed from the nightly nuget


bmacombe:
### Description
![image](https://user-images.githubusercontent.com/15127788/107686230-8fcf6900-6c6a-11eb-9502-87f9708e9c4b.png)

Acts as if the UWP dll isn't being referenced or it's not package correctly.  Works fine in the sample app at the develop branch head at time of report.

Android and iOS work as expected.

### Steps to Reproduce

1. Run the sample app and click the button.

### Expected Behavior
Popup opens

### Actual Behavior
Not supported exception

### Basic Information

- Version with issue: Nightly v.1.1.0-preview1010
- Last known good version:
- IDE: VS 19
- Platform Target Frameworks: 
  - UWP: 1903
- Nuget Packages: Nightly v.1.1.0-preview1010
- Affected Devices:

### Reproduction Link

https://github.com/bmacombe/XCT-Popup-Issue

bmacombe:
@ahoefling @pictos FYI

pictos:
I can confirm it, using the NuGet I can't open the popUp on UWP, but if I reference the XCT source code everything works fine.

pictos:
@jfversluis any thought?

jfversluis:
Well that's interesting. That looks like a rather custom message that is thrown in that exception, but I can't seem to find it.

bmacombe:
Hey @jfversluis, sorry been a bit busy.  But here is where the custom error comes from.

https://github.com/xamarin/XamarinCommunityToolkit/blob/3f7bc8c05f75c979342f1c3cc204e6da9ec363d7/src/CommunityToolkit/Xamarin.CommunityToolkit/Extensions/Navigation/NavigationExtensions.netstandard.macos.tvos.watchos.tizen.wpf.gtk.cs#L1-L16

My un-educated guess is somehow that file is being built for UWP instead of the *.uwp.cs.

ahoefling:
As others posted, I am confirming this is happening with the sample project. I added the XCT as a submodule and the latest code in the `develop` branch appears to work.

My initial investigation points to a NuGet packaging issue. I copied the dll from my packages directory for `uap10.0.19041` and then added it as a direct dll reference to the UWP app. When I ran the application the popup displayed without issue. I think the exception is being thrown because the .NET Standard dll is being used from the shared project and it is not swapping out the UWP dll in the project head.

[...]

PR #653, opened 2020-12-04
Adds Popup Control

ahoefling:
### Description of Change ###
Adds native popups to iOS/Android/UWP

This PR ports the original popup control implementation from Xamarin.Forms to the Xamarin Community Toolkit. This PR is more complete than the original PR and there is better feature parity between the 3 major platform (iOS/Android/UWP). Please see the table below which documents what features are implemented on each platform.

### Feature Add ###
- Closes #292

### API Changes ###
Added: 
- `Popup`
- `Popup<T>`
- `BasePopup`

The popup control allows developers to create popups and they can be extended to return data to the invocation code. This is why there are 3 different classes for XCT Popups. The `BasePopup` is what is needed at minimum to render a popup on screen. Then `Popup<T>` allows generic results where `Popup` will just return an `object`.

Below documents all the APIs added

#### BasePopup
```c#
View Content { get; set; }
Color Color { get; set; }
LayoutOptions VerticalOptions { get; set; }
LayoutOptions HorizontalOptions { get; set; }
Color BorderColor { get; set; }
View Anchor { get; set; }
Size Size { get; set; }
bool IsLightDissmissEnabled { get; set; }
event EventHandler<PopupDismissedEventArgs> Dismissed;
event EventHandler<PopupOpenedEventArgs> Opened;
void LightDismiss();
```

#### Popup<T>
```c#
void Reset();
void Dismiss(T result);
Task<T> Result;
```

#### Popup
Nothing different than `Popup<T>` it just uses `object` as the type definition

### Feature Parity Table
This is a very large Pull Request that adds a set of complex features to the 3 major platform (iOS, Android and UWP). It wasn't possible to get perfect feature parity across the platforms and the table below documents what is implemented in each platform

| Feature                               | Android   | iOS | UWP |
|------------------------------|------------|-----|-------|
| Simple Popup                     | ✅           | ✅ | ✅     |
| Light Dismiss (Enabled)      | ✅           | ✅ | ✅     |
| Light Dismiss (Disabled)     | ✅           | ✅ | ❌     |
| Transparent Backgrounds   | ✅           | ❌ | ✅     |
| Custom Background Color | ✅           | ✅ | ✅     |
| Event - Opened                  | ✅           | ✅ | ✅     |
| Event - Dismissed               | ✅           | ✅ | ✅     |
| Layout Positioning              | ✅           | ✅ | ✅     |
| Dynamic Sizing                   | ✅           | ✅ | ✅     |
[...]

PR #942, opened 2021-02-19
Uwp retarget to 17763 / 1809

inforithmics:
### Bugs Fixed ###

- Fixes https://github.com/xamarin/XamarinCommunityToolkit/issues/889

### API Changes ###

Changed Target Windows version back to 17763 / 1809
from 19031 / 2004

### Behavioral Changes ###

works now again on Windows 17763 / 1809

### PR Checklist ###


jfversluis:
Hm I would think we've set it to that target for a reason :D

pictos:
@jfversluis if I'm not wrong was for popup control. Need to test this change to see if this doesn't break anything else

jfversluis:
@ahoefling might know?

Let's not merge this until we know/properly test this

inforithmics:
Here the link around the discussion about this.

https://github.com/xamarin/XamarinCommunityToolkit/pull/653#discussion_r536516886

ahoefling:
The UWP Flyout control added the ability to specify a transparent background in 19041. @inforithmics the discussion you linked explains this in my first comment https://github.com/xamarin/XamarinCommunityToolkit/pull/653#discussion_r537581216. This change appears to be a workaround to our packaging problem.

The change provided here downgrades the lowest version supported and then applies the check for the feature a different way. I still believe the issue with UWP is packaging and not a code change like suggested here.

We should focus our efforts on getting this to work with windows build 19041.

ahoefling:
I just added a comment on the original issue. This is not a packaging bug and can be solved by updating the min target version.

I think we can close both PR and issue as not a bug

https://github.com/xamarin/XamarinCommunityToolkit/issues/889#issuecomment-784553594

inforithmics:
I agree that increasing the Min Version would solve the problem. But when you want to use the Xamarin Community Toolkit Popup on the oldest supported Windows Version you can't at moment.

https://en.wikipedia.org/wiki/Windows_10_version_history

Here you see all Windows 10 Supported Versions 1803 / 1809 are still supported until May 11 2021 (Enterprise Version)
And 1809 is still supported until January 9 2029 as a LTSC Version.

So that's why I retargeted it to Version 1809 as it was before.

With WinUI 3 this won't matter anymore, but unfortunatly the earlist that WinUI 3 on UWP will be supported is autumn of this year (not commited yet by WinUI 3).

I although created a new Xamarin.Forms UWP project and the default Minimum UWP Version was 16299 /1709 (The Minimum Version of Xamarin.Forms).

I like the Xamarin Community Toolkit Popup and would like to use it, but I can't at the moment because I'm only able to set the Minimum Supported Version to 17763/1809.

inforithmics:
They changed now the WinUI 3 roadmap WinUI 3 on UWP is now included as a Preview Feature in March 2021 

https://github.com/microsoft/microsoft-ui-xaml/blob/master/docs/roadmap.md

But I don't know what happens when Xamarin Forms stays on UWP and Xamarin Community Toolkit uses WinUI 3 Toolkit

YZahringer:
I agree with @inforithmics, the UWP min version should be `1809`, it is an `LTSC` version and we always support it in all our projects.

bmacombe:
@jfversluis @pictos  I remember last summer talking about possibly needing to split the toolkit into more packages.  While I'd like to avoid that, it might be something to consider if the min version can't be moved back past 2004.   I think the Popup addition is great, but it's blocking me on XCT now because of the Win 10 min version in my user base.   And I'm lucky that I only have users back to 1909.  I know I could fork and roll my own, but that seems to defeat the purpose of the toolkit.

image

I'm not an expert at packaging and version detection, but what harm would this change do?  If it would cause the popup to not work correctly and seem buggy or something on Win 10 < 2004 without the transparency setting, maybe it could just throw a not-supported exception or something similar.

inforithmics:
If someone wants to test the changes the nuget package can be downloaded from this link:
https://dev.azure.com/xamarin/public/_build/results?buildId=38666&view=artifacts&pathAsName=false&type=publishedArtifacts

is for the https://github.com/xamarin/XamarinCommunityToolkit/pull/942/commits/667855ba0e1ac3564ca8a240e316e845cb252679 checkin.

PR #946, opened 2021-02-20
Msbuild hotfix

pictos:
With these changes now we're able to debug Android again and release nugets with source link. Also tested on iOS and is working

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<PropertyGroup Condition=" '$(Configuration)'=='Release' And '$(OS)' == 'Windows_NT' ">
+    <DebugType>embedded</DebugType>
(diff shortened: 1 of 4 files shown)


src/Directory.Build.props
```diff
@@ -0,0 +1,26 @@
...
+<ItemGroup Condition=" '$(Configuration)'=='Release'">
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
+</ItemGroup>
+
+<PropertyGroup Condition=" '$(Configuration)'=='DEBUG' And '$(OS)' == 'Windows_NT' ">
+    <DebugType>full</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+</PropertyGroup>
+
+  <Import Project="../Directory.Build.props" />
```
jfversluis, line 23: Doesn't this happen automatically?

jfversluis (approved):

src/Directory.Build.props
pictos: From my research we need to import, what's automatically is the csproj import the Build.props

PR #854, opened 2021-02-07
[iOS] Remove popup arrow if anchor isn't set

geskill:
### Description of Change ###
This removes the arrow of the popup if the anchor isn't set.

### Bugs Fixed ###

- Fixes #855

### API Changes ###

None.

### Behavioral Changes ###

None.

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [x] Has samples (if omitted, state reason in description)
- [ ] Rebased on top of main at time of PR
- [x] Changes adhere to coding standard

- [ ] Updated [documentation](https://github.com/MicrosoftDocs/xamarin-communitytoolkit)

PR #949, opened 2021-02-20
Platform specific API

### LINKED ISSUE #948: ♻ Provide the base implementation for PlatformSpecific APIs
With PopUp control, we have our first PlatformSpecificAPI (e.g. `PopoverArrowDirection`), and we already have specs that want to support features in only one platform. So we need to think in a way to keep it organized. The Forms implementation for that. Forms team made [this API public](https://github.com/xamarin/Xamarin.Forms/pull/12278) for us, so let's use it!

pictos:
This PR implements the Platform Specific APIs using Xamarin.Forms API. We need to agree with:

1. Folder structure;
2. Namespace structure;

With this implementation, we are able to implement platform-specific features in the same way that Forms does.

We can do it on C#
```csharp
using Xamarin.Forms.PlatformConfiguration;
using Xamarin.CommunityToolkit.PlatformConfiguration.iOSSpecific;
using Xamarin.CommunityToolkit.PlatformConfiguration.WindowsSpecific;

// API usage example
On<iOS>().UseArrowDirection(PopoverArrowDirection.Right);
On<Windows>().SetBorderColor(Xamarin.Forms.Color.Red);
```
And in XAML

```xml
xmlns:uwpXct="clr-namespace:Xamarin.CommunityToolkit.PlatformConfiguration.WindowsSpecific;assembly=Xamarin.CommunityToolkit"
xmlns:iosXct="clr-namespace:Xamarin.CommunityToolkit.PlatformConfiguration.iOSSpecific;assembly=Xamarin.CommunityToolkit"
iosXct:PopUp.ArrowDirection="Right"
uwpXct:PopUp.BorderColor="Blue"
```

Following the Forms implementation, the `Xamarin.CommunityToolkit.PlatformConfiguration` shouldn't be in our custom schema (`http://xamarin.com/schemas/2020/toolkit`).

Fixes #948


src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/iOSSpecific/PopUp.shared.cs
```diff
@@ -0,0 +1,36 @@
...
+		public static readonly BindableProperty ArrowDirectionProperty = BindableProperty.Create(
+			"ArrowDirection", typeof(PopoverArrowDirection), typeof(BasePopup), PopoverArrowDirection.None);
+
+		public static void SetArrowDirection(BindableObject element, PopoverArrowDirection color) =>
+			element.SetValue(ArrowDirectionProperty, color);
+
+		public static PopoverArrowDirection GetArrowDirection(BindableObject element) =>
+			(PopoverArrowDirection)element.GetValue(ArrowDirectionProperty);
+
+		public static IPlatformElementConfiguration<XFPC.iOS, XCTElement> UseArrowDirection(this IPlatformElementConfiguration<XFPC.iOS, XCTElement> config, PopoverArrowDirection value)
```
pictos, line 19: need to use the `XFPC.iOS` because the compiler complains that `iOS` is a namespace

AndreiMisiukevich:
Do we really want to handle it by "PlatformConfiguration" as Forms does?

Maybe, we might add properties to shared classes with some comment telling that a certain property is supported only by iOS, for example, or iOS and Android etc.
Or even add iOS or Android prefix/postfix.

```IOSArrowDirection```
```ArrowDirectionPlatformSpecific```
```
/// This propertu is iOS-specific
ArrowDirection
```

CC @jfversluis @pictos @sthewissen @jsuarezruiz

**Comment by pictos** [comment]
> Maybe, we might add properties to shared classes with some comment telling that a certain property is supported only by iOS, for example, or iOS and Android etc.

@AndreiMisiukevich the problem that I see with this approach, is that isn't very clear/explicit. In that way at first glance looks like that feature is supported by all platforms. I'm ok to not use the Forms API and create our own but we need to make it in a way that will be explicitly and reusable.

> Or even add iOS or Android prefix/postfix.

This may help, but I need to think a bit more about it, right now I can which can lead to some confusion for contributors.

**Comment by Cfun1** [comment]
As discussed with @pictos it could be problematic when for the same feature/functionality (at shared code level) some flags are only supported by specific platform, while in others it doesn't exist or with a different name, not sure yet but I think this could be a solution to that problem. We could rename or map the flags internally but it won't be a nice approach I think.

@AndreiMisiukevich my concern is xaml platform specific flags like in this spec https://github.com/xamarin/XamarinCommunityToolkit/issues/718.

File: samples/XCT.Sample/Pages/Views/Popups/TransparentPopup.xaml [hunk 2]
```diff
@@ -1,19 +1,21 @@
...
-           iosxct:Popup.ArrowDirection="Up"
-           uwpxct:Popup.BorderColor="Transparent"
-           x:Class="Xamarin.CommunityToolkit.Sample.Pages.Views.Popups.TransparentPopup">
+<xct:Popup
+    x:Class="Xamarin.CommunityToolkit.Sample.Pages.Views.Popups.TransparentPopup"
+    xmlns="http://xamarin.com/schemas/2014/forms"
+    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+    xmlns:local="clr-namespace:Xamarin.CommunityToolkit.Sample.Pages.Views.Popups"
+    xmlns:xct="http://xamarin.com/schemas/2020/toolkit"
+    Size="{x:Static local:PopupSize.Tiny}"
```
jfversluis, line 8: Totally unrelated, I think we can create an extension for this, right? I remember us doing it for something else, that will make this syntax better

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/iOSSpecific/PopUp.shared.cs
```diff
@@ -0,0 +1,36 @@
...
+
+	public enum PopoverArrowDirection
+	{
+		None,
+		Up,
+		Down,
+		Left,
+		Right,
+		Any,
+		Unknown
```
jfversluis, line 34: What's the use of `Unknown`? @ahoefling?

jfversluis:
> Do we really want to handle it by "PlatformConfiguration" as Forms does?
> 
> Maybe, we might add properties to shared classes with some comment telling that a certain property is supported only by iOS, for example, or iOS and Android etc.
> Or even add iOS or Android prefix/postfix.
> 
> `IOSArrowDirection`
> `ArrowDirectionPlatformSpecific`
> 
> ```
> /// This propertu is iOS-specific
> ArrowDirection
> ```

As already mentioned by the other people above there is some concerns with that too. Both ways have their up- and downsides for sure. Picking between these I think the best choice is to be consistent with Forms and use the platform-specific stuff. People know that approach and there is no confusion over what is implemented in what platform or clutter our property names with prefixes etc.

But I appreciate you trying to find an alternative Andrei! It would be easy to just copy what Forms does without thinking, we should stay sharp :)

ahoefling:
@AndreiMisiukevich I requested the iOS specific arrows class be added in since it is not implemented in Android or UWP. If we get to a point where arrows are implemented in the 3 major platform iOS, Android and UWP I think we can remove that class and make it a shared class.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/iOSSpecific/PopUp.shared.cs
ahoefling: This was added in #854 by @geskill and I approved the changes.

I am not 100% sure what `Unknown` or `Any` is used for. Looking at the code this maps 1:1 with the iOS implementation. Since this is iOS specific I think it is okay to leave both of them in. If we get to a point where we want to make this a shared API instead of attached property we would need to determine if it is still applicable.

samples/XCT.Sample/Pages/Views/Popups/TransparentPopup.xaml
ahoefling: What kind of markup extension are you thinking? The `PopupSize` is a helper for the sample project to make it easier to specify size.

Are you thinking about making this something that consumers of XCT could use for their Popups? That may be a really useful API.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/iOSSpecific/PopUp.shared.cs
Cfun1: Reference: https://developer.apple.com/documentation/uikit/uipopoverarrowdirection/

pictos:
@AndreiMisiukevich @jsuarezruiz @jfversluis can you re-review it?

pictos:
@brminnick , @AndreiMisiukevich @jfversluis @jsuarezruiz can you re-review this API? That way we can merge it and unblock a couple of PRs

brminnick (approved):

jfversluis (approved):

PR #950, opened 2021-02-20
Don't thrown when tap/selected is null, return null. For deselect.

jamesmontemagno:
### Description of Change ###
For example I want to write this in my VM:

```csharp
        Coffee selectedCoffee;
        public Coffee SelectedCoffee
        {
            get => selectedCoffee;
            set => SetProperty(ref selectedCoffee, value);
        }

        async Task Selected(Coffee coffee)
        {
            if (coffee == null)
                return;

            await Application.Current.MainPage.DisplayAlert("Selected", coffee.Name, "OK");
            SelectedCoffee = null;
        }
```

And this in my XAML:
```xml
    <ListView
        SelectedItem="{Binding SelectedCoffee, Mode=TwoWay}"
>
        <ListView.Behaviors>
            <xct:EventToCommandBehavior
                Command="{Binding SelectedCommand}"
                EventArgsConverter="{StaticResource ItemSelectedEventArgsConverter}"
                EventName="ItemSelected" />
        </ListView.Behaviors>
```

Today, this will throw an exception because it it null and is not of the type it is expecting, but that should be allowed.

The only work around is to do something like `AsyncCommand<object>` instead of `AsyncCommand<Coffee>`

### Bugs Fixed ###

- Fixes #

### API Changes ###

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
``` -->

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/ItemSelectedEventArgsConverter_Tests.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/ItemSelectedEventArgsConverter_Tests.cs
@@ -15,9 +15,10 @@ namespace Xamarin.CommunityToolkit.UnitTests.Converters
 		{
             // We know it's deprecated, still good to test it
 #pragma warning disable CS0618 // Type or member is obsolete
-            new object[] { new SelectedItemChangedEventArgs(expectedValue), expectedValue},
+            new object[] { new SelectedItemChangedEventArgs(expectedValue), expectedValue },
+			new object[] { null, null },
 #pragma warning restore CS0618 // Type or member is obsolete
-        };
+		};
 
 		[Theory]
 		[MemberData(nameof(GetData))]
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/ItemTappedEventArgsConverter_Tests.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/ItemTappedEventArgsConverter_Tests.cs
@@ -15,9 +15,10 @@ namespace Xamarin.CommunityToolkit.UnitTests.Converters
 			{
             // We know it's deprecated, still good to test it
 #pragma warning disable CS0618 // Type or member is obsolete
-                new object[] { new ItemTappedEventArgs(null, expectedValue), expectedValue},
+                new object[] { new ItemTappedEventArgs(null, expectedValue), expectedValue },
+                new object[] { new ItemTappedEventArgs(null, null), null },
 #pragma warning restore CS0618 // Type or member is obsolete
-            };
+			};
 
 		[Theory]
 		[MemberData(nameof(GetData))]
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by pictos** [comment]
@jamesmontemagno thanks for this PR, can you rephrase this part, I'm a little lost in there

> The only work around is to do something like **AsyncCommand** instead of **AsyncCommand**

**Comment by jamesmontemagno** [comment]
Whoops -> `AsyncCommand<object>` instead of `AsyncCommand<Coffee>`

**Comment by jamesmontemagno** [comment]
The problem is when you set "SelectedCoffee = null" in the Command it throws an exception because this code is false:
`value is SelectedItemChangedEventArgs selectedItemChangedEventArgs`

so you need to check to see if value is null first.

**Comment by jamesmontemagno** [comment]
I wasn't able to run tests locally, so someone probably needs to pull this PR down.

**Comment by pictos** [comment]
I can do it!

**Review by pictos — APPROVED** [review]

**Comment by jamesmontemagno** [comment]
Fixed up my example

**Comment by jamesmontemagno** [comment]
Take a look here -> https://github.com/jamesmontemagno/MyCoffeeApp/blob/master/MyCoffeeApp/MyCoffeeApp/ViewModels/CoffeeEquipmentViewModel.cs#L61 

I should be able to do `AsyncCommand<Coffee>` and also `async Task Selected(Coffee coffee)`. This works fine until I do `SelectedCoffee = null`

**Comment by pictos** [comment]
Yeah, I can confirm that the problem is the `null` return in the converter, to fix that we need to think in a way to know what is the type of `SelectedItemChangedEventArgs.SelectedItem` and `ItemTappedEventArgs.Item`, to cast that null to them.

image

Obs.: XCT.AsyncCommand also throws the same exception.

**Comment by pictos** [comment]
@jamesmontemagno I was able to _fix it_ on XCT but is a very ugly fix. So I would say that is a known bug and the workaround is using the object as a parameter. The null check that you added on converters is enough, I believe.

**Review by jfversluis — APPROVED** [review]

**Comment by jfversluis** [comment]
Thanks James! <3

**Comment by jamesmontemagno** [comment]
Did this actually fix the full bug or does AsyncCommand also need to be patched?

**Comment by pictos** [comment]
@jamesmontemagno the fix will be on EventToCommandBehavior, you can see my spec here #953, if you have time, please let me know what you think about it (:

**Comment by jamesmontemagno** [comment]
Ahhh gotcha! cool! Did my code do anything? I saw @jfversluis pulled it in.. :)

**Comment by pictos** [comment]
@jamesmontemagno I tested yesterday using your code and works like a charm. Now is just a matter to see if my idea is good and safe to implement (:

**Comment by jamesmontemagno** [comment]
Gotcha, so it will be a combination of your fix and my fix that fixes it up fully for me. Got it

**Comment by thecuong064** [comment]
I'm using XCT version 1.2.0 and still facing this issue. Is there anything else I'm missing?

image

This is my code.

```
<ListView
SelectedItem="{Binding SelectedItem, Mode=TwoWay}">

<ListView.Behaviors>
    <xct:EventToCommandBehavior
        Command="{Binding SelectItemCommand}"
        EventArgsConverter="{StaticResource ItemSelectedEventArgsConverter}"
        EventName="ItemSelected" />
</ListView.Behaviors>
```

```
private Item selectedItem;
public Item SelectedItem 
{ 
    get => selectedItem;
    set => SetProperty(ref selectedItem, value);
}

public AsyncCommand<Item> SelectItemCommand { get; }
private async Task ExecuteSelectItemCommand(Item item)
{
    if (item == null)
        return;

    SelectedItem = null;

    await Application.Current.MainPage.DisplayAlert("Selected", item.Text, "OK");
}
```

PR #963, opened 2021-02-22
Fix NullReferenceException when displaying Toast or SnackBar

maxkoshevoi:
### Description of Change ###

Now instead of an exception, toast/snackbar wouldn't be shown. 
Not perfect, but seems better then the exception.

### Bugs Fixed ###

- Fixes #

### API Changes ###

None

### Behavioral Changes ###

Toast/snackbar wouldn't be shown in rare cases where previously exception would be thrown.

### PR Checklist ###

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.android.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.android.cs
@@ -16,7 +16,12 @@ namespace Xamarin.CommunityToolkit.UI.Views
 	{
 		internal void Show(Page sender, SnackBarOptions arguments)
 		{
-			var view = Platform.GetRenderer(sender).View;
+			var view = Platform.GetRenderer(sender)?.View;
+			if (view == null)
+			{
+				return;
+			}
+
 			var snackBar = AndroidSnackBar.Make(view, arguments.MessageOptions.Message, (int)arguments.Duration.TotalMilliseconds);
 			var snackBarView = snackBar.View;
 			if (arguments.BackgroundColor != Forms.Color.Default)


maxkoshevoi:
@VladislavAntonyuk, @AndreiMisiukevich, @pictos Could you please take a look at this? Is there a better fix for this issue?

VladislavAntonyuk:
It looks good in terms of code stability, but we do not show the Snackbar. 
I suggest to throw custom exception that this page cannot be the "parent" for the Snackbar. 
Thoughts?

maxkoshevoi:
> this page cannot be the "parent" for the Snackbar

The thing is that the page can be "parent" for the Snackbar. This code fails only if we try to show Snackbar at the same time when we change current page. Both pages are fully capable of displaying Snackbar, but when we are in the middle of transitioning between them, `GetRenderer()` returns null

VladislavAntonyuk:
As described in the issue, you run the Snackbar on page appearing.
Looks like that sometimes the Snackbar Snow method is called before the renderer set the control.
So with your code you won't see the Snackbar, while you actually expect it.
What about other platforms? Can you reproduce it for UWP or iOS?
I would suggest to create some kind of retry policy: if the getRenderer returns null, we can wait for a few seconds and try again.
But the root cause of the issue from my point of view is that OnAppearing calls Snackbar before we actually can get the view

pictos:
@maxkoshevoi Is this just occur with Shell?

maxkoshevoi:
> Can you reproduce it for UWP or iOS?

Didn't try to reproduce it on other platforms. My app can be built only for Android as of now. Don't have time to create a new project and test it on UWP (don't have Mac to test it on iOS)

> I would suggest to create some kind of retry policy

Updated PR with retry logic instead of just returning. Tested and it works with 50 ms retry.
Forgot that we can use `await` in platform specific logic too 😄

> But the root cause of the issue from my point of view is that OnAppearing calls Snackbar before we actually can get the view

Well, actually, `OnAppearing` triggers a network request, and if it fails (which happens really quickly if device isn't connected to the internet) toast is displayed. Seems like a valid use of a toast.

> Is this just occur with Shell?

I'm not sure about that. What I can say is that it definitely happens is `Shell` app
1) When switching between tabs
2) When `Pop`ing page of the stack

pictos:
@maxkoshevoi because of that I said to wait until we see the issue and discuss it first.

Since this only happens on Shell (did you test on TabbedPage?) and in a very specific moment, navigation I would that's a Shell problem and not our lib.
Shell has some navigations corners that lead into edge cases just like this one, and I don't see value in trying to fix it here, because, maybe, this is a workaround and not a fix.

Also, let's move this discussion to the issue and please can you provide a repro?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.android.cs
```diff
@@ -70,6 +77,21 @@ internal void Show(Page sender, SnackBarOptions arguments)
...
 
+		static async Task<IVisualElementRenderer> GetRendererWithRetries(Page sender, int maxRetries = 5)
+		{
+			var retriesLeft = maxRetries;
+
+			var renderer = Platform.GetRenderer(sender);
+			while (renderer == null && retriesLeft > 0)
+			{
+				retriesLeft--;
+				await Task.Delay(50);
```
VladislavAntonyuk, line 88: still not an ideal solution.
it would be better to investigate why the renderer is null (is it XF issue or you call DisplayXXX too early - before the page is initialized).

maxkoshevoi: It's a XF Shell bug. We cannot control when we show SnackBar since we show it when user asks us to. And issue happens when we show SnackBar on one page and immediately switch to the different page. User can switch pages from the UI, so we have no way of knowing if we currently transitioning between pages (well, we can check renderer for `null`).

pictos:
@maxkoshevoi can you solve these conflicts?

maxkoshevoi:
@pictos @brminnick Why is this retargeted to `develop`? It's a bugfix that doesn't change an API.

pictos:
@maxkoshevoi I believe that happens because of the nullable PR, at that time we moved all PRs to target `develop`

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.android.cs
pictos: I agree with @VladislavAntonyuk at this, we should move with another approach, I liked the way how Shane proposed, listening the `PropertyChanged` and using a `TaskCompletionSource`, we also could add a `CancelationToken` with a timeout to avoid dead-locks.

Little snippet provided by Shane on our discord channel.
```csharp
if(Platform.GetrendereR(page) == null)
 page.ElementPropertyFChanged += WatchForRendererToSet 


void WatchForRendererToSet (_, args) 
if(args.PropertyName == "Renderer") 
{
      page.ElementPropertyChanged -= WatchForRendererToTest
       task.SetResult(Platform.GetRenderer(page));
}

```

maxkoshevoi: @pictos I thought this PR will never be merged anyway (with any implementation), since the bug is in XF, and not in this lib.

I don't agree with this. If this lib provides some functionality, it should ensure that is works (even if that mean including workarounds for bugs in 3rd party code). But that's not my decision to make.

So will this PR be able to move forward if I do the implementation you described?

pictos: @maxkoshevoi, I agree in parts with you. I agree that our lib breaking an app is really bad, even if the issue isn't here. Also, we need to try to minimize that impact by finding a workaround for it. And I believe (here is where I disagree with you) that we shouldn't add a lot of code to work around a bug, mainly if this workaround changes various APIs.

My suggestion also adds a lot of changes to this API, so maybe isn't the better of the world.
The short answer is don't work on this right now, I'll discuss with other members and see which approach we will do.
1. Asking for devs to use the workaround;
2. make our API work with this bug.

maxkoshevoi: I agree that we shouldn't add a lot of code just to workaround the bug, but this workaround is not that big. It just adds one method:

image

The rest is just about making `StanckBar.Show` method return `Task` instead of `void`
image

So there is no logic change there, I basically just added `await`s where this method is called

pictos: We also make some methods `async`, that we discussed to not do in this [PR](https://github.com/xamarin/XamarinCommunityToolkit/pull/973)

maxkoshevoi: @pictos This satiation is a bit different. Currently we have one task that we return from our public `ShowSnackBar` method.

Since nothing in this method depends on that task, there are two ways to pass it to the caller:
- If `ShowSnackBar` is not `async`, `Task` can be returned from the method (current implementation)
- If `ShowSnackBar` is `saync`, `Task` can be `await`ed (chnage proposed in that PR)

Result is pretty much the same.

But is this PR we have two tasks, since `snackBar.Show` is now also returns a `Task`. So now we need to make `ShowSnackBar` `async` in order to `await` `snackBar.Show` and preserve execution flow.

AndreiMisiukevich (commented):
?
```csharp
// COMMENT EXPLAINING THE PURPOSE OF THIS METHOD
static async Task<IVisualElementRenderer?> GetRendererWithRetries(Page page, int retryCount = 5)
{
	var renderer = Platform.GetRenderer(page);
	if (renderer != null || retryCount <= 0)
		return renderer;

	await Task.Delay(50);
	return await GetRendererWithRetries(page, retryCount - 1);
}
```

As for ```Show``` method.
Can we preserve Return-Type to be ```void```?

Maybe we can put the code after GetRendererWithRetries call to ContinueWith?


```csharp
internal void Show(Page page, SnackBarOptions arguments) {
    GetRendererWithRetries(sender).ContinueWith(....);
}
```

Any thoughts?

maxkoshevoi:
> ```cs
> COMMENT EXPLAINING THE PURPOSE OF THIS METHOD
> ```

Good point. I'll add it

> Can we preserve Return-Type to be `void`?

@AndreiMisiukevich Can I ask why do you want to preserve it as `void`? This is internal API, so users won't notice it.
Anyway, `ContinueWith` won't help you here, since you still need to wait until task completes (otherwise method will return while task is still running). 

There are two options to preserve `void` return type and return when method executed to completion:
1) Replace `await` with `.Result`
2) Replace `Task.Delay` with `Thread.Sleep`

brminnick:
> Can we preserve Return-Type to be void?
@AndreiMisiukevich Can you explain why you need to preserve the `void` return type? 

@maxkoshevoi No, do not ever use `Thread.Sleep` or `.Result`

maxkoshevoi:
> @maxkoshevoi No, do not ever use `Thread.Sleep` or `.Result`

Yeah, it's bad to use them since they block the thread. But that's the only way to preserve `void` return type.
Just provided some options to do that =)

> If we do need to preserve the void return type, we'll use the .SafeFireAndForget() extension method.

It won't help here since
- It doesn't return result of the task (and we need result of `GetRendererWithRetries`)
- It doesn't wait until task completes, so it breaks existing execution flow

AndreiMisiukevich:
> > Can we preserve Return-Type to be void?
> 
> @AndreiMisiukevich Can you explain why you need to preserve the `void` return type?
> 

I thought it is related to public API

AndreiMisiukevich (commented):
Well, let's add a comment explaining what this method does, and then we will be able to get it merged, I believe. If there are no other objections.

And if you wish you can consider the "recursion-style" implementation of this method that I proposed above (up to you).

Thank you!

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.android.cs
maxkoshevoi: @pictos Interestingly, this doesn't fix the issue. 
```cs
static Task<IVisualElementRenderer> GetRenderer(Page page)
{
	TaskCompletionSource<IVisualElementRenderer> tcs = new ();

	var renderer = Platform.GetRenderer(page);
	if (renderer != null)
	{
		tcs.SetResult(renderer);
	}
	else
	{
		page.PropertyChanged += WatchForRendererToSet;
		void WatchForRendererToSet(object sender, PropertyChangedEventArgs args)
		{
			if (args.PropertyName == "Renderer")
			{
				page.PropertyChanged -= WatchForRendererToSet;
				tcs.SetResult(Platform.GetRenderer(page));
			}
		}
	}

	return tcs.Task;
}
```

Method itself works as expected, but SnackBar creation fails based on a renderer that is returned here `tcs.SetResult(Platform.GetRenderer(page));`

image

Anyway, fix using retries works, so leaving it as it.

maxkoshevoi:
@AndreiMisiukevich Added comment and changed implementation to recursive one.

AndreiMisiukevich (dismissed):

AndreiMisiukevich:
@pictos can you please take a look again?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.gtk.cs
```diff
@@ -13,7 +14,7 @@ class SnackBar
 	{
 		Timer? snackBarTimer;
 
-		public void Show(Page page, SnackBarOptions arguments)
+		public Task Show(Page page, SnackBarOptions arguments)
```
pictos, line 17: can we make this a `ValueTask`?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.ios.macos.cs
```diff
@@ -17,7 +17,7 @@ namespace Xamarin.CommunityToolkit.UI.Views
 {
 	class SnackBar
 	{
-		internal void Show(Page sender, SnackBarOptions arguments)
+		internal Task Show(Page sender, SnackBarOptions arguments)
```
pictos, line 20: can we make this a `ValueTask`?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.gtk.cs
maxkoshevoi: Return type should be the same for all versions of the method, so it should be `Task` everywhere or `ValueTask` everywhere (including android version).

Should I change it to `ValueTask` everywhere?

pictos: I believe so, the places that we just return the  `Task.Completed` will not be affected by Task allocation

maxkoshevoi: I was wrong. We can use `ValueTask` on some platforms, and `Task` on others because of the way `Show` method is compiled. I thought is was interface, or partial method, or something.
 
 Replaced `Task` with `ValueTask` everywhere except in android implementation.

pictos: @maxkoshevoi you can use a ValueTask on android as well, that way will be easier to maintain and the compiler will take care of it.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.android.cs
```diff
@@ -15,10 +16,10 @@ namespace Xamarin.CommunityToolkit.UI.Views
 {
 	class SnackBar
 	{
-		internal void Show(Page sender, SnackBarOptions arguments)
+		internal async Task Show(Page sender, SnackBarOptions arguments)
```
pictos, line 19:
```suggestion
		internal async ValueTask Show(Page sender, SnackBarOptions arguments)
```

pictos (commented):
just this little change

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.android.cs
maxkoshevoi: Done =)

pictos:
@maxkoshevoi thanks for the changes, I'll take a look today night or tomorrow during the day.

maxkoshevoi:
@pictos Just a friendly reminder that PR is ready for merge

pictos (approved):
@maxkoshevoi we have a TestPage folder inside the sample app, do you think that is possible to add a scenario there that proves that this PR fixes the issue? That isn't required to get this merged

maxkoshevoi:
@pictos It should be possible, but complicated. We would need to create mini-Shell environment with two pages inside our sample app

Issue #959, opened 2021-02-22
[Bug] NullReferenceException when displaying Toast or SnackBar


maxkoshevoi:
### Description

It's a bit tricky to reproduce, but I think it happens when you display toast and change page at the same time.

Here's how ViewModel displays the toast:
```cs
await Shell.Current?.CurrentPage?.DisplayToastAsync(LN.AutoupdateFailed);
```

And here's exception stack:
```cs
Xamarin.CommunityToolkit.UI.Views SnackBar.Show (Xamarin.Forms.Page sender, Xamarin.CommunityToolkit.UI.Views.Options.SnackBarOptions arguments)
Xamarin.CommunityToolkit.Extensions PageExtension.DisplayToastAsync (Xamarin.Forms.Page page, System.String message, System.Int32 durationMilliseconds)
```

`Shell.Current.CurrentPage` is not null before (since method gets called) and after (checked in `catch` block) `DisplayToastAsync` execution.

**Update**: It crashes on this line:
![image](https://user-images.githubusercontent.com/6609929/108738699-926c7100-753c-11eb-9e2b-75f9cfbb3397.png)
![image](https://user-images.githubusercontent.com/6609929/108744074-2db41500-7542-11eb-83ab-0f05ea12218c.png)

### Steps to Reproduce

1. Create Shell app with TabbBar and two tabs
2. Create button on one of the tabs that executes this code:
```cs
await Shell.Current.GoToAsync("RouteToOtherTab");
try
{
    await Shell.Current.CurrentPage.DisplayToastAsync("");
}
catch (Exception ex)
{
    await Shell.Current.CurrentPage.DisplayAlert("Error", ex.ToString(), "Ok");
}
```
3. Press the button

### Expected Behavior
Displaying toast or snackbar doesn't throw exception

### Actual Behavior
Displaying toast or snackbar throws exception

### Basic Information

- Version with issue: 1.0.3
- Last known good version: -
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android: 11
  - UWP:   
- Android Support Library Version: 
- Nuget Packages:
- Affected Devices:

### Workaround
Wrap every `DisplayToastAsync` and `DisplaySnackBarAsync` in try-catch

### Reproduction imagery

![xct3](https://user-images.githubusercontent.com/6609929/108751683-30673800-754b-11eb-8bb2-551de991ebf2.gif)

### Reproduction Link

[...]

PR #973, opened 2021-02-24
Fix DisplayToastAsync throws exception even if not awaited

### LINKED ISSUE #972: [Bug] DisplayToastAsync throws exception even if not awaited
### Description
Currently displaying toast or snackbar throws `NullReference` sometimes  #959 
Issue is that exception is still thrown even if `DisplayXxx` is used as ["fire and forget"](https://www.meziantou.net/fire-and-forget-a-task-in-dotnet.htm)

`MyPage.DisplayToastAsync("Some string").Forget();`

This is happening because we never actually use `await` inside `DisplayXxx` so state machine is not created.

### Steps to Reproduce

1. Make `DisplayToastAsync` throw some exception
2. Copy `Forget` method from the article above
3. Execute this `MyPage.DisplayToastAsync("Some string").Forget();`

### Expected Behavior
Exception is not thrown (handled by `Forget` method)

### Actual Behavior
Exception is thrown

### Basic Information

- Version with issue: 1.0.3
- Last known good version: -
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android: 11
  - UWP:   
- Nuget Packages:
- Affected Devices:

### Workaround
Add try-catch around `DisplayXxx` even when not awaiting it.

### Reproduction Link

Click "Show Toast" on SnackBar/Toast view here: https://github.com/maxkoshevoi/XamarinCommunityToolkit/tree/972-demo

maxkoshevoi:
### Description of Change ###

Added `await` to async page extensions, so the exceptions from them are thrown only if returned task is awaited.
This way those methods can be used in combination with ["fire and forget"](https://www.meziantou.net/fire-and-forget-a-task-in-dotnet.htm) methods.

### Bugs Fixed ###

- Fixes #972 

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


maxkoshevoi:
@pictos Could you take a look at this please?

pictos:
@brminnick can you review this one?

src/CommunityToolkit/Xamarin.CommunityToolkit/Extensions/PageExtension.shared.cs
```diff
@@ -10,7 +10,7 @@ namespace Xamarin.CommunityToolkit.Extensions
 {
 	public static class PageExtension
 	{
-		public static Task DisplayToastAsync(this Page page, string message, int durationMilliseconds = 3000)
+		public static async Task DisplayToastAsync(this Page page, string message, int durationMilliseconds = 3000)
```
VladislavAntonyuk, line 13: There are very few differences between the two approaches. Basically, they share the same semantics. However, the version with async/await wraps the execution of the inner task in an outer compiler-generated task. The non-async version does not. Thus, the non-async version is (very marginally) more efficient.

maxkoshevoi: Yes, it's more efficient, but please, read the issue description. Since we currently don't use `await` in those methods exceptions from them are not wrapped inside a task and are thrown immediately (even task is not awaited).
This is the problem, because we don't always want to await tasks. For example, there are "fire and forget" methods that accept that task, and that await the task inside try-chatch (and can perform some action on exception for example).

Those methods don't work with current version of "DisplayXxx" methods since we throw exception immediately and never even call "fire and forget" methos.

See repro steps and demo code in issue description.

VladislavAntonyuk (approved):

brminnick:
This doesn't change or improve the methods.

The two implementations are operationally the same, with returning `Task` (the existing implementation) being slightly more performant.

maxkoshevoi:
> The two implementations are operationally the same

@brminnick They are not the same though.. Could you please take a look to the issue description and issue demo? If we use `await` exceptions from those are wrapped inside the task that they are returning. If we don't use `await` exceptions are thrown immediately (even if we don't `await` the result from `DisplayXxx`). This makes them unusable with "fire are forget" methods.

brminnick:
Your solution is two-fold:
1. Don't pass in `null` values
2. Wrap your call to the API in a try/catch block:

```
try
{
    MyPage.DisplayToastAsync("Some string").Forget();
}
catch(ArgumentNullException e)
{
    /// Handle exception
}
```

maxkoshevoi:
> 1. Don't pass in `null` values

I don't pass `null` values, but method still throws `NullReferenceException` (#959), and `Forget` cannot handle it, because method is not wrapped into single task.

maxkoshevoi:
@brminnick We also can resolve this issue like this: https://github.com/xamarin/XamarinCommunityToolkit/pull/973/commits/73f147ce89c6939c7515ba38fd49406217d0f4c1

This way argument validation exceptions will be thrown immediately, but exceptions from the method itself will be thrown only when returned task is awaited

PR #987, opened 2021-02-25
Fix LocalizedString gets unsubscribed while still in use

### LINKED ISSUE #986: [Bug] 1.1.0-pre1 LocalizedString gets unsubscribed while still in use
### Description

`LocalizationResourceManager` inherits `ObservableObject` that uses `DelegateWeakEventManager`
`LocalizedString` WeekSubscribe. Turns out week events don't mix together with with week subscriptions  😄

### Steps to Reproduce

See new test in PR

### Expected Behavior
`LocalizedString` is always subscribed while in use

### Actual Behavior
`LocalizedString` gets unsubscribed

### Basic Information

- Version with issue: 1.1.0-pre1
- Last known good version: -
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android:  
  - UWP:   
- Nuget Packages:
- Affected Devices:

### Workaround
Implement your own `LocalizedString` without week subscription

maxkoshevoi:
### Description of Change ###

- Replaced event subscribtion in `LocalizedString` to regular one instead of `WeekSubscribe`
- Added test to validate the change. If you run this test on current version of the code, it fails.

Deeper understanding of the issue. Flow of the execution:
- Subscriber subscribes using `WeekSubscribe` => Nothing references object that subscribed to the event except even itself
- `WeakEventManager` stores subscribtion as `WeekReference`
- GC is invoked, and since nothing references subscribtion (it's stored as `WeekReference`), it is removed from the heap (only subscribtion is destroyed, subscriber still exists if someone references it)
- Event is triggered.  `WeakEventManager` sees that subscribtion is now `null` and unsubscribes it

TLDR: `WeekSubscribe` and `WeekEventManager` are both solution to the same problem, and only one of them is enough to solve it. If we use none - subscribtion will prevent object to be collected, if we use both - object will be unsubscribed after first GC run. If we use ether one of them - object will stay subscribed until someone references it, and be garbage collected when no one references it.

### Bugs Fixed ###

- Fixes #986

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###


maxkoshevoi:
@pictos Could you please take a look at this? It's 1.1.0 blocker

pictos:
@maxkoshevoi can you check the build errors?

maxkoshevoi:
> @maxkoshevoi can you check the build errors?

Some unrelated test has failed. Could you rerun pipeline on this PR please?

pictos:
Thinking aloud here, the WeakEventManager shouldn't dispose the propertyChange event if the LocalizationResourceManager is still alive. I'm saying that because we may have a root issue in another place.

maxkoshevoi:
The problem is that `WeakEventManager` works by unsubscribing dead subscribers. And is subscriber is subscribed using `WeekSubscribe`, nothing will reference it (it's the whole point of `WeekSubscribe`), so `WeakEventManager` will unsubscribe it.

`LocalizationResourceManager` won't be disposed since `Current` property always references it.

pictos:
@maxkoshevoi 
> The problem is that WeakEventManager works by unsubscribing dead subscribers. And if subscriber is subscribed using WeekSubscribe, nothing will reference it (it's the whole point of WeekSubscribe), so WeakEventManager will unsubscribe it.


 Didn't have time to debug your sample yet (I should do it at the end of the day tomorrow), instead, I debugged the `develop` branch, and I could see that there's is a direct reference to `LocalizationResourceManager.Current.PropertyChanged` that occurs at the app initialization, and every time that I change my language the code [ends here](https://github.com/xamarin/XamarinCommunityToolkit/blob/01bfcdf98b3c5a34bbad224e4da321f8bc842f4d/src/CommunityToolkit/Xamarin.CommunityToolkit/Helpers/WeakEventManagerService.shared.cs#L51) where the `WeakEventManager` check if the instance is alive.
If I'm not wrong that instance shouldn't be there if the object was collected by GC, and that occurs if nothing references it. And our Current instance is a `static` object so it shouldn't be collected.
I'll take a deeper look this weekend to see what else I can find.

maxkoshevoi:
> If I'm not wrong that instance shouldn't be there if the object was collected by GC

GC wasn't called yet, when you change the language. Add `GC.Collect()` call before you change the language like test does.
After first run of GC all `LocalizedStrings` that were created before it become unsubscribed (but only subscribtion is destroyed, `LocalizedString` still exists) 

xct6

maxkoshevoi:
@pictos I figured out a better solution! We no longer need `WeekSubscribe`. Since `LocalizationResourceManager` uses `WeekEventManager`, `LocalizedString` will be automatically unsubscribed if no one references it.

TLDR: `WeekSubscribe` and `WeekEventManager` are both solution to the same problem, and only one of them is enough to solve it. If we use none - subscribtion will prevent object to be collected, if we use both - object will be unsubscribed after first GC run.

So, do we need `WeekSubscribe` at all in this repository? Since we have `ObservableObject` that uses `WeekEventManager`, problem with objects not being garbage collected because of subscribtion is already solved.

pictos:
Great, this looks good. Thanks for this research and fix

pictos (approved):

AndreiMisiukevich (approved):

PR #1010, opened 2021-03-04
Implement a generic version of the EventToCommandBehavior

### LINKED ISSUE #953: [Enhancement] Generic EventToCommandBehavior
## Summary

Today with our current EventToCommandBehavior implementation we can see some issue using strongly type arguments with `Command<T>` the workaround today is using `T` as `object` and that's a kinda bad-sh thing to do.
If we have an `EventToCommandBehavior<TType>` that inherits from our `EventToCommandBehavior` we will be able to force the `parameter` to be the `TType` and doesn't crash the application.

An example of this issue can be seen here #950

## API Changes
Here's the full implementation so far.

```csharp
public class EventToCommandBehavior<TType> : EventToCommandBehavior
{
    protected override void OnTriggerHandled(object sender = null, object eventArgs = null)
    {
	    var parameter = CommandParameter
		    ?? EventArgsConverter?.Convert(eventArgs, typeof(object), null, null)
		    ?? eventArgs;
    
	    if (parameter.GetType() != typeof(TType))
	    {
		    parameter = null; // not sure if this a good approach, but nulling it we prevent an InvalidCastException (I suppose)
		    parameter = Convert.ChangeType(parameter, typeof(TType));
	    }
    
	    var command = Command;
	    if (command?.CanExecute(parameter) ?? false)
		    command.Execute(parameter);
    }
}
```

## Intended Use Case
Make this scenario works

```csharp
Coffee selectedCoffee;
public Coffee SelectedCoffee
{
    get => selectedCoffee;
    set => SetProperty(ref selectedCoffee, value);
}

async Task Selected(Coffee coffee)
{
    if (coffee == null)
        return;

    await Application.Current.MainPage.DisplayAlert("Selected", coffee.Name, "OK");
    SelectedCoffee = null;
}
```
On XAML the usage will be
```xml
<xct:EventToCommandBehavior
    x:TypeArguments="model:Coffee"
    Command="{Binding SelectedCommand}"
    EventArgsConverter="{StaticResource ItemSelectedEventArgsConverter}"
    EventName="ItemSelected" />
```

## Who Will Do The Work?

pictos:
### Description of Change ###

Make @jamesmontemagno happy :D

### Bugs Fixed ###

- Fixes #953

### API Changes ###

Added: 
 
```csharp
public class EventToCommandBehavior<TType> : EventToCommandBehavior
{
    protected override void OnTriggerHandled(object sender = null, object eventArgs = null)
    {
	    var parameter = CommandParameter
		    ?? EventArgsConverter?.Convert(eventArgs, typeof(object), null, null)
		    ?? eventArgs;
    
	    if (parameter?.GetType() != typeof(TType))
	    {
		    parameter = default(TType);  // not sure if this a good approach, but nulling it we prevent an InvalidCastException (I suppose)
		    parameter = Convert.ChangeType(parameter, typeof(TType));
	    }
    
	    var command = Command;
	    if (command?.CanExecute(parameter) ?? false)
		    command.Execute(parameter);
    }
}
```

### Behavioral Changes ###

Now you can use strong type in your Commands, something like that

```csharp

Coffee selectedCoffee;
public Coffee SelectedCoffee
{
    get => selectedCoffee;
    set => SetProperty(ref selectedCoffee, value);
}

async Task Selected(Coffee coffee)
{
    if (coffee == null)
        return;

    await Application.Current.MainPage.DisplayAlert("Selected", coffee.Name, "OK");
    SelectedCoffee = null;
}
```
On XAML the usage will be
```xml
<xct:EventToCommandBehavior
    x:TypeArguments="model:Coffee"
    Command="{Binding SelectedCommand}"
    EventArgsConverter="{StaticResource ItemSelectedEventArgsConverter}"
    EventName="ItemSelected" />
```

### PR Checklist ###


pictos:
@jamesmontemagno when you have time, can you try [this nugget](https://dev.azure.com/xamarin/6fd3d886-57a5-4e31-8db7-52a1b47c07a8/_apis/build/builds/36263/artifacts?artifactName=nuget&api-version=6.0&%24format=zip) on your coffee project? It should work

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/EventToCommandBehaviorGeneric.shared.cs
```diff
@@ -0,0 +1,29 @@
...
+		{
+			var parameter = CommandParameter
+				?? EventArgsConverter?.Convert(eventArgs, typeof(object), null, null)
+				?? eventArgs;
+
+			if (parameter?.GetType() != typeof(TType))
+			{
+				// nulling it to avoid a cast exception
+				parameter = null;
+				parameter = Convert.ChangeType(parameter, typeof(TType));
```
AndreiMisiukevich, line 21: in this case, TType should be a reference type, right?

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/EventToCommandBehaviorGeneric.shared.cs
```diff
@@ -0,0 +1,29 @@
...
+	/// <typeparam name="TType">The type that you want to receive in your <see cref="Xamarin.Forms.Command"/> </typeparam>
+	public sealed class EventToCommandBehavior<TType> : EventToCommandBehavior
+	{
+		protected override void OnTriggerHandled(object sender = null, object eventArgs = null)
+		{
+			var parameter = CommandParameter
+				?? EventArgsConverter?.Convert(eventArgs, typeof(object), null, null)
+				?? eventArgs;
+
+			if (parameter?.GetType() != typeof(TType))
```
AndreiMisiukevich, line 17: parameter = 2 | **int**
TType is **int?** (nullable int)

I presume it will return false

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/EventToCommandBehaviorGeneric.shared.cs
AndreiMisiukevich: it looks like parameter is always null here

pictos: No, isn't the parameter will be the type of `eventArgs`

```csharp
var parameter = CommandParameter
?? EventArgsConverter?.Convert(eventArgs, typeof(object), null, null)
?? eventArgs;
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/EventToCommandBehaviorGeneric.shared.cs
pictos: Good catch, let me try. But the TType will the whatever the user provides...

`EventToCommandBehavior<int> or EventToCommandBehavior<int?>`

Let me do the test and fix it if needed

jfversluis:
I know there has been some discussion on the tests, but this one seems to actually fail on a test in this change? 

```

  Failed Xamarin.CommunityToolkit.UnitTests.ObjectModel.ICommandTests.AsyncCommandTests.ICommand_AsyncCommandTests.ICommand_Parameter_CanExecuteChanged_DoesNotAllowMultipleExecutions_Test [1 ms]
  Error Message:
   Test execution timed out after 3000 milliseconds
  Stack Trace:
  --- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
```

brminnick:
I migrated the Unit Tests from xUnit to NUnit, but there are a couple tests that don't pass.

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/EventToCommandBehaviorGeneric.shared.cs
```diff
@@ -0,0 +1,29 @@
...
+	/// <typeparam name="TType">The type that you want to receive in your <see cref="Xamarin.Forms.Command"/> </typeparam>
+	public sealed class EventToCommandBehavior<TType> : EventToCommandBehavior
+	{
+		protected override void OnTriggerHandled(object? sender = null, object? eventArgs = null)
+		{
+			var parameter = CommandParameter
+				?? EventArgsConverter?.Convert(eventArgs, typeof(object), null, null)
+				?? eventArgs;
+
+			if (parameter?.GetType() != typeof(TType))
```
brminnick, line 17: Let's use C#9.0 Pattern Matching here to improve readability:

```csharp
if (parameter is not TType)
```

brminnick (changes requested):

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/EventToCommandBehaviorGeneric.shared.cs
pictos: Nice catch!

AndreiMisiukevich: @brminnick as I see here @pictos checks that type is exactly the same as TType.

**is** operator will also return **true** if the parameter type is a subtype of TType.

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/EventToCommandBehaviorGeneric.shared.cs
```diff
@@ -0,0 +1,29 @@
...
+		{
+			var parameter = CommandParameter
+				?? EventArgsConverter?.Convert(eventArgs, typeof(object), null, null)
+				?? eventArgs;
+
+			if (parameter is not TType)
+			{
+				// changing it to the default value to avoid a cast exception
+				parameter = default(TType);
+				parameter = Convert.ChangeType(parameter, typeof(TType));
```
AndreiMisiukevich, line 21: Can you please clarify why do we need to call ```Convert.ChangeType(parameter, typeof(TType))``` if parameter is already typeof TType here (cuz we set it in the line above)

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/EventToCommandBehaviorGeneric.shared.cs
brminnick: Good point.

@pictos Do you want to allow or prevent inherited types?

If we do want to allow inherited types, we should add a Unit Test for it. Something like this:

```csharp
[Test]
public void NoExceptionWhenTheEventArgsAreNotNull_InheritedType()
{
	var vm = new ViewModelCoffe();
	var behavior = new EventToCommandBehavior<Coffee>
	{
		EventName = nameof(ListView.ItemTapped),
		EventArgsConverter = new ItemSelectedEventArgsConverter(),
		Command = vm.SelectedCommand
	};

	Assert.Null(vm.CoffeeName);
	var coffe = new Latte { Id = 1, Name = "Latte" };
	var eventArgs = new SelectedItemChangedEventArgs(coffe, 1);

	var notNullArgs = new object?[] { null, eventArgs };

	TriggerEventToCommandBehavior(behavior, notNullArgs);

	Assert.AreEqual(coffe.Name, vm.CoffeeName);
}

class Latte : Coffee
{
}
```

AndreiMisiukevich (commented):
I propose a bit different approach.

Let's make our current EventToCommandBehavior generic and add logic there.

Then instead of duplicating the logic, we will subclass generic one with **object** type.


```csharp
public class EventToCommandBehavior : EventToCommandBehavior<object> 
{
}
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/EventToCommandBehaviorGeneric.shared.cs
pictos: I believe that this isn't needed anymore, I can't remember why I did it in first place, I'll try without it and see how it goes, if works I push a new version.

AndreiMisiukevich: Also, please consider my suggestion below)

brminnick:
> ```cs
> public class EventToCommandBehavior : EventToCommandBehavior<object> 
> {
> }
> ```

This is definitely a cleaner approach, and it would match how we've implemented other classes like  `class PopupDismissedEventArgs : PopupDismissedEventArgs<object?>`.

For nullability, we'd probably need to use `object?`:

```csharp
public class PopupDismissedEventArgs : PopupDismissedEventArgs<object?>
{
    public PopupDismissedEventArgs(object? result)
	: base(result)
    {
    }
}
```

pictos:
> I propose a bit different approach.
> 
> Let's make our current EventToCommandBehavior generic and add logic there.
> 
> Then instead of duplicating the logic, we will subclass generic one with **object** type.
> 
> ```cs
> public class EventToCommandBehavior : EventToCommandBehavior<object> 
> {
> }
> ```

Not sure if we can do it, since our non-generic implementation inherits from `BaseBehavior<VisualElement>`, so this change is softer than changing all the base logic.

brminnick:
> Not sure if we can easily do it

It's absolutely doable:

```csharp
using System;
using System.Reflection;
using System.Windows.Input;
using Xamarin.CommunityToolkit.Behaviors.Internals;
using Xamarin.Forms;
using Xamarin.Forms.Internals;

namespace Xamarin.CommunityToolkit.Behaviors
{
	/// <summary>
	/// This <see cref="EventToCommandBehavior"/> cast the sender object to a specific type defined by the user.
	/// </summary>
	/// <typeparam name="TType">The type that you want to receive in your <see cref="Xamarin.Forms.Command"/> </typeparam>
	public sealed class EventToCommandBehavior : EventToCommandBehavior<object?>
	{
		protected override void OnTriggerHandled(object? sender = null, object? eventArgs = null)
		{
			var parameter = CommandParameter
				?? EventArgsConverter?.Convert(eventArgs, typeof(object), null, null)
				?? eventArgs;

			var command = Command;
			if (command?.CanExecute(parameter) ?? false)
				command.Execute(parameter);
		}
	}

	/// <summary>
	/// The <see cref="EventToCommandBehavior"/> is a behavior that allows the user to invoke a <see cref="ICommand"/> through an event. It is designed to associate Commands to events exposed by controls that were not designed to support Commands. It allows you to map any arbitrary event on a control to a Command.
	/// </summary>
	public class EventToCommandBehavior<T> : BaseBehavior<VisualElement>
	{
		/// <summary>
		/// Backing BindableProperty for the <see cref="EventName"/> property.
		/// </summary>
		public static readonly BindableProperty EventNameProperty =
			BindableProperty.Create(nameof(EventName), typeof(string), typeof(EventToCommandBehavior), propertyChanged: OnEventNamePropertyChanged);

		/// <summary>
		/// Backing BindableProperty for the <see cref="Command"/> property.
		/// </summary>
		public static readonly BindableProperty CommandProperty =
			BindableProperty.Create(nameof(Command), typeof(ICommand), typeof(EventToCommandBehavior));

		/// <summary>
		/// Backing BindableProperty for the <see cref="CommandParameter"/> property.
		/// </summary>
		public static readonly BindableProperty CommandParameterProperty =
			BindableProperty.Create(nameof(CommandParameter), typeof(object), typeof(EventToCommandBehavior));

		/// <summary>
		/// Backing BindableProperty for the <see cref="EventArgs"/> property.
		/// </summary>
		public static readonly BindableProperty EventArgsConverterProperty =
			BindableProperty.Create(nameof(EventArgsConverter), typeof(IValueConverter), typeof(EventToCommandBehavior));

		readonly MethodInfo eventHandlerMethodInfo = typeof(EventToCommandBehavior).GetTypeInfo()?.GetDeclaredMethod(nameof(OnTriggerHandled)) ?? throw new NullReferenceException($"Cannot find method {nameof(OnTriggerHandled)}");

		Delegate? eventHandler;

		EventInfo? eventInfo;

		/// <summary>
		/// The name of the event that should be associated with <see cref="Command"/>. This is bindable property.
		/// </summary>
		public string? EventName
		{
			get => (string?)GetValue(EventNameProperty);
			set => SetValue(EventNameProperty, value);
		}

		/// <summary>
		/// The Command that should be executed when the event configured with <see cref="EventName"/> is triggered. This is a bindable property.
		/// </summary>
		public ICommand? Command
		{
			get => (ICommand?)GetValue(CommandProperty);
			set => SetValue(CommandProperty, value);
		}

		/// <summary>
		/// An optional parameter to forward to the <see cref="Command"/>. This is a bindable property.
		/// </summary>
		public object? CommandParameter
		{
			get => GetValue(CommandParameterProperty);
			set => SetValue(CommandParameterProperty, value);
		}

		/// <summary>
		/// An optional <see cref="IValueConverter"/> that can be used to convert <see cref="EventArgs"/> values, associated with the event configured with <see cref="EventName"/>, to values passed into the <see cref="Command"/>. This is a bindable property.
		/// </summary>
		public IValueConverter EventArgsConverter
		{
			get => (IValueConverter)GetValue(EventArgsConverterProperty);
			set => SetValue(EventArgsConverterProperty, value);
		}

		protected override void OnAttachedTo(VisualElement bindable)
		{
			base.OnAttachedTo(bindable);
			RegisterEvent();
		}

		protected override void OnDetachingFrom(VisualElement bindable)
		{
			UnregisterEvent();
			base.OnDetachingFrom(bindable);
		}

		static void OnEventNamePropertyChanged(BindableObject bindable, object oldValue, object newValue)
			=> ((EventToCommandBehavior)bindable).RegisterEvent();

		void RegisterEvent()
		{
			UnregisterEvent();

			var eventName = EventName;
			if (View == null || string.IsNullOrWhiteSpace(eventName))
				return;

			eventInfo = View.GetType()?.GetRuntimeEvent(eventName) ??
				throw new ArgumentException($"{nameof(EventToCommandBehavior)}: Couldn't resolve the event.", nameof(EventName));

			_ = eventInfo.EventHandlerType ?? throw new NullReferenceException();
			_ = eventHandlerMethodInfo ?? throw new NullReferenceException($"{nameof(eventHandlerMethodInfo)} is null, maybe it's a linker issue, please open a bug here: https://github.com/xamarin/XamarinCommunityToolkit/issues/");

			eventHandler = eventHandlerMethodInfo.CreateDelegate(eventInfo.EventHandlerType, this) ??
				throw new ArgumentException($"{nameof(EventToCommandBehavior)}: Couldn't create event handler.", nameof(EventName));

			eventInfo.AddEventHandler(View, eventHandler);
		}

		void UnregisterEvent()
		{
			if (eventInfo != null && eventHandler != null)
				eventInfo.RemoveEventHandler(View, eventHandler);

			eventInfo = null;
			eventHandler = null;
		}

		[Preserve(Conditional = true)]
		protected virtual void OnTriggerHandled(object? sender = null, object? eventArgs = null)
		{
			var parameter = CommandParameter
				?? EventArgsConverter?.Convert(eventArgs, typeof(object), null, null)
				?? eventArgs;

			if (parameter is not T)
			{
				// changing it to the default value to avoid a cast exception
				parameter = default(T);
				parameter = Convert.ChangeType(parameter, typeof(T));
			}

			var command = Command;
			if (command?.CanExecute(parameter) ?? false)
				command.Execute(parameter);
		}
	}
}
```

pictos:
The good part of having a strong type is to make sure that when the dev implements the `Command<T>`  he doesn't use a different type. Since we haven't good tooling to refactor XAML + C# my implementation is safer, making sure to match the Type between EventToCommandBehavior on XAML and the `T` type on command

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/EventToCommandBehaviorGeneric.shared.cs
```diff
@@ -0,0 +1,28 @@
...
+		{
+			var parameter = CommandParameter
+				?? EventArgsConverter?.Convert(eventArgs, typeof(object), null, null)
+				?? eventArgs;
+
+			if (parameter is not TType)
+			{
+				// changing it to the default value to avoid a cast exception
+				parameter = default(TType);
+			}
```
AndreiMisiukevich, line 21:
```suggestion
			// changing it to the default value to avoid a cast exception
			if (parameter is not TType)
				parameter = default(TType);
```

pictos: We're not using braces? I recall some recent PRs that we added the braces for one line

AndreiMisiukevich: I don't know really. Need to raise this question during stand-up )

AndreiMisiukevich:
@brminnick good example.
But I think that we even can skip overriding any method

```csharp
public class EventToCommandBehavior : EventToCommandBehavior<object?> 
{
}
```

it should be sufficient and  should work like a charm (because everything is object)

AndreiMisiukevich (dismissed):

brminnick (approved):

PR #1042, opened 2021-03-07
Fix Local Image is not working in AvatarView

### LINKED ISSUE #1025: [Bug] Local Image is not working in AvatarView
### Description
When AvatarView is used with LocalImage it displays Text value instead of image, while the same image is displayed in Xamarin.Forms.Image control.
### Steps to Reproduce

Sample : 
```
<xct:AvatarView Text="Avatar" Source="tab_about"/>
<Image Source="tab_about"/>
```

### Expected Behavior
Image should be displayed.
### Actual Behavior
Text value is displayed.
### Basic Information

- Version with issue: 1.0.3
- Last known good version:
- IDE: VS 16.7.8

AndreiMisiukevich:
### Description of Change ###
We shouldn't check if file exists for android local images, because images can be retrieved from context resources.
Also, we should provide the context to file the image handler on Android. That's why we pass Image to ImageSourceValidator, because we need to retrieve Context somewhere. 

### Bugs Fixed ###
- Fixes #1025

### API Changes ###
None

### Behavioral Changes ###
None

### PR Checklist ###


src/CommunityToolkit/Xamarin.CommunityToolkit/Views/AvatarView/ImageSourceValidator.android.ios.macos.uwp.wpf.gtk.tizen.cs
```diff
@@ -31,16 +32,20 @@ namespace Xamarin.CommunityToolkit.UI.Views
...
 		{
 			var handler = GetHandler(source);
 			if (handler == null)
 				return false;
-
 #if TIZEN
 			return await handler.LoadImageAsync(new NImage(XForms.NativeParent), source).ConfigureAwait(false);
 #elif MONOANDROID
-			var imageSource = await handler.LoadImageAsync(source, null).ConfigureAwait(false);
+			context ??= Platform.GetRenderer(image)?.View?.Context?.ApplicationContext;
```
pictos, line 47: Do we need the application context here? On develop we have a helper method to grab the Context, can you add a todo here, to us replace this with the helper implementation?

pictos (commented):
That looks good, I've just one question for you

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/AvatarView/ImageSourceValidator.android.ios.macos.uwp.wpf.gtk.tizen.cs
AndreiMisiukevich: The good question whether we should use ApplicationContext or not. I don't know) But it works.
Also, we retrieve context only once now (??= operator). Do you think we should get Context every time this method invoked?

pictos: No, because the ApplicationContext doesn't change... Also, on XF applications the ViewContext is the same, in most cases. Let move forward with this and if someone complains in the future we change it.

AndreiMisiukevich: @pictos sounds good

pictos (approved):
Friendly reminder that we can't merge any PR until we merge the nullable PR

pictos (approved):

PR #1053, opened 2021-03-10
allow TextCaseConverter to handle any value type

### LINKED ISSUE #1051: [Enhancement] Make TextCaseConverter support ANY type
## Summary

Currently the TextCaseConverter is pretty explicit about requiring that the value passed in must be null, a string, or a char... This is pretty silly as you may be passing in an Enum or a custom type with an overridden ToString. 

https://github.com/xamarin/XamarinCommunityToolkit/blob/caa3ebc71335d68e4a69758def542b322ccdbf90/src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs#L27-L30

## API Changes

Simply return the convert instead of type checking.

```cs
public object Convert(object value, Type targetType, object parameter, CultureInfo culture) 
 	=> Convert(value?.ToString(), parameter);
```

## Intended Use Case

You may want to use this with an Enum or other custom type which overrides ToString. The only implication would be with ConvertBack and since that throws a NotImplementedException it's a non-issue.

## Who Will Do The Work?

dansiegel:
### Description of Change ###

Updates the TextCaseConverter to allow any input type. Now Enum or other types can be passed in instead of only string's and char's.

### Bugs Fixed ###

- closes #1051

### API Changes ###

none

### Behavioral Changes ###

Instead of throwing an exception when a non-string value is passed to the TextCaseConverter it will simply call ToString and apply the proper casing.

### PR Checklist ###


dansiegel:
yeah I'll try to do that in the morning

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
```diff
@@ -37,19 +35,20 @@ public class TextCaseConverter : ValueConverterExtension, IValueConverter
...
 			TextCaseType.Upper => value?.ToUpperInvariant(),
 			TextCaseType.FirstUpperRestLower when value != null && !string.IsNullOrWhiteSpace(value) => value.Substring(0, 1).ToUpperInvariant() + value.Substring(1).ToLowerInvariant(),
-			_ => value
+			_ => value?.ToString()
 		};
 
-		TextCaseType GetParameter(object? parameter) => parameter == null ? Type : parameter switch
+		TextCaseType GetParameter(object? parameter) => parameter switch
 		{
+			null => Type,
```
AndreiMisiukevich, line 51: I guess the default should return "Type", not just "null".
Also, please make sure Type has TextCaseType.None as default value

dansiegel: @AndreiMisiukevich this isn't changing any actual behavior only simplifying this block of code as there is no point to check if `parameter` is `null` and then return `Type` otherwise go into a switch case for `parameter`...

AndreiMisiukevich: It seems "Type" property wasn't used before (what is wrong obviously)
And I proposed to use Type if parameter is not int or string or TextCaseType

dansiegel: oh I see what you're saying... fix the default case so it's returning Type instead of TextCaseType.None... yeah that makes more sense to me as well.

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
```diff
@@ -37,19 +35,20 @@ public class TextCaseConverter : ValueConverterExtension, IValueConverter
 		/// <param name="parameter">N/A</param>
 		/// <param name="culture">N/A</param>
 		/// <returns>N/A</returns>
-public object ConvertBack(object? value, Type? targetType, object? parameter, CultureInfo? culture)
+		public object ConvertBack(object? value, Type? targetType, object? parameter, CultureInfo? culture)
 			=> throw new NotImplementedException();
 
 		object? Convert(string? value, object? parameter) => GetParameter(parameter) switch
```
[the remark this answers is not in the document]
dansiegel: I've updated per your comments

pictos (approved):
Thanks for this! LGTM!

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/TextCaseConverter.shared.cs
```diff
@@ -37,27 +47,20 @@ public class TextCaseConverter : ValueConverterExtension, IValueConverter
...
-			TextCaseType.Lower => value?.ToLowerInvariant(),
-			TextCaseType.Upper => value?.ToUpperInvariant(),
-			TextCaseType.FirstUpperRestLower when value != null && !string.IsNullOrWhiteSpace(value) => value.Substring(0, 1).ToUpperInvariant() + value.Substring(1).ToLowerInvariant(),
-			_ => value
-		};
-
-		TextCaseType GetParameter(object? parameter) => parameter == null ? Type : parameter switch
+		TextCaseType GetParameter(object? parameter) => parameter switch
 		{
+			null => Type,
```
AndreiMisiukevich, line 55:
```suggestion
```

No need for this anymore, right?

AndreiMisiukevich (approved):

pictos (approved):

PR #1068, opened 2021-03-12
Image resource converter

### LINKED ISSUE #1002: [Enhancement] Add Bindings support for ImageResourceExtension
## Summary
With the actual ImageResource xaml extension [ImageResourceExtension.shared.cs](https://github.com/xamarin/XamarinCommunityToolkit/blob/main/src/CommunityToolkit/Xamarin.CommunityToolkit/Extensions/ImageResourceExtension.shared.cs) it is not possible to use Bindings.

What do you guys think about implementing a converter for this extension like it is already done for some of the converters?

~I have tried to derive from `BindableObject` to use bindable properties but the main issue is that `ProvideValue()`  will always fire first before the property changed event does, so it seems using Bindings with `IMarkupExtension` won't be straightforward.~

~As a solution:~
~- Create a new converter for the same purpose.~
~-  Or apply/update/resolve/manage the Bindings our-self inside `ProvideValue()` starting from a string property (Path & Source of Bindings) set by the dev (seems there is some approaches to it but they look complicated not practical IMHO https://stackoverflow.com/a/43866357/).~

## Who Will Do The Work?

Cfun1:
### Description of Change ###
Add an Image resource converter doing the same purpose as the ImageResourceExtension, but enable us to use Bindings.

### Bugs Fixed ###
- Fixes #1002

### API Changes ###

Added  ImageResourceConverter

### PR Checklist ###


src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/ImageResourceConverter.shared.cs
```diff
@@ -0,0 +1,37 @@
...
+		/// Converts embedded image resource ID to it ImageSource.
+		/// </summary>
+		/// <param name="value">The value to convert.</param>
+		/// <param name="targetType">The type of the binding target property. This is not implemented.</param>
+		/// <param name="parameter">Additional parameter for the converter to handle. This is not implemented.</param>
+		/// <param name="culture">The culture to use in the converter. This is not implemented.</param>
+		/// <returns>The ImageSource related to the provided resource ID of the embedded image. If it's null it will returns null.</returns>
+		public object? Convert(object value, Type targetType, object parameter, CultureInfo culture)
+		{
+			var imageId = (string)value;
```
AndreiMisiukevich, line 25: Can we avoid explicit cast here? It can cause cast-exception here.

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/ImageResourceConverter.shared.cs
```diff
@@ -0,0 +1,37 @@
...
+			return imageId == null
+					? null
+					: ImageSource.FromResource(imageId, Application.Current.GetType()
+#if NETSTANDARD1_0 || UAP10_0
+					.GetTypeInfo()
+#endif
+					.Assembly);
```
AndreiMisiukevich, line 32:
```suggestion
			if (value == null)
				return null;

			if (value is not string imageId)
				throw ArgumentException();

			return ImageSource.FromResource(imageId, Application.Current.GetType()
#if NETSTANDARD1_0 || UAP10_0
				.GetTypeInfo()
#endif
				.Assembly)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/ImageResourceConverter.shared.cs
Cfun1: Do you suggest replacing it with the "as" operator ?

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/ImageResourceConverter.shared.cs
Cfun1: @AndreiMisiukevich  I agree that with this way we have avoided `InvalidCastException` upon failed string cast, but also introduced `ArgumentException` when the type is not a string, does these two exceptions behave differently ? or just to be more explicit/precise about the exception thrown to developer ?

AndreiMisiukevich: The second one :) Just to provide more info on why the exception happened and that we aware of this exception.

samples/XCT.Sample/Pages/Extensions/ExtensionsGalleryPage.xaml
```diff
@@ -1,9 +1,15 @@
...
+                xmlns:pages="clr-namespace:Xamarin.CommunityToolkit.Sample.Pages">
+
+    <StackLayout Margin="30"
+                 Spacing="30">
+        <Label Text="Showing an embeded image by it resource Id using ImageResource extension"/>
+        <Image WidthRequest="150"
+               HeightRequest="150"
+               Source="{xct:ImageResource Id=Xamarin.CommunityToolkit.Sample.Images.logo.png}"/>
+    </StackLayout>
+</pages:BasePage>
```
AndreiMisiukevich, line 15: The gallery page should contain the list of items pointing to extension samples (similar to Converters gallery or behaviors one etc.)

samples/XCT.Sample/ViewModels/Converters/ImageResourceConverterViewModel.cs
```diff
@@ -0,0 +1,44 @@
...
+				imageName = value;
+				OnPropertyChanged();
+			}
+		}
+
+		ICommand? changeImageCommand;
+
+		public ICommand ChangeImageCommand => changeImageCommand ??= new Command(() =>
+			{
+				ImageName = ImageName!.Equals(BuildEmbededImagePath(img1)) ?
```
AndreiMisiukevich, line 30:
```suggestion
				ImageName = (ImageName?.Equals(BuildEmbededImagePath(img1)) ?? false) ?
```

samples/XCT.Sample/Pages/Extensions/ExtensionsGalleryPage.xaml
Cfun1: Oh my bad good catch, I didn't even noticed that thanks :)

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/ImageResourceConverter.shared.cs
```diff
@@ -0,0 +1,40 @@
...
+				return null;
+
+			if (value is not string imageId)
+				throw new ArgumentException("Value is not a string", nameof(value));
+
+			return ImageSource.FromResource(imageId, Application.Current.GetType()
+#if NETSTANDARD1_0 || UAP10_0
+				.GetTypeInfo()
+#endif
+				.Assembly)
```
AndreiMisiukevich, line 35:
```suggestion
				.Assembly);
```

samples/XCT.Sample/ViewModels/Converters/ImageResourceConverterViewModel.cs
```diff
@@ -0,0 +1,44 @@
...
+		string? imageName;
+
+		public string? ImageName
+		{
+			get => imageName;
+			set
+			{
+				imageName = value;
+				OnPropertyChanged();
+			}
```
AndreiMisiukevich, line 23:
```suggestion
			set => SetProperty(ref imageName, value);
```

samples/XCT.Sample/ViewModels/Converters/ImageResourceConverterViewModel.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Windows.Input;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Sample.ViewModels.Converters
+{
+	public class ImageResourceConverterViewModel : BaseViewModel
+	{
+		readonly string img1 = "button.png";
+		readonly string img2 = "logo.png";
```
AndreiMisiukevich, line 10:
```suggestion
		const string img1 = "button.png";
		const string img2 = "logo.png";
```

samples/XCT.Sample/ViewModels/Converters/ImageResourceConverterViewModel.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Windows.Input;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Sample.ViewModels.Converters
+{
+	public class ImageResourceConverterViewModel : BaseViewModel
+	{
+		const string img1 = "button.png";
+		const string img2 = "logo.png";
+		readonly string imagesPath = "Images";
```
AndreiMisiukevich, line 11:
```suggestion
		const string imagesPath = "Images";
		
```

samples/XCT.Sample/ViewModels/Extensions/ExtensionsGalleryViewModel.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Xamarin.CommunityToolkit.Sample.ViewModels.Effects
+{
+	public class ExtensionsGalleryViewModel : BaseGalleryViewModel
+	{
+		protected override IEnumerable<SectionModel> CreateItems() => new[]
+		{
+			new SectionModel(
+				typeof(ImageResourceExtensionPage),
+				nameof(ImageResourceExtension),
+				"The ImageResourceExtension is a xaml extension that will help to display an image from an embedded resource."),
```
AndreiMisiukevich, line 15:
```suggestion
				"The ImageResourceExtension is a XAML extension that will help to display an image from an embedded resource"),
```

samples/XCT.Sample/ViewModels/Extensions/ExtensionsGalleryViewModel.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Xamarin.CommunityToolkit.Sample.ViewModels.Effects
+{
+	public class ExtensionsGalleryViewModel : BaseGalleryViewModel
+	{
+		protected override IEnumerable<SectionModel> CreateItems() => new[]
+		{
+			new SectionModel(
+				typeof(ImageResourceExtensionPage),
+				nameof(ImageResourceExtension),
+				"The ImageResourceExtension is a XAML extension that will help to display an image from an embedded resource"),
```
AndreiMisiukevich, line 15:
```suggestion
				"A XAML extension that helps to display images from embedded resources"),
```

AndreiMisiukevich (approved):

AndreiMisiukevich:
@pictos could you please review?)

pictos (approved):

Issue #1075, opened 2021-03-14
[Bug] ICommand Unit Test Race Condition


brminnick:
### Description

There exists a race condition in the `IAsyncCommand` and `IAsyncValueCommand` unit tests that is causing the the tests to timeout.

The timeout happens in [tests like these](https://github.com/xamarin/XamarinCommunityToolkit/blob/b19a3fda461ef691b7a94da87a47088d530383e4/src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/ObjectModel/ICommandTests/AsyncCommandTests/AsyncCommand_Tests.cs#L409) where we are verifying that `ICommand.CanExecuteChanged` fires.

An example of the Unit Tests timing out can be seen in this CI build; the Unit Tests on .NET Framework never completed, causing the entire Unit Test suite to timeout after an hour:
https://github.com/xamarin/XamarinCommunityToolkit/pull/1005/checks?check_run_id=2100776335

Right now, this race condition appears to be only happening to Unit Tests running on .NET Framework; the Unit Tests running on .NET Core 2.1 and .NET Core 3.1 continue to pass successfully.

I'll start by adding a timeout to these unit tests to ensure they fail after a few seconds instead of running indefinitely; `[Fact(Timeout = 3000)]` will ensure it fails after 3 seconds, which should be more than enough time for it to complete successfully.

Once we now which tests are failing, I can dig deeper into, and fix, the race condition causing the failure.

brminnick:
Keeping this Issue open after merging #1076 after seeing the race condition reappear in this test run: https://dev.azure.com/xamarin/public/_build/results?buildId=37226&view=logs&jobId=7868e7c0-c859-51b7-c3be-08c48f6833d0&j=7868e7c0-c859-51b7-c3be-08c48f6833d0&t=774b7b69-0550-5cc8-416b-64d3f6dd2c9a

brminnick:
Currently investigating known issues with xUnit's implementation of SynchronizationContext:
https://github.com/xunit/xunit/issues/883

brminnick:
Confirmed xUnit’s implementation of SynchronizationContext was causing the race condition. 

Migrating to NUnit resolved the errors https://github.com/xamarin/XamarinCommunityToolkit/pull/1090

PR #1005, opened 2021-03-03
Snackbar/Toast Anchor

VladislavAntonyuk:
### Description of Change ###

Snackbar/Toast extension method for VisualElement

### Bugs Fixed ###

continuation of https://github.com/xamarin/XamarinCommunityToolkit/pull/846
- Fixes #476 

### API Changes ###

SnackBar/Toast is available for any VisualElement

### Behavioral Changes ###

Android:

image

iOS:

image

### PR Checklist ###

- [ ] Has tests (if omitted, state reason in description)
- [x] Has samples (if omitted, state reason in description)
- [x] Rebased on top of main at time of PR
- [x] Changes adhere to coding standard

- [ ] Updated [documentation](https://github.com/MicrosoftDocs/xamarin-communitytoolkit)

PR #1076, opened 2021-03-14
Resolve `IAsyncCommand` / `IAsyncValueValueCommand` Unit Test Race Condition

brminnick:
### Description of Change ###

This PR will resolve the race condition in the `IAsyncCommand` and `IAsyncValueCommand` unit tests that is causing the the tests to timeout.

The timeout happens in [tests like these](https://github.com/xamarin/XamarinCommunityToolkit/blob/b19a3fda461ef691b7a94da87a47088d530383e4/src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/ObjectModel/ICommandTests/AsyncCommandTests/AsyncCommand_Tests.cs#L409) where we are verifying that `ICommand.CanExecuteChanged` fires.

An example of the Unit Tests timing out can be seen in this CI build; the Unit Tests on .NET Framework never completed, causing the entire Unit Test suite to timeout after an hour:
https://github.com/xamarin/XamarinCommunityToolkit/pull/1005/checks?check_run_id=2100776335

Right now, this race condition appears to be only happening to Unit Tests running on .NET Framework; the Unit Tests running on .NET Core 2.1 and .NET Core 3.1 continue to pass successfully.

I'll start by adding a timeout to these unit tests to ensure they fail after a few seconds instead of running indefinitely; `[Fact(Timeout = 3000)]` will ensure it fails after 3 seconds, which should be more than enough time for it to complete successfully.

Once we now which tests are failing, I can dig deeper into, and fix, the race condition causing the failure.

### Bugs Fixed ###
Closes https://github.com/xamarin/XamarinCommunityToolkit/issues/1075

### Behavioral Changes ###

For now, there are no known-behavioral changes; this problem only exists in Unit Tests on .NET Framework

### PR Checklist ###

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -160,7 +160,7 @@ jobs:
   - job: build_macos
     displayName: Build macOS Library
     pool:
-      vmImage: macos-latest
+      vmImage: macos-10.15
     steps:
       # if this is a tagged build, then update the version number
       - powershell: |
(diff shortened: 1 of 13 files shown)


brminnick:
I think I've finally squashed the bug! 

It looks like the race condition was caused by using a `static` field in `BaseCommand`. 

That being said, because this a race condition bug, we _may_ see it pop up in the future if my fix of removing `static` wasn't the actual fix.

In case this race condition does happen again in a future PR, and because  happened only .NET Framework on Windows, I also added Unit Testing to our macOS builds in `azure-pipelines.yml`.

Having Unit Tests running on macOS will give us a useful datapoint for future PRs; if the Windows Unit Tests fail because of this race condition and the maOS Unit Tests pass, then we can be confident that the new code in the PR can be merged (i.e. the PR didn't introduce a new bug).

pictos (approved):
This looks great!

brminnick:
Adding Unit Testing to `azure-pipelines.yml` required writing some extra code because [xUnit doesn't support the command line command dotnet test for .NET Framework ](https://xunit.net/docs/getting-started/netcore/cmdline):

image

Long story short, this is the bash script that is required for us to run .NET Framework tests for xUnit on macOS (that is now included in `azure-pipelines.yml`:

```bash
mono "`find ~/.nuget/packages | grep net461 | grep xunit.console.exe | grep -v config`" "`find . -name Xamarin.CommunityToolkit.UnitTests.dll | grep bin | grep 461`"
```

PR #1090, opened 2021-03-17
Migrate Unit Tests to NUnit

### LINKED ISSUE #1083: [Enhancement] Migrate Xamarin.CommunityToolkit.UnitTests to NUnit
## Summary

This enhancement will migrate `Xamarin.CommunityToolkit.UnitTests` from xUnit (which it currently uses) to NUnit.

The reason for migrating to NUnit is two fold:
1. `Xamarin.CommunityToolkit.Markup.UnitTests` currently uses NUnit
  - Keeping the unit testing framework consistent will allow for a consistent experience for contributors and maintainers 
2. [xUnit doesn't support the command line command dotnet test for .NET Framework ](https://xunit.net/docs/getting-started/netcore/cmdline):

  - Long story short, this is the bash script that is required for us to run .NET Framework tests for xUnit on macOS (that is now included in [`azure-pipelines.yml`](https://github.com/xamarin/XamarinCommunityToolkit/blob/5a688543c8d6fe27c15bcecdbd91eeada1f79b36/azure-pipelines.yml#L211):
```bash
mono "`find ~/.nuget/packages | grep net461 | grep xunit.console.exe | grep -v config`" "`find . -name Xamarin.CommunityToolkit.UnitTests.dll | grep bin | grep 461`"
```

## Who Will Do The Work?

brminnick:
### Description of Change ###

This PR migrates `Xamarin.CommunityToolkit.UnitTests` from xUnit (which it currently uses) to NUnit.

The reason for migrating to NUnit is two fold:
1. `Xamarin.CommunityToolkit.Markup.UnitTests` currently uses NUnit
  - Keeping the unit testing framework consistent will allow for a consistent experience for contributors and maintainers 
2. [xUnit doesn't support the command line command `dotnet test` for .NET Framework ](https://xunit.net/docs/getting-started/netcore/cmdline):

  - Long story short, this is the bash script that is required for us to run .NET Framework tests for xUnit on macOS (that is now included in [`azure-pipelines.yml`](https://github.com/xamarin/XamarinCommunityToolkit/blob/5a688543c8d6fe27c15bcecdbd91eeada1f79b36/azure-pipelines.yml#L211):
```bash
mono "`find ~/.nuget/packages | grep net461 | grep xunit.console.exe | grep -v config`" "`find . -name Xamarin.CommunityToolkit.UnitTests.dll | grep bin | grep 461`"
```

### Bugs Fixed ###

- Fixes #1083 

### PR Checklist ###


src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Behaviors/UserStoppedTypingBehavior_Tests.cs
```diff
@@ -20,13 +20,13 @@ public async Task ShouldExecuteCommandWhenTimeThresholdHasExpired()
 
 			// act
 			entry.Text = "1";
-			await Task.Delay(defaultTimeThreshold + 100);
+			await Task.Delay(defaultTimeThreshold * 2);
```
bartlannoeye, line 23: This commit popped out in my notifications, so quick question:

- A 100ms difference should be granular enough to trigger the command. If not: is this wide range in possible execute timing an issue?
- Since we're using the same offset in every test, why not introduce a second parameter: testTimeTreshold?

brminnick: In a perfect world, yes an extra 100ms is enough time. However [`Task.Delay` isn’t a perfect Timer](https://stackoverflow.com/a/31126576/5953643), and it was causing these tests to fail randomly. Increasing the timeout guarantees we won’t get any false failures.

jfversluis (approved):

PR #1102, opened 2021-03-21
Implemented shadow on UWP

### LINKED ISSUE #847: [Enhancement] Support ShadowEffect on UWP
## Summary
Support `ShadowEffect` on UWP platform. Currently only supported by `Android/iOS/macOS` on #725.

## API Changes
No API change, add UWP implementation of `ShadowEffect`. Maybe using [DropShadow](https://docs.microsoft.com/en-us/uwp/api/windows.ui.composition.dropshadow).

## Intended Use Case
Use `ShadowEffect` on UWP platform.

## Who Will Do The Work?
- [x] Just putting this out there for someone to pick up

sattew:
### Description of Change ###

Well, I've implemented this effect via DropShadow (next DS). But DS has strange behavior. DS appears above the element and I decided to pack an element to StackLayout. 

This StackLayout has a structure: 
StackLayout
	> Grid (For shadow)
	> Element 

After these changes I appended shadow to Grid, and it's working :)

Maybe my solution isn't very good, but I think we have just one way to implement shadow on UWP. 

### Bugs Fixed ###

- Fixes #847

### API Changes ###

None

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
@@ -59,7 +59,7 @@ namespace Xamarin.CommunityToolkit.UWP.Effects
 
 		protected override void OnAttached()
 		{
-			if (!(Element is View elementView))
+			if (Element is not View elementView)
 				return;
 
 			switch (state)
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
@@ -1,0 +1,201 @@
+﻿using System;
+using System.ComponentModel;
+using System.Linq;
+using System.Numerics;
+using Windows.UI.Composition;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Hosting;
+using Windows.UI.Xaml.Shapes;
+using Xamarin.CommunityToolkit.Effects;
+using Xamarin.CommunityToolkit.UWP.Effects;
+using Xamarin.Forms;
+using Xamarin.Forms.Platform.UWP;
+using Grid = Xamarin.Forms.Grid;
+using Image = Windows.UI.Xaml.Controls.Image;
+
+[assembly: ExportEffect(typeof(PlatformShadowEffect), nameof(ShadowEffect))]
+
+namespace Xamarin.CommunityToolkit.UWP.Effects
+{
+	public class PlatformShadowEffect : PlatformEffect
+	{
+		enum ShadowEffectState
+		{
+			Initialized,
+			PanelCreated,
+			Attached
+		}
+
+		const float defaultRadius = 10f;
+
+		const float defaultOpacity = 1f;
+
+		ShadowEffectState state;
+
+		SpriteVisual? spriteVisual;
+
+		Layout<View>? shadowPanel;
+
+		DropShadow? shadow;
+
+		FrameworkElement View => Control ?? Container;
+
+		protected override void OnElementPropertyChanged(PropertyChangedEventArgs args)
+		{
+			base.OnElementPropertyChanged(args);
+
+			switch (args.PropertyName)
+			{
+				case nameof(ShadowEffect.ColorPropertyName):
+				case nameof(ShadowEffect.OpacityPropertyName):
+				case nameof(ShadowEffect.RadiusPropertyName):
+				case nameof(ShadowEffect.OffsetXPropertyName):
+				case nameof(ShadowEffect.OffsetYPropertyName):
+					UpdateShadow();
+					break;
+			}
+		}
+
+		protected override void OnAttached()
+		{
+			if (!(Element is View elementView))
+				return;
+
+			switch (state)
+			{
+				case ShadowEffectState.Attached:
+					return;
+
+				case ShadowEffectState.Initialized:
+					shadowPanel = new StackLayout()
+					{
+						Children = { new Grid() }
+					};
+					state = ShadowEffectState.PanelCreated;
+
+					MoveElementTo(elementView, shadowPanel);
+					return;
+
+				case ShadowEffectState.PanelCreated:
+					AppendShadow();
+					state = ShadowEffectState.Attached;
+					return;
+
+				default:
+					throw new ArgumentOutOfRangeException();
+			}
+		}
+
+		protected override void OnDetached()
+		{
+			if (state != ShadowEffectState.Attached)
+				return;
+
+			View.SizeChanged -= ViewSizeChanged;
+			shadow?.Dispose();
+			shadow = null;
+			spriteVisual?.Dispose();
+			spriteVisual = null;
+
+			state = ShadowEffectState.PanelCreated;
+		}
+
+		void AppendShadow()
+		{
+			var view = ElementCompositionPreview.GetElementVisual(View);
+
+			var compositor = view.Compositor;
+			shadow ??= compositor.CreateDropShadow();
+			UpdateShadow();
+			spriteVisual = compositor.CreateSpriteVisual();
+			spriteVisual.Shadow = shadow;
+			spriteVisual.Size = View.ActualSize;
+			spriteVisual.Size = new Vector2(1170, 40);
+
+			View.SizeChanged += ViewSizeChanged;
+
+			var renderer = shadowPanel?.Children.First().GetOrCreateRenderer();
+			spriteVisual.ParentForTransform = ElementCompositionPreview.GetElementVisual(View);
+			ElementCompositionPreview.SetElementChildVisual(renderer?.ContainerElement, spriteVisual);
+		}
+
+		void MoveElementTo(View element, Layout<View> to)
+		{
+			Device.BeginInvokeOnMainThread(() =>
+			{
+				if (Element.Parent is Layout<View> layout)
+				{
+					var index = layout.Children.IndexOf(element);
+					layout.Children.Insert(index, to);
+
+					if (layout is Grid)
+					{
+						var row = Grid.GetRow(element);
+						var rowSpan = Grid.GetRowSpan(element);
+						var column = Grid.GetColumn(element);
+						var columnSpan = Grid.GetColumnSpan(element);
+
+						Grid.SetRow(to, row);
+						Grid.SetRowSpan(to, rowSpan);
+						Grid.SetColumn(to, column);
+						Grid.SetColumnSpan(to, columnSpan);
+					}
+				}
+				else if (Element.Parent is ScrollView scrollView)
+				{
+					scrollView.Content = to;
+				}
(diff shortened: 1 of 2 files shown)


src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
```diff
@@ -0,0 +1,201 @@
...
+				case nameof(ShadowEffect.OffsetXPropertyName):
+				case nameof(ShadowEffect.OffsetYPropertyName):
+					UpdateShadow();
+					break;
+			}
+		}
+
+		protected override void OnAttached()
+		{
+			if (!(Element is View elementView))
```
pictos, line 62:
```suggestion
			if (Element is not View elementView)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
```diff
@@ -0,0 +1,201 @@
...
+		{
+			var view = ElementCompositionPreview.GetElementVisual(View);
+
+			var compositor = view.Compositor;
+			shadow ??= compositor.CreateDropShadow();
+			UpdateShadow();
+			spriteVisual = compositor.CreateSpriteVisual();
+			spriteVisual.Shadow = shadow;
+			spriteVisual.Size = View.ActualSize;
+			spriteVisual.Size = new Vector2(1170, 40);
```
pictos, line 114: What do those magic numbers?

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
```diff
@@ -0,0 +1,201 @@
...
+
+			View.SizeChanged += ViewSizeChanged;
+
+			var renderer = shadowPanel?.Children.First().GetOrCreateRenderer();
```
pictos, line 118: Are we ok to throw an exception if there are no Children?

pictos (commented):
@sattew thanks for this PR! Is good to see some love to UWP. Well I've some suggestions and questions for you.

Also I see that you wrap the content inside another layout (stack or grid) but when the use detaches the effect we don't revert that, we're adding more elements in the visual tree and that's bad for perf. So I would ask.
1. Can we use another approach? I mean on top of your head do you know another thing that we can do?
2. If you answer "no" for the previous question. Can we remove the extra layout on `OnDettached` method?

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
sattew: Oops, i tested smth and forget to delete

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
sattew: This can't happen. Because I initialized shadowPanel with one child (Grid) always

sattew:
> @sattew thanks for this PR! Is good to see some love to UWP. Well I've some suggestions and questions for you.
> 
> Also I see that you wrap the content inside another layout (stack or grid) but when the use detaches the effect we don't revert that, we're adding more elements in the visual tree and that's bad for perf. So I would ask.
> 
>     1. Can we use another approach? I mean on top of your head do you know another thing that we can do?
> 
>     2. If you answer "no" for the previous question. Can we remove the extra layout on `OnDettached` method?

1. I have no idea how we can do this any other way. Use ThemeShadow but this shadow doesn't have props to control (color, blur etc.)
2. Yes, we can remove extra layout, but it means we need to change page's structure again and we can lose in performance, I gues. For example if we have a lot of elements with shadow (In CollectionView or something else) and there will be a lot of operations to set the parent.

pictos:
> 1. I have no idea how we can do this any other way. Use ThemeShadow but this shadow doesn't have props to control (color, blur etc.)
2. Yes, we can remove extra layout, but it means we need to change page's structure again and we can lose in performance, I gues. For example if we have a lot of elements with shadow (In CollectionView or something else) and there will be a lot of operations to set the parent.

Yeah... you have a good point here on 2, so let's assume that if someone wants shadow, that person will not remove it during the usage, like something dynamic. I believe that we can do move forward with this implementation and see how it goes.

sattew:
> 
> 
> > 1. I have no idea how we can do this any other way. Use ThemeShadow but this shadow doesn't have props to control (color, blur etc.)
> 
>     1. Yes, we can remove extra layout, but it means we need to change page's structure again and we can lose in performance, I gues. For example if we have a lot of elements with shadow (In CollectionView or something else) and there will be a lot of operations to set the parent.
> 
> 
> Yeah... you have a good point here on 2, so let's assume that if someone wants shadow, that person will not remove it during the usage, like something dynamic. I believe that we can do move forward with this implementation and see how it goes.

let me some time also, I want to check one thing about question 1.

sattew:
Yeah. I checked again one thing and it's not working. I tried to add shadow to current parrent of element. But the shadow is still above element. Not working.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
```diff
@@ -0,0 +1,200 @@
...
+		{
+			base.OnElementPropertyChanged(args);
+
+			switch (args.PropertyName)
+			{
+				case nameof(ShadowEffect.ColorPropertyName):
+				case nameof(ShadowEffect.OpacityPropertyName):
+				case nameof(ShadowEffect.RadiusPropertyName):
+				case nameof(ShadowEffect.OffsetXPropertyName):
+				case nameof(ShadowEffect.OffsetYPropertyName):
```
AndreiMisiukevich, line 54: I spotted an issue with that (and exists on iOS and Android as well, but I fixed it in my PR)

https://github.com/xamarin/XamarinCommunityToolkit/pull/1109/files#r599173646

Please amend it for UWP

Also, Please add nameof(VisualElement.Width) and nameof(VisualElement.Height)

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
```diff
@@ -0,0 +1,200 @@
...
+					MoveElementTo(elementView, shadowPanel);
+					return;
+
+				case ShadowEffectState.PanelCreated:
+					AppendShadow();
+					state = ShadowEffectState.Attached;
+					return;
+
+				default:
+					throw new ArgumentOutOfRangeException();
```
AndreiMisiukevich, line 86:
```suggestion
				case ShadowEffectState.Initialized:
					shadowPanel = new StackLayout()
					{
						Children = { new Grid() }
					};
					
					state = ShadowEffectState.PanelCreated;
					MoveElementTo(elementView, shadowPanel);
					break;
				case ShadowEffectState.PanelCreated:
					AppendShadow();
					state = ShadowEffectState.Attached;
					break;
				default:
					break;
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
```diff
@@ -0,0 +1,200 @@
...
+			}
+		}
+
+		protected override void OnDetached()
+		{
+			if (state != ShadowEffectState.Attached)
+				return;
+
+			View.SizeChanged -= ViewSizeChanged;
```
AndreiMisiukevich, line 95: please check View for null

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
```diff
@@ -0,0 +1,200 @@
...
+			shadow = null;
+			spriteVisual?.Dispose();
+			spriteVisual = null;
+
+			state = ShadowEffectState.PanelCreated;
+		}
+
+		void AppendShadow()
+		{
+			var view = ElementCompositionPreview.GetElementVisual(View);
```
AndreiMisiukevich, line 106: need to check both View and View for null

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
```diff
@@ -0,0 +1,200 @@
...
+				TextBlock textBlock => textBlock.GetAlphaMask(),
+				Shape shape => shape.GetAlphaMask(),
+				Image image => image.GetAlphaMask(),
+				_ => shadow.Mask
+			};
+		}
+
+		void ViewSizeChanged(object sender, SizeChangedEventArgs e)
+		{
+			if (spriteVisual == null)
```
AndreiMisiukevich, line 192:
```suggestion
			if (spriteVisual == null || View == null)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
sattew: Can you tell me when View can be null, please. I only know that View can be null in ctor of effect.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
sattew: Maybe without default: break; ?

AndreiMisiukevich: Why not?) Do you expect that new values appear in this enum?

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
AndreiMisiukevich: That's "just in case" :)

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs
sattew: Definitely nope) I mean there is no code after switch)

AndreiMisiukevich: ah, gotcha) so yeah, both break and return can work here)

AndreiMisiukevich (approved):

Issue #1007, opened 2021-03-03
[Enhancement] Add Scale support for SideMenuView


dersia:
## Summary

The `SideMenuView` is a nice View and I think we can make it a bit more Fancy by adding a scale animation to the main-view.
I'd like to get something like what @ahoefling shows in this BlogPost: https://www.andrewhoefling.com/Blog/Post/xamarin-forms-fancy-flyout-menu

![Fancy Flyout](https://www.andrewhoefling.com/Portals/2/adam/Image%20&%20Lightbox/ZwoCOjUggEeMV168OpX1zQ/Image/fancy-menu-android.gif)

## API Changes

There is already a `Shift` property on the `SideMenuView` which is used to translate the menu with an animation, so we would only need a `Scale` property which is also taken into account when executing the animation.
This should be a small change and shouldn't have any implications.

## Intended Use Case
More fancy `SideMenuView`

## Who Will Do The Work?

- [ ] I am willing to take this on myself
- [x] Just putting this out there for someone to pick up

AndreiMisiukevich:
@dersia that's a great idea)
I thought about this enhancement. And definitely, we should implement it (it shouldn't be super difficult, I believe)

dersia:
Ok, can I also then just throw this into the pool: honestly we need this for pages not just views. Having a fancy flyout that supports pages makes the app look so much better.
I know this is a different proposal, but I wasn't sure if that's something that within the responsibility of the toolkit, so far there are no pages, just views ;)

ahoefling:
I am more than happy to contribute the fancy menu for this. Is the goal to implement this as an option on SideMenuView or make a new control for the toolkit?

AndreiMisiukevich:
> Ok, can I also then just throw this into the pool: honestly we need this for pages not just views. Having a fancy flyout that supports pages makes the app look so much better.
> I know this is a different proposal, but I wasn't sure if that's something that within the responsibility of the toolkit, so far there are no pages, just views ;)

Hmm, what the difference?)

AndreiMisiukevich:
> I am more than happy to contribute the fancy menu for this. Is the goal to implement this as an option on SideMenuView or make a new control for the toolkit?

I'll be happy to assign this enhancement to you :)
But I think that we should enhance the existing SideMenuView.

ahoefling:
@AndreiMisiukevich go ahead and assign this to me. I'll look at how I can upgrade the current SideMenuView to include scaling.

[...]

PR #1103, opened 2021-03-21
Adds StateChanged event to the SideMenuView

SkyeHoefling:
### Description of Change ###
Adds `StateChanged` event to the `SideMenuView`. This allows developers to listen for changes to the `State`. My goal with this is to trigger the event only when the `SideMenuView` is animated. 

This will provide the remaining feature set required to finish implementing the suggestion in #1007. With a `StateChanged` event listener you can add a corner radius if the sidemenu is visible and remove it when it is hidden.

### Bugs Fixed ###
N/A

### API Changes ###

Added: 
 
```csharp
// Triggered right after animation
event EventHandler<SideMenuStateChangedEventArgs> StateChanged;

// Triggered right before animation
event EventHandler<SideMenuStateChangingEventArgs> StateChanging;
```

Event Arguments 
```csharp
public class SideMenuChangedEventArgs
{
    public SideMenuState State { get; set; }
}
```

```csharp
public class SideMenuChangingEventArgs
{
    public SideMenuState State { get; set; }
}
```

### Behavioral Changes ###
N/A

### PR Checklist ###


src/CommunityToolkit/Xamarin.CommunityToolkit/Views/SideMenuView/SideMenuStateChangedEventArgs.shared.cs
```diff
@@ -0,0 +1,24 @@
...
+		/// <param name="state">
+		/// The new <see cref="SideMenuState"/> for the current <see cref="SideMenuView"/>.
+		/// </param>
+		public SideMenuStateChangedEventArgs(SideMenuState state) =>
+			State = state;
+
+		/// <summary>
+		/// Gets or sets the <see cref="SideMenuState"/>.
+		/// </summary>
+		public SideMenuState State { get; set; }
```
pictos, line 22:
```suggestion
		public SideMenuState State { get; }
```

pictos (commented):
Just a small change

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/SideMenuView/SideMenuStateChangedEventArgs.shared.cs
ahoefling: Makes sense to me, thanks for the feedback

ahoefling:
Maybe we should add 2 events? StateChanged would trigger after the animation and StateChanging could trigger right before? Does this make sense or should we leave it at just 1?

This PR was submitted because I was to add animation effects to the MainView while it is animating and just afterwards

ahoefling:
I just pushed my change to add the `StateChanging` event, I'll go an update the PR description

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/SideMenuView/SideMenuStateChangingEventArgs.shared.cs
```diff
@@ -0,0 +1,25 @@
...
+		/// <param name="state">
+		/// The new <see cref="SideMenuState"/> for the current <see cref="SideMenuView"/>.
+		/// </param>
+		public SideMenuStateChangingEventArgs(SideMenuState state) =>
+			State = state;
+
+		/// <summary>
+		/// Gets or sets the <see cref="SideMenuState"/>.
+		/// </summary>
+		public SideMenuState State { get; set; }
```
AndreiMisiukevich, line 23:
```suggestion
		public SideMenuState State { get; }
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/SideMenuView/SideMenuView.shared.cs
```diff
@@ -93,6 +94,29 @@ public SideMenuView()
 			#endregion
 		}
 
+		readonly WeakEventManager<SideMenuStateChangingEventArgs> stateChangingEventManager = new WeakEventManager<SideMenuStateChangingEventArgs>();
+		readonly WeakEventManager<SideMenuStateChangedEventArgs> stateChangedEventManager = new WeakEventManager<SideMenuStateChangedEventArgs>();
```
AndreiMisiukevich, line 98: Shouldn't we create a single weak event manager for all events like we do in other places?

```csharp
readonly WeakEventManager weakEventManager = new WeakEventManager();
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/SideMenuView/SideMenuView.shared.cs
```diff
@@ -332,13 +356,17 @@ void PerformUpdate(bool isAnimated = true)
 				SetOverlayViewInputTransparent(state);
 				return;
 			}
+
+			stateChangingEventManager.RaiseEvent(this, new SideMenuStateChangingEventArgs(State), nameof(StateChanging));
```
AndreiMisiukevich, line 360: events won't be called if isAnimated = false
Also, events will be called if the state wasn't changed (if the user drags menu a little and release the touch)

ahoefling: I think if `isAnimated = false` we should just trigger the `StateChanged` event. That will follow the intent of this PR where `StateChanged` is triggered after the animation or state has changed. I'll have updated this part in my next code push.

Your 2nd point about events firing even though the state hasn't changed is going to require a little more rules discussion. I just ran through it and it fires multiple events, yet they all return the same state. Maybe we can update the `State` Property to see if the new state is different than the previous state prior to applying it. 

Do you think it is best to simplify this API and just have it be `StateChanged`? Will trying to add `StateChanging` cause too many edge cases that will be difficult to maintain? In the gesture scenario you brought up there is no way for us to identify if the state is actually going to change, we can only make the assumption that the sidemenu is moving. Maybe we rename `StateChanging` to `MenuSliding` to indicate that the menu has started sliding but it hasn't completed. Another option is we could have a position delta event that is thrown every update cycle with a state. I have seen manipulation APIs like that in both WPF and UWP.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/SideMenuView/SideMenuView.shared.cs
ahoefling: This is a good idea, I have added it in my latest code push

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/SideMenuView/SideMenuView.shared.cs
```diff
@@ -318,6 +341,7 @@ void PerformUpdate(bool isAnimated = true)
 			{
 				TryUpdateShift(end, true, false);
 				SetOverlayViewInputTransparent(state);
```
AndreiMisiukevich, line 343: I propose to change the name of SetOverlayViewInputTransparent method and invoke **changED** event there (just to avoid code duplication).

Also, I propose to move **changING** event invocation before ```if (!isAnimated)``` check to be invoking every time.

pictos:
@ahoefling are you able to move this forward?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/SideMenuView/SideMenuView.shared.cs
ahoefling: As suggested I have renamed `SetOverlayViewInputTransparant` to `ChangeMenuState` which I think works. If you or other reviewers have strong opinions on a different name, let's come up with the best name.

I have moved the event invocation as recommend, please take another look

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/SideMenuView/SideMenuView.shared.cs
ahoefling: I think this is covered in the comment below - let me know if there is anything else we need to do on this comment

ahoefling:
@AndreiMisiukevich @pictos 
I have rebased this PR on the latest changes in the `develop` branch and made all the requested changes. It has been some time since I had looked at this code but everything appears to still be working. Let me know if there is anything else we need to do for this.

ahoefling:
I'm not sure if I understand why the build failed. I looked at the failing test and it doesn't appear to be in code that I modified in this PR. I am unable to reproduce the unit test failure in my local branch. Any advice on what might have caused it to fail?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/SideMenuView/SideMenuView.shared.cs
```diff
@@ -102,6 +103,28 @@ public SideMenuView()
 			#endregion
 		}
 
+		readonly WeakEventManager weakEventManager = new WeakEventManager();
```
AndreiMisiukevich, line 106:
```suggestion
		readonly WeakEventManager weakEventManager = new();
```

AndreiMisiukevich (dismissed):

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/SideMenuView/SideMenuStateChangedEventArgs.shared.cs
```diff
@@ -0,0 +1,25 @@
...
+		/// <param name="state">
+		/// The new <see cref="SideMenuState"/> for the current <see cref="SideMenuView"/>.
+		/// </param>
+		public SideMenuStateChangedEventArgs(SideMenuState state) =>
+			State = state;
+
+		/// <summary>
+		/// Gets or sets the <see cref="SideMenuState"/>.
+		/// </summary>
+		public SideMenuState State { get; }
```
pictos, line 23: makes sense to add an OldState and NewState, like the OnPropertyChanged pattern on BP?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/SideMenuView/SideMenuStateChangingEventArgs.shared.cs
```diff
@@ -0,0 +1,25 @@
...
+		/// <param name="state">
+		/// The new <see cref="SideMenuState"/> for the current <see cref="SideMenuView"/>.
+		/// </param>
+		public SideMenuStateChangingEventArgs(SideMenuState state) =>
+			State = state;
+
+		/// <summary>
+		/// Gets or sets the <see cref="SideMenuState"/>.
+		/// </summary>
+		public SideMenuState State { get; }
```
pictos, line 23: makes sense to add an OldState and NewState, like the OnPropertyChanged pattern on BP?

pictos (commented):
A small suggestion/question before moving this forward

DavidOrakpo:
Any update on this implementation? The statechanged event is still not in release

jfversluis:
Now that we're so close to the sunsetting of Xamarin (and therefore also Xamarin Community Toolkit) unfortunately we won't be able to take this in anymore, we're really sorry about that. Nevertheless, thank you so much for your time and effort that you have put into this PR, we really appreciate it.

Please have a look at the evolution of the Xamarin Community Toolkit, [The .NET MAUI Community Toolkit](https://github.com/CommunityToolkit/Maui). A lot of development has been going on there. Hopefully this issue was already fixed in that codebase. If not, feel free to port this over to there.

Again, thank you so much for being a contributor and Xamarin user!

PR #1106, opened 2021-03-22
Deprecate WeekSubscribe method due to high probability of misuse

maxkoshevoi:
### Description of Change ###

As discussed in #1001, deleted `WeekSubscribe` since it's most likely will be misused. 

~~It's not released yet, so this is not a breaking change.~~
This method is now released in 1.1.0. Maybe it's deletion should be reevaluated.

### Bugs Fixed ###

- Fixes #1001

### API Changes ###

Removed:
- `void WeakSubscribe<T>(this INotifyPropertyChanged target, T subscriber, Action<T, object?, PropertyChangedEventArgs> action)` 
 
### Behavioral Changes ###

None

### PR Checklist ###


brminnick:
We would need to first mark it as `[Obsolete]` in the next release.

Then we could remove it in the next major release.

pictos:
@brminnick oh yeah, We already released the 1.1 in a stable channel! I forgot it. @maxkoshevoi please restore that file and add the `Obsolete` attribute as suggested by @brminnick, thanks.

maxkoshevoi:
@pictos Done

src/CommunityToolkit/Xamarin.CommunityToolkit/ObjectModel/Extensions/INotifyPropertyChangedExtension.shared.cs
```diff
@@ -5,6 +5,7 @@ namespace Xamarin.CommunityToolkit.ObjectModel.Extensions
 {
 	public static class INotifyPropertyChangedExtension
 	{
+		[Obsolete("This method shouldn't be used week events (like one in ObservableObject). This method is deprecated due to high probability of misuse.")]
```
pictos, line 8:
```suggestion
        [EditorBrowsable(EditorBrowsableState.Never)]
		[Obsolete("This method shouldn't be used week events (like one in ObservableObject). This method is deprecated due to high probability of misuse.")]
```

pictos:
Thanks, @maxkoshevoi, now we just need to fix the CI to merge this puppy (:

maxkoshevoi:
@pictos This is not related to localization. `WeekSubscribe` to subscribe to any property change event.

pictos:
> @pictos This is not related to localization. `WeekSubscribe` to subscribe to any property change event.

but this isn't created because of `LocalizedString` feature? I didn't found any better label for that

maxkoshevoi:
> but this isn't created because of `LocalizedString` feature? I didn't find any better label for that

Yes, but it's still a general purpose method. 
Yeah, there isn't really a label for it 😄 What about `Breaking change`? Most users have `Treat warnings as errors set`, so this will be a breaking change for them.

pictos (approved):

PR #1127, opened 2021-03-26
[Converters] Added ListCountConverter

GUOLDEV:
### Description of Change ###

This PR added ListCountConverter.

### Bugs Fixed ###

- Fixes #

None

### API Changes ###

Added: 
 
- `void Class.Method();`
`ListCountConverter`

Changed:

```csharp
}
``` -->

Added: 

- `ListCountConverter`

### Behavioral Changes ###

None

### PR Checklist ###

### DIFF
--- a/samples/XCT.Sample/Pages/Converters/ListCountConverterPage.xaml
+++ b/samples/XCT.Sample/Pages/Converters/ListCountConverterPage.xaml
@@ -1,0 +1,45 @@
+﻿<?xml version="1.0" encoding="UTF-8" ?>
+<pages:BasePage
+    x:Class="Xamarin.CommunityToolkit.Sample.Pages.Converters.ListCountConverterPage"
+    xmlns="http://xamarin.com/schemas/2014/forms"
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by maxkoshevoi** [comment]
Cannot you just use 

```
{Binding Items.Count}
```

Instead of 

```
{Binding Items, Converter={xct:ListCountConverter}}
```

GUOLDEV:
Hi @maxkoshevoi, thanks for the feedback.
It is not an issue, it is a converter in case checking the number of elements in a collection is necessary and you do not need to check if `Items` is null for example. 
It reduces unnecessary null checks and returns what I want, just the number of elements.

maxkoshevoi:
Hi @GUOLDEV, every PR needs to have an issue associated with it (event one with new features).
Could you please create one, and let's move the discussion there.

When you are creating the issue there is a template for feature requests.

GUOLDEV:
Hi @maxkoshevoi, done! I have just created an [issue](https://github.com/xamarin/XamarinCommunityToolkit/issues/1128) for this PR.

PR #1130, opened 2021-03-26
Implement FullScreenEffect for Android

### LINKED ISSUE #718: [Enhancement] Provide an option to hide/unhide the status bar
## Summary

Provide an option to hide/unhide the status bar on some pages only, at Xamarin.Forms level, currently on Android one needs to use a `DependencyService` to achieve this. 

## API Changes
New effect to be used with `ContentPage` type?  `xct:StatusBarEffect.IsVisible="False"`

## Intended Use Case
My initial goal is to set a page in full screen with an attached effect to it.
Used besides `Shell.NavBarIsVisible="False"`  it will enable full screen mode, or do we need an extra effect `xct:StatusBarEffect.IsFullScreen="True"`
 What do you guys think?

## Who Will Do The Work?

Cfun1:
### Description of Change ###

Implement FullScreenEffect for shared and Android platform.

### Bugs Fixed ###
- Fixes #718

### PR Status ###
Code is still under work some parts could be removed/modified.
More importantly it needs to be made platform specific.

### PR Checklist ###


Cfun1:
It's still a draft and in a relative early stage, but if you guys have any remarks/review that needs to be pointed out, please don't hesitate.

maxkoshevoi:
I don't think it should be an effect for two reasons.
- It has only Android implementation, and there is a platform specific API for such things.
- It makes sense to be able to trigger it from code, which you cannot do with effect (except maybe setting it to `None` and to something else later)

Take a look at `NavigationBar` class in https://github.com/xamarin/XamarinCommunityToolkit/pull/812 for example of platform specific functionality.

Cfun1:
Hey @pictos I pushed an attempt to make it platform specific, but a part from folders/namespaces I still feel like it's not the good approach especially that `OnModeChanged()` will be duplicate on each platform implementation, your guidance is needed an appreciated :)

pictos:
> Hey @pictos I pushed an attempt to make it platform specific, a part from folders locations/namespaces I still feel like it's not the best approach especially that `OnModeChanged()` will be duplicate on each platform implementation, your guidance is needed an appreciated :)

@Cfun1 I'll take a look asap

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/FullScreen/AndroidSpecific/FullScreenEffect.shared.cs
```diff
@@ -0,0 +1,48 @@
...
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Effects.AndroidSpecific
+{
+	public class FullScreenEffect : Effects.FullScreenEffect
+	{
+		[Flags]
+		public enum FullScreenMode
+		{
+			Disabled = -1,
```
pictos, line 12: We need to change the notation here to use the `1 < 2` notation.

pictos (changes requested):
I took a look on my end, the implementation looks good to me, I would suggest moving files to another place.

`Xamarin.CommunityToolkit.Effects.AndroidSpecific.FullScreenEffect` should be moved to 
`PlatformConfiguration/AndroidSpecific/FullScreenEffect`, for the namespace follow the same pattern used by iOS and UWP.

Will this feature be supported on other platforms? IF yes ignore the next lines


Also, I saw that you used the effect API, but that isn't ideal, because we can't use it with this API
`On<Android>().FullScren()`, like we have [here](https://github.com/xamarin/XamarinCommunityToolkit/blob/c138f205148f544f398703e943222d0930f73125/samples/XCT.Sample/Pages/Views/Popups/TransparentPopup.xaml.cs#L14)

If you need help changing the API, please let me know

Cfun1:
@pictos thanks for your assistance, yes this feature will be supported for the other platforms but with specific full screen modes to each platform.
Although I am still not 100% sure if we should keep the Effect API in this case or try to completely replace it with the platform specific api, I think both approaches will work but not sure which one is the most appropriate one for this case.

pictos:
> @pictos thanks for your assistance, yes this feature will be supported for the other platforms but with specific full-screen modes to each platform.
> Although I am still not 100% sure if we should keep the Effect API in this case or try to completely replace it with the platform-specific API, I think both approaches will work but not sure which one is the most appropriate one for this case.

Can we match all platform configurations into a single Enum? Using the Android and UWP values? The iOS one will be Enabled if any value different from `Disabled` is set

Cfun1:
@pictos maybe I didn't understand you correctly, but I don't see how we can match them all in one single enum. What if the dev want to set the `Immersive` mode for Android and `PreferredLaunchViewSize` for UWP?

This is how I see the picture: In order to have a consistent behavior across platforms the dev must set a full screen mode for each platform, ie if the dev set only for Android `android:FullScreenEffect.Mode="ImmersiveDroid"` and don't set any mode for the other platforms it's their responsibility.

Maybe the way I set the table enumerating the modes in the linked issue could be confusing, in fact they are not grouped by the same behavior ie `Android.LeanBack` doesn't necessary mean it has the same behavior as `UWP.Maximized`

pictos:
> avior as

@Cfun1 I see now, said that I think you can move with your current implementation for enums

jfversluis:
Where are we on this one? Is it still going to happen? :)

Cfun1:
hey @jfversluis I will try to move forward with this during the starting week. (good to see you around :) )

Cfun1:
hey @pictos @jfversluis I have checked the current state of this pr, and I think it's ready to receive an initial review.

pictos:
@Cfun1 I'll take a look at this during this week. Thanks for the work (:

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/FullScreen/FullScreenEffectRouter.shared.cs
```diff
@@ -0,0 +1,19 @@
...
+	public class FullScreenEffectRouter : RoutingEffect
+	{
+		public FullScreenEffectRouter()
+			: base(EffectIds.FullScreenEffect)
+		{
+#if __ANDROID__
+			if (System.DateTime.Now.Ticks < 0)
+				_ = new Xamarin.CommunityToolkit.Android.Effects.FullScreenEffectRouter();
+#else
+			throw new NotImplementedException();
```
pictos, line 15: not sure if we want to throw an exception here...

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/FullScreen/PlatformFullScreenEffect.android.cs
```diff
@@ -0,0 +1,136 @@
...
+namespace Xamarin.CommunityToolkit.Android.Effects
+{
+	public class FullScreenEffectRouter : PlatformEffect
+	{
+		bool isDetaching;
+
+		protected override void OnAttached()
+		{
+			// should we use WeakEventManager instead?
+			(Element as Page)!.Disappearing += FullScreenEffectRouter_Disappearing;
```
pictos, line 22: Instead of casting Element as Page everywhere, let's just do:

```csharp
               Page PageElement => (Page)Element;

		protected override void OnAttached()
		{
			PageElement.Disappearing += FullScreenEffectRouter_Disappearing;
                        // ....
                 }
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/FullScreen/PlatformFullScreenEffect.android.cs
```diff
@@ -0,0 +1,136 @@
...
+			(Element as Page)!.Disappearing -= FullScreenEffectRouter_Disappearing;
+			isDetaching = true;
+			ResetStatusBar();
+		}
+
+		protected override void OnElementPropertyChanged(PropertyChangedEventArgs args)
+		{
+			base.OnElementPropertyChanged(args);
+
+			if (!args.PropertyName.Equals(ModeProperty.PropertyName) || isDetaching)
```
pictos, line 42: Even after `Detached` is executed, does this method fire?

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/FullScreen/PlatformFullScreenEffect.android.cs
```diff
@@ -0,0 +1,136 @@
...
+			{
+				DisableFullScreen();
+			}
+		}
+
+		// Still under work to handle different APIs
+		void EnableFullScreen(FullScreenMode mode)
+		{
+			var window = Container.Context.GetActivity()?.Window;
+			if (window is null)
```
pictos, line 82:
```suggestion
			if (window == null)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/FullScreen/PlatformFullScreenEffect.android.cs
```diff
@@ -0,0 +1,136 @@
...
+		{
+			var fullScreenMode = GetMode(Element);
+			var isPersistent = GetIsPersistent(Element);
+			if (fullScreenMode != FullScreenMode.Disabled && !isPersistent)
+			{
+				DisableFullScreen();
+			}
+		}
+
+		// Still under work to handle different APIs
```
pictos, line 78: Do we need more work here?

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/FullScreen/PlatformFullScreenEffect.android.cs
```diff
@@ -0,0 +1,136 @@
...
+			var window = Container.Context.GetActivity()?.Window;
+			if (window is null)
+			{
+				return;
+			}
+
+			// Supported in Android API > 17 (deprecated in API30 (Android 11) which is not supported by xct yet)
+			if (Build.VERSION.SdkInt >= BuildVersionCodes.JellyBeanMr1)
+			{
+				var view = window?.DecorView;
```
pictos, line 90: we already checked for null at line 82
```suggestion
				var view = window.DecorView;
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/FullScreen/PlatformFullScreenEffect.android.cs
```diff
@@ -0,0 +1,136 @@
...
+			var window = Container.Context.GetActivity()?.Window;
+			var view = window?.DecorView;
+			if (Build.VERSION.SdkInt >= BuildVersionCodes.JellyBeanMr1)
+			{
+				if (view != null)
+				{
+					view.SystemUiVisibility = (StatusBarVisibility)SystemUiFlags.Visible;
+					NavigationPage.SetHasNavigationBar(Element as Page, InitialHasNavigationBar);
+				}
+			}
```
pictos, line 127:
```suggestion
			if (Build.VERSION.SdkInt >= BuildVersionCodes.JellyBeanMr1 && view != null)
			{
				view.SystemUiVisibility = (StatusBarVisibility)SystemUiFlags.Visible;
				NavigationPage.SetHasNavigationBar(Element as Page, InitialHasNavigationBar);
			}
```

pictos (changes requested):
Thanks for this PR, sorry for taking so long to review it (:

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/FullScreen/FullScreenEffectRouter.shared.cs
Cfun1: This is temporary because this pr is implementing Android part only.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/FullScreen/PlatformFullScreenEffect.android.cs
Cfun1: I thought we have switched to pattern matching for null checks.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/FullScreen/PlatformFullScreenEffect.android.cs
Cfun1: Depends on what is the minimum Android API level that we should support ?

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/FullScreen/PlatformFullScreenEffect.android.cs
Cfun1: Yes I hit that case that why I added that field, but it only happened when I was in the middle of debugging and set the effect to "Disabled" on a page and try navigating to it initially or after having navigated to a page having IsPersistent on. After restarting the debug session this funny behavior is replaced by the fact that OnModeChanged is not called for that page since the default value is "Disabled".

So I suggest we add another enum value named "Default" just to ensure `OnModeChanged` of pages with effect value set to "Disabled" get triggered, it will be useful when the app is in full screen mode either by a persistent effect on a previous page or by something else, we want that to get triggered so that the effect disable the full screen mode on that page and restore it when disappears.

Cfun1:
@pictos any update on this?

pictos:
@Cfun1 sorry didn't have time to ship it in 1.2, but this will in 1.3 for sure :D , as soon we have a new version of XF I'll work on this

pictos:
@Cfun1 I took a look into this one and I believe that makes sense to implement this feature per platform using the `platform-specific` API. Do you agree?

brminnick:
@pictos It looks like we missed this one in the v1.3 Release.

Do you still want to merge it, or should we defer it to .NET MAUI Community Toolkit?

brminnick:

PR #1142, opened 2021-03-29
[Converter] Combined Converters

### LINKED ISSUE #553: [Enhancement] Converters modification
## Summary

There are converters which have return the value and the opposite value. They can be combined into 1 converter by using converter parameter.
e.g. IsNullOrEmptyConverter vs IsNotNullOrEmptyConverter. ListIsEmptyConverter vs ListIsNotEmptyConverter

## API Changes

1. remove *Not*Converters
2. use boolean parameters.

## Who Will Do The Work?

GUOLDEV:
Fixes #553 

Combined Converters IsNotNullOrEmptyConverter, ListIsNullOrEmptyConverter and ListIsNotNullOrEmptyConverter into IsNullOrEmptyConverter, implementing InvertCheck flag to invert the result.

If it is ok, IsNotNullOrEmptyConverter, ListIsNullOrEmptyConverter and ListIsNotNullOrEmptyConverter can be removed as all of them can be replaced using IsNullOrEmptyConverter with InvertCheck flag.

### Bugs Fixed ###

NA

### API Changes ###

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
``` -->
Changed:

- IsNullOrEmptyConverter
- IsNotNullOrEmptyConverter
- ListIsNullOrEmptyConverter
- ListIsNotNullOrEmptyConverter

### Behavioral Changes ###

NA

### PR Checklist ###

### CODE REVIEW

**Comment by GUOLDEV** [comment]
@Cfun1 thanks for the feedback, the PR targeting develop branch.

File: src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/IsNullOrEmptyConverter.shared.cs [hunk 1]
```diff
@@ -18,10 +24,25 @@ public class IsNullOrEmptyConverter : ValueConverterExtension, IValueConverter
...
-		public object Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture) => ConvertInternal(value);
+		public object Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture)
+			=> InvertCheck != ConvertInternal(value);
+
+		internal static bool ConvertInternal(object? value, bool isListCheck = false)
+		{
+			if (value == null)
+				return true;
+
+			if (value is IEnumerable list)
```
maxkoshevoi, line 35: String is an array of `chars`, so this `if` will be true for all strings. 
1) This means that  `value is string str` bellow is redundant
2) This method will return `false` for string that contains whitespace, while previously it returned `true` in such cases

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/IsNullOrEmptyConverter.shared.cs
```diff
@@ -10,6 +11,11 @@ namespace Xamarin.CommunityToolkit.Converters
 	/// </summary>
 	public class IsNullOrEmptyConverter : ValueConverterExtension, IValueConverter
 	{
+		/// <summary>
+		/// A value that specifies whether or not invert the result.
+		/// </summary>
+		public bool InvertCheck { get; set; }
```
maxkoshevoi, line 17: I think better name would be simply `Invert` or `InvertResult`

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/IsNullOrEmptyConverter_Tests.cs
```diff
@@ -1,15 +1,25 @@
...
+		public static IEnumerable<object?[]> GetData() => new List<object?[]>
+		{
+			new object?[] { null, true },
+			new object[] { string.Empty, true },
+			new object[] { "Test", false },
+			new object[] { typeof(IsNullOrEmptyConverter), false },
+			new object[] { new List<string>(), true },
+			new object[] { new List<string>() { "TestValue" }, false },
+			new object[] { Enumerable.Range(1, 3), false },
+		};
```
maxkoshevoi, line 20: Add test for whitespace string:
`new object[] { " ", true },`

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/IsNullOrEmptyConverter.shared.cs
```diff
@@ -18,10 +24,25 @@ public class IsNullOrEmptyConverter : ValueConverterExtension, IValueConverter
 		/// <param name="parameter">Additional parameter for the converter to handle. This is not implemented.</param>
 		/// <param name="culture">The culture to use in the converter. This is not implemented.</param>
 		/// <returns>A <see cref="bool"/> indicating if the incoming value is null or empty.</returns>
-		public object Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture) => ConvertInternal(value);
+		public object Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture)
+			=> InvertCheck != ConvertInternal(value);
```
maxkoshevoi, line 28: It's not obvious (at least to me) that `InvertCheck !=` inverts the result of `ConvertInternal`. I think inversion should be written more clearly for better readability.

```suggestion
		{
			var result = ConvertInternal(value);
			if (InvertCheck)
				result = !result;
			
			return result;
		}
```

maxkoshevoi (commented):
Thank you for the PR. Great idea!

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/IsNullOrEmptyConverter.shared.cs
GUOLDEV: Done!

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/IsNullOrEmptyConverter.shared.cs
GUOLDEV: Done!

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/IsNullOrEmptyConverter_Tests.cs
GUOLDEV: Done!

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/IsNullOrEmptyConverter.shared.cs
GUOLDEV: Done!

GUOLDEV:
Thanks for all the feedback!

GUOLDEV:
Hey @maxkoshevoi , is there something else that I need to do here?

pictos:
Closing this PR, as mentioned in this [issue](https://github.com/xamarin/XamarinCommunityToolkit/issues/553#issuecomment-816856684)

PR #1161, opened 2021-04-02
Allow a List or Array of Enums to be passed as ConverterParameters in…

solomonfried:
…stead of using TrueValues.

With the current implementation the TrueValues property is used when comparing against a list of Enums. This works, but it requires that I declare a new EnumToBoolConverter for each case when the compared enum list is different.

Using the code above, a user can declare an Array or List in Xaml and pass it as a ConverterParamater
for example. 
assuming an enum like;

namespace MyApp {
   public enum AppState
    {
        Active,
        Waiting,
        Paused,
        Stopped
    }
}
my xaml can contain a list

  xmlns:generic="clr-namespace:System.Collections.Generic;assembly=netstandard"
<ContentPage.Resources>
  <EnumToBoolConverter x:Key="EnumToBoolConverter" />

 <generic:List x:TypeArguments="MyApp:AppState" x:Key="ValidStates" >
            <testUiHelpers:AppState>Active</testUiHelpers:AppState>
            <testUiHelpers:AppState>Waiting</testUiHelpers:AppState>
  </generic:List>

   
</ContentPage.Resources>
and I can use it as follows

        <Label x:Name="EnumBoolTest" Text="Only for Valid States" FontSize="16" 
                   IsVisible="{Binding CurrentState, 
                Converter={StaticResource EnumToBoolConverter}, ConverterParameter={StaticResource ValidStates}}" />

### Bugs Fixed ###

- Fixes #

### API Changes ###

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
``` -->

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/EnumToBoolConverter.shared.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/EnumToBoolConverter.shared.cs
@@ -38,6 +38,35 @@ namespace Xamarin.CommunityToolkit.Converters
 			if (value is not Enum enumValue)
 				throw new ArgumentException("The value should be of type Enum", nameof(value));
 
+ 			if (TrueValues.Count == 0 && parameter != null && parameter is not Enum)
+            {
+                var valueType = value.GetType();
+				// Check if parameter is an Array
+                if (parameter is Array arrayParms)
+                {
+                    foreach (var arrayParm in arrayParms)
+                    {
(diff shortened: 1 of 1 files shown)


pictos:
@solomonfried can you please open an issue first? I see that you're a first-time contributor so maybe you should read our Contribution guideline first (:, you can find it [here](https://github.com/xamarin/XamarinCommunityToolkit/blob/main/CONTRIBUTING.md). You close this PR for now, we can reopen it in the future if it makes sense

PR #1172, opened 2021-04-05
Implement CompareConverter

### LINKED ISSUE #1149: [Enhancement] Add ConditionalDoubleConverter
## Summary

Add `ConditionalDoubleConverter` which seem to have similar name as existing `IntToBoolConverter` but the aim is totally different.

Input is a double and output is a bool, the difference with `IntToBoolConverter` is that the value of this bool depends on a condition set by the dev, <, >, ==, <=, >=, !=.

## API Changes
```
When ScrollY > 10 returns true otherwise false
<SomeView IsVisible="{Binding ScrollY, Converter={xct:ConditionalDoubleConverter Condition='>', Value=10}"/>

When ScrollY <= 10 returns true otherwise false
<SomeView IsVisible="{Binding ScrollY, Converter={xct:ConditionalDoubleConverter Condition='<=', Value=10}"/>
```
or better we could extend to more generic than `bool` like FalseObject TrueObject thing like in `BoolToObjectConverter`.

```
When ScrollY > 10 returns blue otherwise red
<SomeView Color="{Binding ScrollY, Converter={xct:ConditionalDoubleConverter Condition='>', Value=10, TrueObject= 'Blue', FalseObject= 'Red'}"/>
```
if `TrueObject` and `FalseObject` are not specified it will return the related bool.

## Intended Use Case
A real example could be used to implement a parallax effect, once the scroll value reach a certain threshold you want to set invisible a view. 

## Who Will Do The Work?

Cfun1:
### Description of Change ###
Added a new converter `CompareConverter`.

### Bugs Fixed ###
- Fixes #1149

### API Changes ###

Added a converter: `CompareConverter`

### PR Checklist ###

- [x] Updated documentation https://github.com/MicrosoftDocs/xamarin-communitytoolkit/pull/112


pictos:
@Cfun1 can you check the build errors? Thanks

Cfun1:
@pictos it's ready for review.

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/CompareConverter.shared.cs
```diff
@@ -0,0 +1,125 @@
...
+	public class CompareConverter : CompareConverter<object>
+	{
+	}
+
+	/// <summary>
+	/// Converts an object that implements IComparable to an object or a boolean based on a comparaison.
+	/// </summary>
+	public class CompareConverter<TObject> : ValueConverterExtension, IValueConverter
+	{
+		[Flags]
```
pictos, line 20: if we use flags, should we add the shift notation (1 << 0 )notation into the values?

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/CompareConverter.shared.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Globalization;
+using Xamarin.CommunityToolkit.Extensions.Internals;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Converters
+{
+	/// <summary>
+	/// Converts an object that implements IComparable to an object or a boolean based on a comparaison.
+	/// </summary>
+	public class CompareConverter : CompareConverter<object>
```
pictos, line 11:
```suggestion
	public sealed class CompareConverter : CompareConverter<object>
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/CompareConverter.shared.cs
```diff
@@ -0,0 +1,125 @@
...
+	{
+	}
+
+	/// <summary>
+	/// Converts an object that implements IComparable to an object or a boolean based on a comparaison.
+	/// </summary>
+	public class CompareConverter<TObject> : ValueConverterExtension, IValueConverter
```
pictos, line 18: Since we allow inheritance from this class should this be abstract?

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/CompareConverter.shared.cs
```diff
@@ -0,0 +1,125 @@
...
+
+		/// <summary>
+		/// Converts an object that implements IComparable to a specified object or a boolean based on a comparaison result.
+		/// </summary>
+		/// <param name="value">The value to convert.</param>
+		/// <param name="targetType">The type of the binding target property. This is not implemented.</param>
+		/// <param name="parameter">Additional parameter for the converter to handle. This is not implemented.</param>
+		/// <param name="culture">The culture to use in the converter.  This is not implemented.</param>
+		/// <returns>The object assigned to <see cref="TrueObject"/> if (value <see cref="ComparisonOperator"/> <see cref="ComparingValue"/>) equals True and <see cref="TrueObject"/> is not null, if <see cref="TrueObject"/> is null it returns true, otherwise the value assigned to <see cref="FalseObject"/>, if no value is assigned then it returns false.</returns>
+		public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
```
pictos, line 67: Since we allow inheritance from this class, let's make this `virtual`
```suggestion
		public virtual object Convert(object value, Type targetType, object parameter, CultureInfo culture)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/CompareConverter.shared.cs
```diff
@@ -0,0 +1,125 @@
...
+				Operator.GreaterOrEqual => EvaluateCondition(result >= 0),
+				Operator.Greater => EvaluateCondition(result > 0),
+				_ => throw new ArgumentNullException(nameof(ComparisonOperator), $"\"{ComparisonOperator}\" is not supported."),
+			};
+		}
+
+		object EvaluateCondition(bool v)
+			=> v ? mode == Modes.Object ?
+					TrueObject! : true
+						: mode == Modes.Object ? FalseObject! : false;
```
pictos, line 112: can you expand this to use normal `if/else`, it's hard to read

pictos (changes requested):
@Cfun1 thanks for this, I have some suggestions for you

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/CompareConverter.shared.cs
Cfun1: What we will gain from using the shift notation? I mean we are not going to do an "and" or an "or operation on flags, are we? the only thing is that we can build GreaterOrEqual  and  SmallerOrEqual  from: ` GreaterOrEqual = Greater OR Equal` and `SmallerOrEqual =Smaller OR Equal`, or i am missing something else ?

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/CompareConverter.shared.cs
Cfun1: I was taking BoolToObjectConverter as a reference, so not sure if this needs to be sealed or no, if we change it we should do the same for [BoolToObjectConverter](https://github.com/xamarin/XamarinCommunityToolkit/blob/26c9dda6ef1ccc84a0f1fa53134a6888bc1da449/src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/BoolToObjectConverter.shared.cs#L11) also.

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/CompareConverter.shared.cs
Cfun1: Could you provide a use case where this needs to be overridden?

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/CompareConverter.shared.cs
Cfun1: Could you provide an example where inheritance is required? otherwise I think w can make it sealed as you suggested in your previous comment.

pictos: you do it at line 11, so the dev can take that as inspiration and want to use some strongly typed type.

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/CompareConverter.shared.cs
pictos: It's needed, from the [guidelines](https://docs.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=net-5.0#guidelines-for-flagsattribute-and-enum)

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/CompareConverter.shared.cs
pictos: you do it at line 11, so the dev can take that as inspiration and want to use some strongly typed type, and for it the user maybe can override those methods to add his on implementation

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/CompareConverter.shared.cs
Cfun1: My bad I didn't got your point at the beginning (I thought combining flags that could be combined), sure I will update and give them specific values.

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/CompareConverter.shared.cs
```diff
@@ -0,0 +1,143 @@
...
+
+		/// <summary>
+		/// This method is not implemented and will throw a <see cref="NotImplementedException"/>.
+		/// </summary>
+		/// <param name="value">N/A</param>
+		/// <param name="targetType">N/A</param>
+		/// <param name="parameter">N/A</param>
+		/// <param name="culture">N/A</param>
+		/// <returns>N/A</returns>
+		public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
```
pictos, line 140:
```suggestion
		public virtual object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/CompareConverter.shared.cs
```diff
@@ -0,0 +1,143 @@
...
+	/// Converts an object that implements IComparable to an object or a boolean based on a comparaison.
+	/// </summary>
+	public sealed class CompareConverter : CompareConverter<object>
+	{
+	}
+
+	/// <summary>
+	/// Converts an object that implements IComparable to an object or a boolean based on a comparaison.
+	/// </summary>
+	public abstract class CompareConverter<TObject> : ValueConverterExtension, IValueConverter
```
pictos, line 18: @Cfun1 here, the `TObject` should implement `IComparable`?

Cfun1: Yes, you are thinking to add the condition `where TControl : IComparable` ?

pictos: Yes! You read my mind. I'll commit this change in a bit, I'm running here on my machine

pictos: But if I want to do 

` public MyConverter : CompareConvertert<MyModel>`

I need to make sure that `MyModel : IComparable` no?

pictos: So we should change the summary, because it says that object implements IComparable, or am I missing something?
"Converts an object that implements IComparable to an object or a boolean based on a comparaison."

Cfun1: The object that needs to implement `IComparable` is alue and `ComparingValue`, while `TObject` is the type of `TrueObject` and `FalseObject` which are the returned objects, no comparison is made on those ones so no need for `IComparable`.

pictos: I see now, thanks

pictos:
@Cfun1 I just fixed the Flags notations and simplified one if expression to use ternary notation. I believe that we are good to go with this one. Can you make a doc PR?

Cfun1:
@pictos I check "the Update documentation checkbox but forget to edit and add pr link to here, here is it https://github.com/MicrosoftDocs/xamarin-communitytoolkit/pull/112

pictos (approved):

PR #1181, opened 2021-04-07
Add F# Project

brminnick:
This PR adds `Xamarin.CommunityToolkit.Sample.FSharp.fsproj`.

In an earilier version of Xamarin.CommunityToolkit, it was reported that F# projects couldn't use the Xamarin.CommunityToolkit NuGet package: https://github.com/xamarin/XamarinCommunityToolkit/commit/e569ee3a2f4cabaffdcfad11edbc2ef495cbe51b

This project ensures that all future releases of Xamarin.CommunityToolkit will continue to be compatible with F#.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,12 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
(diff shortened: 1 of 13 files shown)


kevcrooks:
@brminnick is there a planned release we can use with this change included, as we'd like to reference the toolkit from our F# project?

PR #1200, opened 2021-04-14
Add DeviceCulture to get language that is currently selected on device 

### LINKED ISSUE #1201: [Enhancement] Add a way to get device's currently selected language
## Summary

XCT provides some great localization features that allow to change app language in runtime, but there is one gap.
If user changes devise language while app is running, there is no easy way to update app's language accordingly.

If user changes device language while app is running, `CultureInfo.CurrentCulture` and other ways to see current culture won't be updated, and continue to return device language that was set during app start.

Also there is no event that Xamarin app can subscribe to, but that's a minor issue since `App.OnStart` will be called on device language change anyway, so it would be great to be able to do something like this:

```cs
protected override void OnStart()
{
    LocalizationResourceManager.Current.CurrentCulture = DeviceInfo.CurrentCulture;
}
```

## API Changes

- `CultureInfo DeviceInfo.CurrentCulture { get; }`

## Who Will Do The Work?

maxkoshevoi:
### Description of Change ###

If user changes device language while app is running, `CultureInfo.CurrentCulture` and other ways to see current culture won't be updated, and continue to return device language that was set during app start.

Added multiplatform API to get language currently selected on device

### Bugs Fixed ###

- Fixes #1201

### API Changes ###

Added: 
 
```cs
static class DeviceInfo
{
    static IReadOnlyList<string> PreferredCultureStrings { get; }
}
```

### Behavioral Changes ###

None

### Testing ###
Tested on Android and UWP.
Could someone test it on iOS please?

### PR Checklist ###


src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/DeviceInfo/DeviceInfo.ios.macos.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Foundation;
+
+namespace Xamarin.CommunityToolkit.PlatformConfiguration.Multiplatform
+{
+	public static partial class DeviceInfo
+	{
+		static string CurrentCultureName => NSLocale.CurrentLocale.LanguageCode;
```
dimonovdd, line 7: This code will be more correct if you have correctly configured the localization in info.plist file

```csharp
NSBundle.MainBundle.PreferredLocalizations?.Length > 0
                ? NSBundle.MainBundle.PreferredLocalizations.First()
                : null;
```

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/DeviceInfo/DeviceInfo.android.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Java.Util;
+
+namespace Xamarin.CommunityToolkit.PlatformConfiguration.Multiplatform
+{
+	public static partial class DeviceInfo
+	{
+		static string CurrentCultureName => Locale.Default.Language;
```
dimonovdd, line 7: Localization requires not only the language but also the region

dimonovdd (commented):
You should also consider that you may get back a string that is not supported by `System.Globalization.CultureInfo`.

**Sample:**
- Android: `ar_EG_#u-nu-latn`
- iOS: `ar_AE@numbers=arab`

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/DeviceInfo/DeviceInfo.android.cs
maxkoshevoi: I initially tried using `Locale.Default.ToString().Replace('_', '-')` to get the region as well, but this was error prone. Not all android language-region pairs are supported by C#.

So I'm returning language only to definitely get valid `CultureInfo`.

Do people have different localization values for same language, but different regions?

dimonovdd: Yes my app uses different region settings for Arabic language. The numbers, calendar, date and time differ in different regions

dimonovdd:
And don't forget about the iOS features. 
We can change the language only in the app's system settings. We can't do this inside the app.

maxkoshevoi:
> You should also consider that you may get a string that is not supported by `System.Globalization.CultureInfo`.

Yeah, I tried to avoid that by returning only two letter language code. I can try convert language + region to `CultureInfo`, and if it will fail, than return language without region.

maxkoshevoi:
> And don't forget about the iOS features.
> We can change the language only in the app's system settings. We can't do this inside the app.

Yes, that's why this API is read only

dimonovdd:
And on Android, after changing the language in your app, you should definitely change the localization in the native app


> Yeah, I tried to avoid that by returning only two letter language code. I can try convert language + region to `CultureInfo`, and if it will fail, than return language without region.

This is also incorrect, in the future, the symbols located after the region will be important for my app

maxkoshevoi:
> in the future, the symbols located after the region will be important for my app

Can they be converted to `CultureInfo` somehow?

How about this: This API can return two values
- Native language string (with region and any other values).
- Parsed `CultureInfo` that may lack some info, but is guaranteed to have correct language.

So that users that need maximum precision, can parse platform specific value manually, and users with simpler use cases that just need a language can use already parsed value.

dimonovdd:
No you should have a list of supported localizations. And your unique rule for checking the native localization and supported by your app. Based on this rule and this list, you get an `CultureInfo `

dimonovdd:
The issue of localization is very complex, and it cannot be made convenient for everyone. A good article or video tutorial would be the best solution in this topic

maxkoshevoi:
> you should have a list of supported localizations

Makes sense. This will allow better localization experience even on first app start (initially I thought this API will be useful only when device language is changed while app is running) since currently `CultureInfo.CurrentCulture` is set to primary device culture regardless of what app supports.

Hm, this issue is more complicated than I thought.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/DeviceInfo/DeviceInfo.android.cs
CZEMacLeod: @maxkoshevoi  Even with 'en' language, both date and time formatting vary between regions.
As this returns a string, I think it should be left to the programmer to ensure that the culture is supported by .Net and what fallback they want to take.
It might be that the developer has defined custom cultures to match the android ones etc.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/DeviceInfo/DeviceInfo.ios.macos.cs
AndreiMisiukevich: Then just ```NSBundle.MainBundle.PreferredLocalizations?.FirstOrDefault();``` should work)

maxkoshevoi:
Ported https://github.com/xamarin/Essentials/pull/336 here.

Also made some changes to it:
- API returns all supported languages instead of just one
- API is fully customizable. Every step of `CultureInfo` parsing can be overridden by the user.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/DeviceInfo/DeviceInfo.ios.macos.cs
dimonovdd: in my case, throw is used

maxkoshevoi: @dimonovdd I updated implementation so that all preferred languages are returned.

PS: I think your solutions are identical. You check that `Length > 0`, so `First()` will never throw an exception

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/DeviceCulture/DeviceCulture.ios.macos.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System.Collections.Generic;
+using Foundation;
+
+namespace Xamarin.CommunityToolkit.PlatformConfiguration.Multiplatform
+{
+	public static partial class DeviceCulture
+	{
+		private static partial IReadOnlyList<string> GetPreferredCultureStrings() =>
```
dimonovdd, line 8: Why is this a list?

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/DeviceCulture/DeviceCulture.ios.macos.cs
```diff
@@ -0,0 +1,28 @@
...
+
+		class iOSCultureInfoParser : CultureInfoParser
+		{
+			protected override string ToDotnetFallbackLanguage(string languageCode)
+			{
+				languageCode = base.ToDotnetFallbackLanguage(languageCode);
+
+				var netLanguage = languageCode switch
+				{
+					"pt" => "pt-PT", // Fallback to Portuguese (Portugal)
```
dimonovdd, line 21: There are much more such cases

dimonovdd (commented):
If you still want to add this Then I advise you to change it.
1) delete everything 😂
2) Make a method that returns just a full platform culture string
3) Make a method for changing the culture on android

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/DeviceCulture/DeviceCulture.ios.macos.cs
maxkoshevoi: Do you know some examples? If so, I would be happy to add them. This method is meant to be expanded (as written in it's description)

dimonovdd: ar_EG_#u-nu-lat

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/DeviceCulture/DeviceCulture.ios.macos.cs
maxkoshevoi: Short answer: Because `NSLocale.PreferredLanguages` is a list
Long answer: Because user (device) can have several preferred languages, and if app doesn't support first one, it would be better to use second one instead of just using default (I think you mentioned it somewhere as well).

maxkoshevoi:
> I advise you to change it

Could you please clarify, what's wrong with current approach? Currently each platform has it's own implementation to provide device's preferred cultures. Even more so, common parsing code is abstracted, and end user can override any part of it.

Also current implementation provides native culture strings as well as their parsed `CultureInfo` equivalents, so user can choose what is more appropriate to use in their case.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/DeviceCulture/DeviceCulture.ios.macos.cs
dimonovdd: This is wrong. 
You must use the languages from the info.plist file. The first from the list will be that is selected in the app settings. I have provided you with the code aboveюб

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/DeviceCulture/DeviceCulture.ios.macos.cs
maxkoshevoi: This method only processes language code (there is different one for language + region). I think `ar` should be supported in .net. 
If `ar_EG` is not supported, could you please tell closest equivalent in .net?

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/DeviceCulture/DeviceCulture.ios.macos.cs
maxkoshevoi: Wait, this API is meant to get device's language, not application's

dimonovdd: Please read the articles about iOS localization. We should not use device localization

maxkoshevoi: This API is not about localization specifically. It's about getting current language of the device itself (irrespective of the language of the app). And then it can be used for any purposes.

If it's bad practice to use device language for localization on iOS, that users just won't use this API for localization purposes on iOS (platform check is pretty easy in Xamarin).

PS: Just curious, why shouldn't iOS users use device language?

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/DeviceCulture/DeviceCulture.android.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+using Java.Util;
+
+namespace Xamarin.CommunityToolkit.PlatformConfiguration.Multiplatform
+{
+	public static partial class DeviceCulture
+	{
+		private static partial IReadOnlyList<string> GetPreferredCultureStrings() =>
+			new[] { Locale.Default.ToString() ?? FallbackCulture };
```
dimonovdd, line 9: If The `Locale.Default` can be null, then use the null concatenation operator:
`Locale.Default?.ToString()`
I would return an empty list or null instead of `FallbackCulture`. Not everywhere English is the default language.

dimonovdd (commented):
I completely forgot about this PR. It looks good in my opinion. 
It would also be great if you could write a tutorial about changing the localization in the native part of an app.

src/CommunityToolkit/Xamarin.CommunityToolkit/PlatformConfiguration/Multiplatform/DeviceCulture/DeviceCulture.android.cs
maxkoshevoi: `Locale.Default` cannot be `null`, but `Locale.ToString()` can. It's possible that for `Locale.Default` value of `ToString()` will never be `null`.

What do you think  about using `toLanguageTag` here instead? Documentation [says](https://developer.android.com/reference/java/util/Locale#toString()):
> This behavior is designed to support debugging and to be compatible with previous uses of toString that expected language, country, and variant fields only. To represent a Locale as a String for interchange purposes, use `toLanguageTag()`.

brminnick:

Issue #894, opened 2021-02-12
[Feature Request] [CameraView] [iOS] Allow photos to be taken in landscape. 


JoelPickin:
### Description
I want to be able to take landscape photos and view them in landscape. I have tried two different routes to achieve this. Locking the iOS device orientation in portrait and rotating the phone to landscape causes the photo to be taken as a landscape photo but orientated as portrait. 

I have also tried creating a separate view for landscape so it uses this when the phone rotates but for some reason it rotates the camera as well as the UI so landscape photos are taken in inverse portrait. 

My explanation may not be the best but I have provided a repro to demonstrate this.

Also to note in the CommunityToolkit sample app, if the phone is rotated, the camera view is not rotated but the rest of the UI is. If you enter the CameraView in the sample app in landscape mode, the UI is rotated but the CameraView is not.

### Steps to Reproduce

1. Run the repro
2. Rotate the device
3. The landscape view will be used.
4. Camera is flipped to inverse portrait.

### Expected Behavior

1. Allow the camera to be used in landscape orientation when rotating the device
2. When a landscape orientation is used, have the camera rotated the same way.

### Actual Behavior

- CameraView not set to landscape doesn't resize to match the screen.
- CameraView in a landscape view flips the camera upside down, or inverse portrait.

### Basic Information

- Version with issue: 1.0.2
- IDE: Visual Studio 2019

### Reproduction Link

https://github.com/JoelPickin/CameraFlipBug

pictos:
This looks like more an `enhancement` than a `bug`.

michaldivis:
This would be useful

JoelPickin:
I have created a pull request for this feature on iOS

https://github.com/xamarin/XamarinCommunityToolkit/pull/1017

Thanks

brminnick:
Thanks! However, we are no longer adding new features to Xamarin Community Toolkit, focusing on the [.NET MAUI Community Toolkit](https://github.com/communitytoolkit/maui).

Please open a [New Feature Discussion](https://github.com/CommunityToolkit/Maui/discussions/new) to implement this feature in the .NET MAUI Community Toolkit.

I've posted more information about the Future Of Xamarin Community Toolkit here: https://devblogs.microsoft.com/xamarin/the-future-of-xamarin-community-toolkit/?WT.mc_id=mobile-0000-bramin

jrahma:
> This looks like more an `enhancement` than a `bug`.

[...]

PR #1224, opened 2021-04-22
iOS landscape rotation

JoelPickin:
### Description of Change ###

•	Register when the device is rotated by the OnElementPropertyChanged property of Height and Width.
•	When rotated set the bounds of the screen.
•	When rotated set the camera orientation of the capture connection.

### Bugs Fixed ###

This previously was a bug I raised myself but was understandably changed to a feature request.
#894

### API Changes ###

Added:
•	Added method in FormsCameraView.ios.cs: GetViewOrientation - Returns the AVCaptureVideoOrientation of the device.
•	Added method in FormsCameraView.ios.cs: SetCameraOrientation - Sets the previewLayer.Connection.VideoOrientation to the AVCaptureVideoOrientation from GetViewOrientation and then assigned the previewLayer.Connection to captureConnection.

Changed:
•	In method InitializeCamerain FormsCameraView.ios.cs added the line previewLayer.Connection.VideoOrientation = GetViewOrientation(); to set the correct orientation when the camera is being initialized.
•	In CameraViewRenderer.ios.cs - Added cases to switch in method OnElementPropertyChanged to check when Height & Width are changed ti trigger Control.SetBounds(Element.Width, Element.Height); & Control.SetCameraOrientation();

### Behavioral Changes ###

Rotating the device from landscape to the opposite landscape flips the camera view as the renderer still doesn't register the difference in rotation change. Rotating from portrait to landscape, then to portrait and landscape again works fine.

Images are captured in landscape also.

### PR Checklist ###

- [x] Has samples (if omitted, state reason in description) - Works in CameraView sample app


src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs
```diff
@@ -89,6 +89,32 @@ void SetStartOrientation()
 			}
 		}
 
+		public void SetCameraOrientation()
+		{
+			previewLayer.Connection.VideoOrientation = GetViewOrientation();
```
AndreiMisiukevich, line 94:
```suggestion
			if (previewLayer.Connection != null)
				previewLayer.Connection.VideoOrientation = GetViewOrientation();
				
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/iOS/FormsCameraView.ios.cs
```diff
@@ -517,6 +545,8 @@ void InitializeCamera()
 
 				captureSession.CommitConfiguration();
 
+				previewLayer.Connection.VideoOrientation = GetViewOrientation();
```
AndreiMisiukevich, line 548:
```suggestion
				if (previewLayer.Connection != null)
					previewLayer.Connection.VideoOrientation = GetViewOrientation();
```

jfversluis:
Great work @JoelPickin, thanks! I'm confused about this part:

> Rotating the device from landscape to the opposite landscape flips the camera view as the renderer still doesn't register the difference in rotation change. Rotating from portrait to landscape, then to portrait and landscape again works fine.

What does it mean? Reading the description it seems to work as you would expect? But the way you write the last part implies something different :D

jfversluis:
@JoelPickin awaiting your answer, thanks!

devopsoftware:
@JoelPickin  when this PR can be merged?

JoelPickin:
> Great work @JoelPickin, thanks! I'm confused about this part:
> 
> > Rotating the device from landscape to the opposite landscape flips the camera view as the renderer still doesn't register the difference in rotation change. Rotating from portrait to landscape, then to portrait and landscape again works fine.
> 
> What does it mean? Reading the description it seems to work as you would expect? But the way you write the last part implies something different :D

Apologies for the late response. Life and new a job got in the way :)

I'll try my best to explain it, if not I will get some screenshots if I get time. 

It works fine if you rotate from portrait to landscape. The camera flips and it resizes. If I remember correctly the issue was if you go from landscape, flip it 180 degrees to the opposite landscape, then the screen rotates, but the camera view doesn't. So you end up seeing a camera view upside down. It only registers in the code when there is a resizing of the screen to reset the camera and its orientation. If you rotate the device back to portrait, let the screen rotate, then rotate to landscape, the camera will show correctly in the correct orientation.

A future way of solving this may be to register the change in the orientation through the gyroscope, so it detects any movement and can reset the camera that way.

Hope that answers your question. I will be available much more now to answer any question as quick as possible.

GUOLDEV:
@jfversluis yes, #1524 should fix the rotation issue.

GUOLDEV:
@jfversluis @JoelPickin the CameraView Sample is fixed as well, so it should be easy to confirm if it works 👍

brminnick:
> versluis @JoelPickin the CameraView Sample is fixed as well, so it should be easy to confirm if it works 👍

@JoelPickin Could you confirm?

WaldemarCichon:
Seems some tests are failing and also a rebase a needed

WaldemarCichon:
I had compiled this changes but I get the following errors. How to avoid this (VS for Mac, newest version)

/Users/waldi/Projects/XamarinCommunityToolkit/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/iOS/CameraViewRenderer.ios.cs(5,5): Error CS0152: The switch statement contains multiple cases with the label value '"Height"' (CS0152) (Xamarin.CommunityToolkit)

The same with Width

Regards

Waldemar

bijington:
@WaldemarCichon we believe this issue should now be fixed in the latest release (1.3.0) but haven't been able to get confirmation from the author that it has been. 

Have you seen the same issue? If so are you able to confirm whether it has been fixed for us?

WaldemarCichon:
> @WaldemarCichon we believe this issue should now be fixed in the latest release (1.3.0) but haven't been able to get confirmation from the author that it has been.
> 
> Have you seen the same issue? If so are you able to confirm whether it has been fixed for us?

Hi bjington,

yes i was facing the same issue. Currently downloaded the 1.3.0 and it is gone. Hooray. Thanks to the implementer, thanks to the reports.

Regards

Waldemar

bijington:
@WaldemarCichon thabk you very much for confirming that the issue has been resolved. I shall close this issue down now.

@GUOLDEV big thanks for sorting this issue

bijington:
Resolved in #1534

PR #1238, opened 2021-04-26
Attempt to fix UWP target regression

jfversluis:
### Description of Change ###

TBD

### Bugs Fixed ###

- Fixes #

### API Changes ###

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
``` -->

### Behavioral Changes ###

### PR Checklist ###

### CODE REVIEW

**Comment by inforithmics** [comment]
Would it work when the Property would be retrieved by reflection FlyoutPresenter.IsDefaultShadowEnabledProperty
```
const string isDefaultShadowEnabledProperty = "IsDefaultShadowEnabledProperty";

if (ApiInformation.IsPropertyPresent(typeof(FlyoutPresenter).FullName, isDefaultShadowEnabledProperty))
{
	var property = typeof(FlyoutPresenter).GetProperty(isDefaultShadowEnabledProperty, BindingFlags.Static | BindingFlags.Public);
	if (property?.GetValue(null) is DependencyProperty enabledProperty)
	{
		flyoutStyle.Setters.Add(new Windows.UI.Xaml.Setter(enabledProperty,false));
	}
}
```

I tested it and it seems to work.

then it could target 1809 and work.

**Comment by jfversluis** [comment]
Hey @inforithmics could you please try the NuGet I reference in [this comment](https://github.com/xamarin/XamarinCommunityToolkit/issues/1171#issuecomment-826762273) to see if this change fixes it too? I think it does kind of the same thing, but different :)

**Comment by inforithmics** [comment]
I tested it and the problem ist still, that the MinVersion for the nuget packages is Windows 2004 / 19041

Maybe I should elaborate a little bit more about the problem.
My UWP Xamarin.Forms project has the following settings.

image

This means that it looks for uap10.0.17763 in the nuget. (The Minimum Version and not the target Version which is uap10.0.19041).

Your nuget pull request nuget package

image

Here it takes the the .netstandard2.0 package (which doesn't have the UWP Renderer and functionality in it).

My pull request nuget package.

image

Here it finds the uap10.0.17763 target.

For a fast test i added following line.

This compiles with my package but doesn't with yours.

`var test = new Xamarin.CommunityToolkit.UWP.Effects.LifeCycleEffectRouter();`

This is a type that is only avaible in the uwp target.

So the problem is that unlike android and ios the nuget target is taken from the min version not the target version that's where the problems with targeting a min version of 1809 stem from. And I don't think this could be changed fast.

So I added following change into my pull request in the project file.

from:
```
<TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);uap10.0.19041;netcoreapp3.1;net472;net471</TargetFrameworks>
```
to:
```
 <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);uap10.0.17763;netcoreapp3.1;net472;net471</TargetFrameworks>
    <TargetPlatformVersion Condition="'$(TargetFramework)' == 'uap10.0.17763'">10.0.19041.0</TargetPlatformVersion>   
```

So what does this, it sets the Min Target to 1809 / 17763 But sets the Target Version to  2004 / 19041.
This isn't something I invented but i copied it from the Xamarin.Forms project.

inforithmics:
Here a pull request against main with Multi targeting 1809 and 2004

https://github.com/xamarin/XamarinCommunityToolkit/pull/1247

inforithmics:
Seems to have worked there are now 2 Targets

image

The only caveat in Windows 1809 is that in the Popup transparent background won't work.
In Windows 2004 everything works as before.

inforithmics:
I tested the package and I discovered strange things.
1. When I reference the nuget package it takes with Min Version 1809 the .net standard Version if I target 2004 it takes the 2004 version.
2. When i directly reference the dll it works. I think there is some nuget/msuild strangnes going on.

inforithmics:
I found out what the problem was I targeted 1903 instead of 1809 (Version mixup)

here updated nuget:
https://dev.azure.com/xamarin/public/_build/results?buildId=39464&view=artifacts&pathAsName=false&type=publishedArtifacts

Issue #1129, opened 2021-03-26
AvatarView doesn't displays image


sunil12345678910:
We are using AvatarView in our Xamarin project. Previous version-1.0.3 works as expected on all platforms. Upgrading to latest version- 1.1.0 doesn't display source image only on UWP,  works on IOS and Android.

maxkoshevoi:
Related to: https://github.com/xamarin/XamarinCommunityToolkit/issues/998

jfversluis:
Thanks for reporting this @sunil12345678910 is there any specific reason you did not use the issue template for report a bug? That would really help us get all the information we need, please use it next time, thanks!

PR #1247, opened 2021-04-29
Multi Target UWP  to 1809 and 2004

inforithmics:
### Bugs Fixed ###

- https://github.com/xamarin/XamarinCommunityToolkit/issues/889

### Probably Fixes
- https://github.com/xamarin/XamarinCommunityToolkit/issues/1129

### API Changes ###

Adds Target Windows 17763 / 1809
Keeps Target Windows 19031 / 2004

### Behavioral Changes ###

works now again on Windows 17763 / 1809

### PR Checklist ###

       Has already tests
       Has already samples


inforithmics:
Don't merge yet, doesn't seem to work

inforithmics:
I picked the wrong uap version number now it should really target 1809 and 2004

inforithmics:
It works now
nuget package under

https://dev.azure.com/xamarin/public/_build/results?buildId=39464&view=artifacts&pathAsName=false&type=publishedArtifacts

jfversluis:
OK, so. This fixes the mentioned issues as it will sort out the SDK that needs to be targeted and it will automatically pick up on the shadows not being there for the popups for instance?

And this closes #942 and #1238 ?

inforithmics:
As I mentioned in.

https://github.com/xamarin/XamarinCommunityToolkit/pull/1238#issuecomment-829087936

When you Target for Min Windows 1809 Then the Transparent Background won't work because of this.
```
#if UWP_18362
if (Element.Color == Color.Transparent)
     flyoutStyle.Setters.Add(new Windows.UI.Xaml.Setter(FlyoutPresenter.IsDefaultShadowEnabledProperty, false));
#endif
```
The 1809 Target won't Compile the Code and therefore it won't be executed which means no Transparent Background for the Popup.

jfversluis:
Right, I noticed that compiler directive in there, yet it breaks right now for people targeting different versions if I understand correctly

inforithmics:
That's why I did the pull request https://github.com/xamarin/XamarinCommunityToolkit/pull/942
Where it Targets the 1809 But makes the newer properties available. But they need to be guarded by Api Checks

inforithmics:
Maybe Multi Targeting and deciding by case by case basis what to do is saver than having runtime crashes.

For example in the develop branch 

https://github.com/xamarin/XamarinCommunityToolkit/blob/develop/src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/PlatformShadowEffect.uwp.cs

View.ActualSize is used which is not availble under 1809 with compilation for 1809 and 2004 this would immediatly be detected.

Here I mentioned a possible Fix for 1809 with transparent background, But I have to this it First in Release (Linking Mode if it still works).

https://github.com/xamarin/XamarinCommunityToolkit/pull/1238#issuecomment-827402540

inforithmics:
I tested the changes and it works for an UWP App in the Debug Mode. But if I Native Compile it (Release) Popup does not work. Although without changes it doesn't work. I have to investigate further what the problem could be. Maybe the wrong Release compilation Settings.

inforithmics:
A fun fact I discovered is that the Xamarin.CommunityToolkit Nuget Package and Xamarin.CommunityToolkit Project behave differently in Version Resolution Management.

1. Original Nuget Targets 19031 / 2004 does not work in UWP App Targeting Min Version of 1809. So a Project using the nuget package doesn't work.
2. Original Project Targets  9031 / 2004 works in UWP App Targeting Min Version of 1809.  So the Test Project works.

inforithmics:
I removed the Release Compilation Crash Fixes from this Pull Request, because they are reviewed in this Pull Request.

https://github.com/xamarin/XamarinCommunityToolkit/pull/1303

YZahringer:
Any news about this? Our projects are blocked in XCT 1.0.3 because of this. Could be part of the 1.3.0 release?

inforithmics:
@jfversluis Can this be merged? or do I need to make some changes?.

I think this is the safer approach than 

https://github.com/xamarin/XamarinCommunityToolkit/pull/942

bijington:
@inforithmics thank you for keeping this PR up to date. To give a brief heads up we are currently unsure what the best course of action is in terms of supporting an older version of UWP and also the implications of targeting multiple versions.

I plan to investigate this over the next week or so in order to hopefully help us all come to a decision on whether this is something we should do. This also relates to #1684

bijington:
Changes made in #1684 were considered safer

Issue #355, opened 2020-10-02
Top level of CameraView receives wrong video file path [Bug] 


SABlyu:
### Description
when you capture a video file using CameraView, after the file gets data it's path is overwritten with newly created file path. After that the new file path is returned to the top level as it was the captured (with actual data) file path.

### Steps to Reproduce

1. start recording a video
2. stop recording a video
3. try to read the file content by the returned path (on CameraView MediaCaptured event)

### Expected Behavior
receive a file path with actually captured content
### Actual Behavior
receive a new file after session reinitialization

pictos:
@SABlyu can you attach a small project that reproduces this issue?

PR #1258, opened 2021-05-04
CameraView Video Recording Fix raising events with empty file on Android

danardelean:
### Description of Change ###
When recording Videos with the CameraView on Android the flow is wrong and after you call the DisposeMediaRecorder and PrepareSession you will actually raise and event hat has the registered file name as

### Bugs Fixed ###
https://github.com/xamarin/XamarinCommunityToolkit/issues/355

- Fixes #
Changing the flow when Stopping a video recorder.
Stop the recorder
Close the session
Raise the event with the correct file name
Restart the Session that will configure the new MediaRecorder

### API Changes ###

No changes

### Behavioral Changes ###
The CameraView will not raise events with wrong file names

### PR Checklist ###


src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs
```diff
@@ -531,9 +532,18 @@ public async Task StopRecord()
 			{
 				IsRecordingVideo = false;
 			}
+			try
+			{
+				CloseSession();
```
jsuarezruiz, line 537: `PrepareSession` will invoke `CloseSession`, is this neccesary?

danardelean: It isn't 
The only necessary thing is to raise the event with the file name before creating a new MediaRecorder which creates a new file and that name will be used for the event which resulted in a file which has 85 bytes in size as the "recorder file"
I also wonder if mediaRecorder.Prepare is necessary before the recording starts as it will actually create the file (85 bytes in size) before starting the recording when you are only in preview mode. The scenario where this will create problems is when you want to delete the files in the folder before starting to record a new one and you will endup deleting the file that media recorder has to use when you start the recording. I think Prepare should be called right before StartRecord but I haven't done any test on performance impact.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs
```diff
@@ -531,9 +532,18 @@ public async Task StopRecord()
...
+				//DisposeMediaRecorder(); //Already called by SetupMediaRecorder
```
jfversluis, line 538:
```suggestion
```

jfversluis (dismissed):
One small thing, let's just remove that commented line. Other than that seems good to me

AndreiMisiukevich (approved):

jfversluis:
@danardelean I think there is something weird with the CLA signing... Did you do it already? Would you mind doing/trying to do it again? :)

danardelean:
> Did you do it already? Would you mind doing/trying to do it again? :)

It does not work On my side it shows as Signed but I still see the button in this conversation. Any idea why? Thanks

jfversluis:
It's weird, I see it was signed and approved at some point. I have asked the question in the appropriate channels, just waiting for feedback there. Sorry it has to hang on something like this... What we could try in parallel maybe is open a new PR with the same changes, if your up for it try that :)

jfversluis:
Somehow it updated, screenshoting it here for proof :)

image

jfversluis:
Thanks for your contribution @danardelean !

acaliaro:
> Thanks for your contribution @danardelean !

Mitico @danardelean, It's nice to hear you again!!!

PR #1109, opened 2021-03-23
[iOS] Make RangeSlider more natural look like

AndreiMisiukevich:
### Description of Change ###
Update RangeSlider appearance (add shadow for iOS and macOS to be more natural look like)

Fixed shadow effect property changed bug

### API Changes ###
Changed default values of ThumbSize (30 -> 28) and TrackSize (3 -> 4).

### Behavioral Changes ###
Added shadow for thumbs on iOS and macOS

### PR Checklist ###
- [ ] Has tests (if omitted, state reason in description)
- [X] Has samples (if omitted, state reason in description)
- [X] Rebased on top of main at time of PR
- [X] Changes adhere to coding standard
- [ ] Updated [documentation](https://github.com/MicrosoftDocs/xamarin-communitytoolkit)

PR #1278, opened 2021-05-10
[Regression] Update shadow layer on property change 

AndreiMisiukevich:
### Description of Change ###
Fixes regression bug introduced after failed merge.
Initially, the fix was added in https://github.com/xamarin/XamarinCommunityToolkit/pull/1109/
but then was lost after an unsuccessful merge.

When shadow effect property changes or when element size changes, we should update the shadow layer. Otherwise, it will be lost.


pictos (approved):
LGTM!

PR #1293, opened 2021-05-14
Drawing View - Fix nullable, add doc to the public method

VladislavAntonyuk:
### Bugs Fixed ###

- Fixes #

### API Changes ###

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
``` -->

### Behavioral Changes ###

### PR Checklist ###

### DIFF
--- a/samples/XCT.Sample.WPF/MainWindow.xaml.cs
+++ b/samples/XCT.Sample.WPF/MainWindow.xaml.cs
@@ -14,7 +14,6 @@ namespace Xamarin.CommunityToolkit.Sample.WPF
 			InitializeComponent();
 			Forms.Forms.Init();
 			PancakeViewRenderer.Init();
-			DrawingViewRenderer.Init();
 			LoadApplication(new Xamarin.CommunityToolkit.Sample.App());
 		}
 	}
--- a/samples/XCT.Sample.WPF/MainWindow.xaml.cs
+++ b/samples/XCT.Sample.WPF/MainWindow.xaml.cs
@@ -14,6 +14,7 @@ namespace Xamarin.CommunityToolkit.Sample.WPF
 			InitializeComponent();
 			Forms.Forms.Init();
 			PancakeViewRenderer.Init();
+			DrawingViewRenderer.Init();
 			LoadApplication(new Xamarin.CommunityToolkit.Sample.App());
 		}
 	}
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Renderer/DrawingViewRenderer.android.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Renderer/DrawingViewRenderer.android.cs
@@ -20,8 +20,8 @@ namespace Xamarin.CommunityToolkit.UI.Views
 		readonly Paint canvasPaint;
 		readonly Paint drawPaint;
 		readonly Path drawPath;
-		Bitmap canvasBitmap;
-		Canvas drawCanvas;
+		Bitmap? canvasBitmap;
+		Canvas? drawCanvas;
 
 		public DrawingViewRenderer(Context context)
 			: base(context)
@@ -67,16 +67,16 @@ namespace Xamarin.CommunityToolkit.UI.Views
 		{
 			base.OnSizeChanged(w, h, oldw, oldh);
 
-			canvasBitmap = Bitmap.CreateBitmap(w, h, Bitmap.Config.Argb8888);
+			canvasBitmap = Bitmap.CreateBitmap(w, h, Bitmap.Config.Argb8888!)!;
 			drawCanvas = new Canvas(canvasBitmap);
 			LoadPoints();
 		}
 
-		protected override void OnDraw(Canvas canvas)
+		protected override void OnDraw(Canvas? canvas)
 		{
 			base.OnDraw(canvas);
 
-			canvas.DrawBitmap(canvasBitmap, 0, 0, canvasPaint);
+			canvas!.DrawBitmap(canvasBitmap!, 0, 0, canvasPaint);
 			canvas.DrawPath(drawPath, drawPaint);
 		}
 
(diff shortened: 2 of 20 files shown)

### CODE REVIEW

File: samples/XCT.Sample.WPF/MainWindow.xaml.cs [hunk 1]
```diff
@@ -14,6 +14,7 @@ public MainWindow()
 			InitializeComponent();
 			Forms.Forms.Init();
 			PancakeViewRenderer.Init();
+			DrawingViewRenderer.Init();
```
pictos, line 17: Sadly this a `no-go`. We can't have any `Init` call on XCT

pictos (changes requested):
@VladislavAntonyuk Thanks for this! I'll not merge yet because of that small observation, I'll wait for your response. Thanks again

samples/XCT.Sample.WPF/MainWindow.xaml.cs
VladislavAntonyuk: Is it a bad practice or any violation in XCT?
As you can see it is used in PancakeView. Why can’t we do the same?

pictos: PacakeView isn't part of XCT.
So this is an agreement that we have on this project, that we want all devs to be able to use it without adding any other line of code. Like plug and play lib.

VladislavAntonyuk: I moved it to the Sample project. Init method is removed

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Renderer/DrawingViewRenderer.android.cs
```diff
@@ -67,16 +67,16 @@ protected override void OnSizeChanged(int w, int h, int oldw, int oldh)
...
 			LoadPoints();
 		}
 
-		protected override void OnDraw(Canvas canvas)
+		protected override void OnDraw(Canvas? canvas)
 		{
 			base.OnDraw(canvas);
 
-			canvas.DrawBitmap(canvasBitmap, 0, 0, canvasPaint);
+			canvas!.DrawBitmap(canvasBitmap!, 0, 0, canvasPaint);
```
pictos, line 79: if we receive a nullable canvas, should we trust that isn't null here? Does the base call check for null?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Service/DrawingViewService.android.cs
```diff
@@ -49,9 +59,9 @@ static class DrawingViewService
 			var drawingHeight = points.Max(p => p.Y) - minPointY;
 			const int minSize = 1;
 			if (drawingWidth < minSize || drawingHeight < minSize)
-				return null;
+				throw new Exception("Image is too small");
```
pictos, line 62: Should add some suggestion, like "please provide a bigger image" or "the image should have at least ${height} x ${width}" ?
```suggestion
				throw new Exception("Image is too small.");
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Service/DrawingViewService.gtk.cs
```diff
@@ -53,7 +62,7 @@ static class DrawingViewService
 			const int minSize = 1;
 			if (drawingWidth < minSize || drawingHeight < minSize)
 			{
-				return null;
+				throw new Exception("Image is too small");
```
pictos, line 65: same here

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Service/DrawingViewService.ios.cs
```diff
@@ -44,7 +53,7 @@ static class DrawingViewService
 			const int minSize = 1;
 			if (drawingWidth < minSize || drawingHeight < minSize)
 			{
-				return null;
+				throw new Exception("Image is too small");
```
pictos, line 56: same here

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Service/DrawingViewService.macos.cs
```diff
@@ -38,32 +47,27 @@ static class DrawingViewService
 			const int minSize = 1;
 			if (drawingWidth < minSize || drawingHeight < minSize)
 			{
-				return null;
+				throw new Exception("Image is too small");
```
pictos, line 50: same here

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Service/DrawingViewService.tizen.cs
```diff
@@ -49,7 +58,7 @@ static class DrawingViewService
 			const int minSize = 1;
 			if (drawingWidth < minSize || drawingHeight < minSize)
 			{
-				return null;
+				throw new Exception("Image is too small");
```
pictos, line 61: same here

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Service/DrawingViewService.uwp.cs
```diff
@@ -53,29 +62,27 @@ static class DrawingViewService
 			const int minSize = 1;
 			if (drawingWidth < minSize || drawingHeight < minSize)
 			{
-				return null;
+				throw new Exception("Image is too small");
```
pictos, line 65: same here

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Service/DrawingViewService.wpf.cs
```diff
@@ -14,10 +14,98 @@ namespace Xamarin.CommunityToolkit.UI.Views
...
+			Color backgroundColor)
+		{
+			var minPointX = points.Min(p => p.X);
+			var minPointY = points.Min(p => p.Y);
+			var drawingWidth = points.Max(p => p.X) - minPointX;
+			var drawingHeight = points.Max(p => p.Y) - minPointY;
+			const int minSize = 1;
+			if (drawingWidth < minSize || drawingHeight < minSize)
+			{
+				throw new Exception("Image is too small");
```
pictos, line 61: here as well

pictos (commented):
@VladislavAntonyuk I believe that this will be my last comments before merging this amazing work❣

pictos (approved):
Awesome!

PR #1316, opened 2021-05-18
CornerRadiusEffect

### LINKED ISSUE #1315: [Enhancement] CornerRadiusEffect
## Summary
Support `CornerRadius` on any `VisualElement`.

## API Changes
- `CornerRadiusEffect.CornerRadius`

## Intended Use Case
`CornerRadius` everywhere

## Who Will Do The Work?

YZahringer:
### Description of Change ###
Adding CornerRadius support everywhere by Clipping.
- .NET MAUI will support this: https://github.com/dotnet/maui/pull/650
- Clip issue on UWP: https://github.com/xamarin/Xamarin.Forms/issues/14275
- Works on iOS and Android. 

### Bugs Fixed ###
- Fixes #1315 
- Fixes https://github.com/xamarin/Xamarin.Forms/issues/12857

### API Changes ###
````cs
public class CornerRadiusEffect : Effect
{
    CornerRadius CornerRadius;
}
````

### Behavioral Changes ###
None

### PR Checklist ###

- [x] Has samples

### CODE REVIEW

**Comment by jfversluis** [comment]
Very cool, thank you @YZahringer 

I think @jsuarezruiz is a good candidate to have a look at this one? :)

File: src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/CornerRadius/CornerRadiusEffect.shared.cs [hunk 1]
```diff
@@ -0,0 +1,118 @@
...
+			if (GetCornerRadius(bindable) == default)
+			{
+				foreach (var cornerRadiusEffect in cornerRadiusEffects.ToArray())
+					elementView.Effects.Remove(cornerRadiusEffect);
+
+				return;
+			}
+
+			if (!cornerRadiusEffects.Any())
+				elementView.Effects.Add(new CornerRadiusEffect());
```
pictos, line 40: Let's create an array and reduce one LINQ call.
```suggestion
			var cornerRadiusEffects = elementView.Effects.OfType<CornerRadiusEffect>().ToArray();

			if (GetCornerRadius(bindable) == default)
			{
				foreach (var cornerRadiusEffect in cornerRadiusEffects)
					elementView.Effects.Remove(cornerRadiusEffect);

				return;
			}

			if (cornerRadiusEffects.Length == 0)
				elementView.Effects.Add(new CornerRadiusEffect());
```

pictos (changes requested):
@YZahringer thanks for this PR! I also have a question for you... Why use an Effect, instead to use a Behavior or just an Attached property? Since we don't need to  access any platform-specific API

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/CornerRadius/CornerRadiusEffect.shared.cs
YZahringer: Maybe the same fix here? https://github.com/xamarin/XamarinCommunityToolkit/blob/c6e30062681b5733870ac87932753cb9fb7dd10f/src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Shadow/ShadowEffect.shared.cs#L106

YZahringer:
@pictos thanks for the review. I followed the same idea as `ShadowEffect`. But yes, currently there is no platform specific implementation, we could use an `AttachedBehavior` for that.

There is an issue with UWP and I hope a workaround can be found. This one may require access to the native control and in that case using the Effect makes more sense?

pictos:
@YZahringer yes, if we need to access the platform api, for sure an Effect will be the way to go.

FYI, I could make it work with behavior and I belive that I can make it work just with an attached property, that way can have lighter api for this feature.

Let me know what you think and if you can make it work on uwp using shared code.

AndreiMisiukevich:
I'd prefer to leave it as effect.
Because it's just visual effect:)) I doesn't bring any behavior.

But let's collect more opinions

pictos:
> I'd prefer to leave it as effect.
> Because it's just visual effect:)) I doesn't bring any behavior.
> 
> But let's collect more opinions

I believe that we can keep the `Effect` name but implement it in a such way that we don't need the `XF.Effect` API

pictos:
@YZahringer and others, please let me know what do you think about it
https://gist.github.com/pictos/f410a6983e75b2f3a993ce092c970a36

I tested on my end, using your sample and works as expected.

AndreiMisiukevich:
> @YZahringer and others, please let me know what do you think about it
> https://gist.github.com/pictos/f410a6983e75b2f3a993ce092c970a36
> 
> I tested on my end, using your sample and works as expected.

Looks good to me)

YZahringer:
@pictos thanks, I have adapted the code with your proposal 👍

For UWP I have no solution unfortunately. I think the problem is [here](https://github.com/xamarin/Xamarin.Forms/blob/5.0.0/Xamarin.Forms.Platform.UAP/Extensions/ClipExtensions.cs) with the `GeometryGroup`. Maybe we can try to fix it in another PR.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/CornerRadius/CornerRadiusEffect.shared.cs
```diff
@@ -1,82 +1,47 @@
...
-			=> (CornerRadius)(bindable?.GetValue(CornerRadiusProperty) ?? throw new ArgumentNullException(nameof(bindable)));
-
-		public static void SetCornerRadius(BindableObject? bindable, CornerRadius value)
-			=> bindable?.SetValue(CornerRadiusProperty, value);
-
-		static void TryGenerateEffect(BindableObject? bindable, object oldValue, object newValue)
+		static void OnCornerRadiusPropertyChanged(BindableObject bindable, object oldValue, object newValue)
 		{
-			if (bindable is not VisualElement elementView)
+			if (!(bindable is VisualElement elementView))
```
pictos, line 18: when I did that code I didn't use C#9
```suggestion
			if (bindable is not VisualElement elementView))
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/CornerRadius/CornerRadiusEffect.shared.cs
```diff
@@ -1,82 +1,47 @@
...
 				return;
 
-			var cornerRadiusEffects = elementView.Effects.OfType<CornerRadiusEffect>().ToArray();
-
-			if (GetCornerRadius(bindable) == default)
-			{
-				foreach (var cornerRadiusEffect in cornerRadiusEffects)
-					elementView.Effects.Remove(cornerRadiusEffect);
+			elementView.SizeChanged -= ElementViewSizeChanged;
+			elementView.SizeChanged += ElementViewSizeChanged;
```
pictos, line 22: Just to point out, that we need to see if this will not cause a memory leak. I believe that doesn't, but just want to make sure to bring more eyes to it.

pictos:
@YZahringer thanks for moving this forward, I'll do a last review during this week and hopefully merge it (:

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/CornerRadius/CornerRadiusEffect.shared.cs
```diff
@@ -0,0 +1,83 @@
...
+
+			elementView.SizeChanged -= ElementViewSizeChanged;
+			elementView.SizeChanged += ElementViewSizeChanged;
+
+			UpdateClip(elementView);
+		}
+
+		static void ElementViewSizeChanged(object? sender, EventArgs e)
+		{
+			if (!(sender is VisualElement elementView))
```
YZahringer, line 29:
```suggestion
			if (sender is not VisualElement elementView)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/CornerRadius/CornerRadiusEffect.shared.cs
YZahringer: Yes, we don't unsubscribe from the event as there is no detach method. Maybe use WeakEventManager?

pictos: Not sure if we can use the WeakEventManager to subscribe to other events. I'll confirm if this can be an issue, if so I'll implement a fix during this week (;

YZahringer: Yes, unfortunately the current implementation of `WeakEventManager` does not support subscription to other events.

Maybe an enhancement that could be added, like with WPF: https://docs.microsoft.com/en-us/dotnet/desktop/wpf/advanced/weak-event-patterns?view=netframeworkdesktop-4.8#using-an-existing-weak-event-manager-class

pictos: I really like this idea. @brminnick do you think that could be add on WeakEventManager or would be better to have this as a separated internal feature?

brminnick: Interesting! @YZahringer is there an open Issue for it? If not, could you create one and I'll take a look!

YZahringer:
Any news about this? We have to wait #1410?

pictos:
Hey @YZahringer no news yet. I'll try to return back to this one next week. I was focused on the 1.2 release

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/CornerRadius/CornerRadiusEffect.shared.cs
```diff
@@ -0,0 +1,83 @@
...
+
+			UpdateClip(elementView);
+		}
+
+		static void ElementViewSizeChanged(object? sender, EventArgs e)
+		{
+			if (sender is not VisualElement elementView)
+				return;
+
+			UpdateClip(elementView);
```
pictos, line 32:
```suggestion
		if (sender == null)
				return;

			UpdateClip((VisualElement)sender);
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/CornerRadius/CornerRadiusEffect.shared.cs
pictos: Taking [this old MS app](https://github.com/Microsoft/SmartHotel360-Mobile/blob/master/Source/SmartHotel.Clients/SmartHotel.Clients/Behaviors/ItemTappedCommandListView.cs#L20) as a reference it's ok to use this approach. I think that we're good to go

brminnick:
It looks like this feature was already approved before the deadline for more features.

@pictos Could you give it one more review?

pictos (approved):
LGTM

Issue #1135, opened 2021-03-27
[Spec] Calendar


DamianAntonowicz:
# Calendar

Calendar control that displays calendar ;) Spec is based on working solution from my GitHub: https://github.com/DamianAntonowicz/XamarinForms.CalendarControl

Calendar is using CalendarDay control to display specific days. Calendar is also using CalendarWeekDayHeader control to display at the top week day headers: Monday, Tuesday, etc.

# API

### Properties

| Property | Type | Description |
| ------------- | ------------- | ------------- |
| Date | DateTime | Current date that calendar is displaying. |
| SelectionMode | CalendarSelectionMode | Selection mode for days: SingleSelect or MultiSelect. |
| ShowDaysFromOtherMonths | bool | If set to `true` calendar will show days from other months. |
| ShowWeekends | bool | If set to `true` calendar will show weekends. |
| FirstDayOfWeek | DayOfWeek | First day of week: Monday, Sunday, etc. |
| SelectedDays | IReadOnlyCollection\<DateTime> | Days that are currently selected. |
| Days | IReadOnlyCollection\<CalendarDay> | Days that are currently being displayed.. |
| DayControlTemplate | ControlTemplate | ControlTemplate for all days. |
| WeekDayHeaderControlTemplate | ControlTemplate | ControlTemplate for week day header at the top of calendar. |

### Events

| Event | Args | Description |
| ------------- | ------------- | ------------- |
| DayTapped | CalendarDayTappedEventArgs | Event that is raised when user tapped on specific day. EventArgs allows get access to specific CalendarDay control. |
| DayUpdated | CalendarDayUpdatedEventArgs | Event that is raised when Calendar has updated day. EventArgs allows get access to specific CalendarDay control. |
 
## CalendarDay : TemplatedView

Displays specific day.

### Properties

| Property | Type | Description |
| ------------- | ------------- | ------------- |
| IsSelectable | bool | If set to `true` specific CalendarDay can be selected by user. |
| IsSelected | bool | If set to `true` specific CalendarDay is currently selected. |
| Date | DateTime | Current date that specific CalendarDay is displaying. |

## CalendarWeekDayHeader : TemplatedView

Displays week day header at the top of calendar.

### Properties

| Property | Type | Description |
| ------------- | ------------- | ------------- |
| DayOfWeek | DayOfWeek | Day of week that specific CalendarWeekDayHeader is displaying. |

# Scenarios

## Calendar

```xml
[...]

PR #1331, opened 2021-05-23
Added Calendar control.

DamianAntonowicz:
### Description of Change ###

Added Calendar control.

### API Changes ###

Detailed Spec is in this issue: https://github.com/xamarin/XamarinCommunityToolkit/issues/1135

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
``` -->

### PR Checklist ###

- [x] Has a linked Issue, and the Issue has been `approved` - this is the issue: https://github.com/xamarin/XamarinCommunityToolkit/issues/1135
- [ ] Has tests (if omitted, state reason in description) - not sure, what part of code should have tests, but I'm open for suggestions
- [x] Changes adhere to coding standard - I did my best, but this is my first PR so I'm open for suggestions :)

### CODE REVIEW

**Comment by pictos** [comment]
@DamianAntonowicz can you solve the issue related to the build error? It's about naullability, a friendly reminder that we treat naullability warning as error in this project. If you need help please let me know.

**Comment by dimonovdd** [comment]
Oh that's very cool. Consider also swiping your finger right and left to change the month and displaying only a row with one week.

**Comment by DamianAntonowicz** [comment]
@pictos I will need help on these build nullability errors on Windows. I'm working mainly on macOS and I don't get this build nullability errors (using Rider or VS for Mac). Also I can see that building Azure Pipeline for macOS Library was successful.

I've switched to Windows on my machine and also don't get this build errors about nullability. Am I missing some configuration?

**Comment by pictos** [comment]
@DamianAntonowicz no worries, sometimes VS doesn't do a good job showing the nullable errors. I'll take a look at it by tomorrow and make this beautiful PR green on our CI

File: src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs [hunk 1]
```diff
@@ -0,0 +1,564 @@
...
+		{
+			var calendar = (Calendar)bindable;
+			calendar.UpdateCalendarDays();
+		}
+
+		/// <summary>
+		/// Backing BindableProperty for the <see cref="SelectedDays"/> property.
+		/// </summary>
+		public static readonly BindableProperty SelectedDaysProperty =
+			BindableProperty.Create(nameof(SelectedDays), typeof(IReadOnlyCollection<DateTime>), typeof(Calendar), Enumerable.Empty<DateTime>().ToList(), propertyChanged: OnSelectedDaysChanged);
```
brminnick, line 158: Just curious - how come we're using `DateTime` instead of `DateTimeOffset`?

I'd prefer to use `DateTimeOffset` in order to support timezones.

Let me know if there is context that I'm missing!

brminnick: If we move forward with `DateTimeOffset`, we'd need to change all `DateTime` to `DateTimeOffset` in this PR.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
```diff
@@ -0,0 +1,533 @@
...
+
+		static void OnSelectedDaysChanged(BindableObject bindable, object oldValue, object newValue)
+		{
+			var daysToSelect = (IList<DateTime>)newValue;
+			var calendar = (Calendar)bindable;
+
+			if (calendar.SelectionMode is CalendarSelectionMode.SingleSelect
+				&& daysToSelect.Count > 1)
+			{
+				throw new InvalidOperationException(
```
pictos, line 176: Do we want to throw an exception? If we throw an exception here, I believe that dev can't catch, mainly if it's using XAML, in other words the app will crash, isn't better to log this message and just return?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
```diff
@@ -0,0 +1,533 @@
...
+			var calendar = (Calendar)bindable;
+			calendar.SelectedDays = Enumerable.Empty<DateTime>().ToList();
+		}
+
+		/// <summary>
+		/// Initializes a new instance of the <see cref="Calendar"/> class.
+		/// </summary>
+		public Calendar()
+		{
+			var layoutRoot = new Grid
```
pictos, line 230: @DamianAntonowicz here at XCT we have an [`UniformGridLayout`](https://github.com/xamarin/XamarinCommunityToolkit/blob/main/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/UniformGrid.shared.cs) do you think that could be useful here? Since all items will have the same size

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
```diff
@@ -0,0 +1,533 @@
...
+
+					UpdateDay(newDate, week, isVisible: ShowDaysFromOtherMonths);
+				} while (newDate.DayOfWeek != FirstDayOfWeek);
+			}
+			else if (week == weeksInMonth
+					 && date.Day == daysInMonth
+					 && date.DayOfWeek != FirstDayOfWeek.PreviousOrFirst())
+			{
+				var newDate = date;
+				var lastDayOfWeek = FirstDayOfWeek.PreviousOrFirst();
```
pictos, line 389: can we cache the result for `FirstDayOfWeek.PreviousOrFirst()` avoiding duplicate calls?
```suggestion
                    // somewhere at the top
                    var firstDayOfWeek = FirstDayOfWeek.PreviousOrFirst();
			else if (week == weeksInMonth
					 && date.Day == daysInMonth
					 && date.DayOfWeek != firstDayOfWeek)
			{
				var newDate = date;
				var lastDayOfWeek = firstDayOfWeek;
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
```diff
@@ -0,0 +1,533 @@
...
+				calendarDay.ControlTemplate = DayControlTemplate;
+			}
+
+			calendarDay.Opacity = isVisible ? 1 : 0;
+
+			if (isVisible)
+				DayUpdated?.Invoke(this, new CalendarDayUpdatedEventArgs(calendarDay));
+		}
+
+		void CalendarDay_OnTapped(object? sender, EventArgs e)
```
pictos, line 444: Can we rename it? Maybe something like `OnCalendarDayTapped`

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
```diff
@@ -0,0 +1,533 @@
...
+		{
+			var calendarDay = (CalendarDay)(sender ?? throw new ArgumentNullException($"{nameof(sender)} cannot be null"));
+
+			if (calendarDay.IsSelectable)
+			{
+				if (SelectionMode is CalendarSelectionMode.SingleSelect)
+				{
+					if (SelectedDays.Any() && SelectedDays.First().Date == calendarDay.Date)
```
pictos, line 452: Let's avoid unnecessary LINQ calls
```suggestion
					if (SelectedDays.Count > 0 && SelectedDays.First().Date == calendarDay.Date)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
```diff
@@ -0,0 +1,533 @@
...
+					else
+					{
+						newSelectedDays.Add(calendarDay.Date);
+					}
+
+					SelectedDays = new ReadOnlyCollection<DateTime>(newSelectedDays);
+				}
+				else
+				{
+					throw new NotSupportedException($"{SelectionMode} is not supported");
```
pictos, line 474: Again, not sure if we want to throw an exception here.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/CalendarWeekDayHeader.shared.cs
```diff
@@ -0,0 +1,26 @@
...
+		public DayOfWeek DayOfWeek { get; }
+        
+		/// <summary>
+		/// Initializes a new instance of the <see cref="CalendarWeekDayHeader"/> class.
+		/// </summary>
+		/// <param name="dayOfWeek">day of the week</param>
+		public CalendarWeekDayHeader(DayOfWeek dayOfWeek)
+		{
+			DayOfWeek = dayOfWeek;
+			var a = DayOfWeek; // Without this line ControlTemplate doesn't work.
```
pictos, line 23: Can you talk more about it? I'm curious why that happens and what happened if we remove this line. Is it failing on XCT or in the original project as well?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
```diff
@@ -0,0 +1,533 @@
...
+					gridDays.RowDefinitions.RemoveAt(gridDays.RowDefinitions.Count - 1);
+
+					var daysToRemove = days.Where(x => Grid.GetRow(x) == row - 1).ToList();
+
+					foreach (var calendarDay in daysToRemove)
+					{
+						gridDays.Children.Remove(calendarDay);
+					}
+				}
+			}
```
pictos, line 331: Here we can cache the `gridDays.RowDefinitions.Count` avoiding to calling the `get` method everywhere. If I'm correct that value will not change inside this method.
```suggestion			
                        var gridDaysRowDefinitionsCount = gridDays.RowDefinitions.Count;
			if (gridDaysRowDefinitionsCount < weeksInMonth)
			{
				for (var row = gridDaysRowDefinitionsCount; row < weeksInMonth; row++)
				{
					gridDays.RowDefinitions.Add(new RowDefinition { Height = GridLength.Auto });

					var daysToAdd = days.Where(x => Grid.GetRow(x) == row).ToList();

					foreach (var calendarDay in daysToAdd)
					{
						gridDays.Children.Add(calendarDay);
					}
				}
			}
			else if (gridDaysRowDefinitionsCount > weeksInMonth)
			{
				for (var row = gridDaysRowDefinitionsCount; row > weeksInMonth; row--)
				{
					gridDays.RowDefinitions.RemoveAt(gridDaysRowDefinitionsCount - 1);

					var daysToRemove = days.Where(x => Grid.GetRow(x) == row - 1).ToList();

					foreach (var calendarDay in daysToRemove)
					{
						gridDays.Children.Remove(calendarDay);
					}
				}
			}
```

pictos (changes requested):
@DamianAntonowicz I didn't run the code on my end yet, but I went through the code and I made some comments here. Let me know if you need any help

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
brminnick: I support throwing the exception because it means that the code has gotten to an invalid state; a developer should never be in a scenario where `calendar.SelectionMode is CalendarSelectionMode.SingleSelect &&  daysToSelect.Count > 1`

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
brminnick: I added this exception.

The `CalendarSelectionMode` enum only has two values: `SingleSelect` and `MultiSelect`.

If we ever add another value to the `CalendarSelectionMode` enum in the future, the `NotSupportedException` ensures that we don't forget to handle the new enum value.

pictos: I see that makes sense, thanks for the clarification

DamianAntonowicz:
@brminnick @pictos thanks for the code review. I will push the changes this week

samples/XCT.Sample/Xamarin.CommunityToolkit.Sample.csproj
```diff
@@ -11,6 +11,10 @@
     <None Remove="resources\**" />
     <EmbeddedResource Include="Images\**" />
     <EmbeddedResource Include="Fonts\**" />
+    <Compile Update="Pages\Views\Calendar\CalendarPage.xaml.cs">
+      <DependentUpon>CalendarPage.xaml</DependentUpon>
+      <SubType>Code</SubType>
+    </Compile>
```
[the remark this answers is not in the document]
DamianAntonowicz: sure, removed

src/CommunityToolkit/Xamarin.CommunityToolkit/Extensions/DateTimeExtensions.shared.cs
```diff
@@ -0,0 +1,82 @@
...
+                    currentDayOfWeek = System.DayOfWeek.Monday;
+                }
+                
+                dayOfWeek++;
+            }
+
+            return dayOfWeek;
+        }
+        
+        private static int WeekOfYear(this DateTime dateTime, DayOfWeek firstDayOfWeek)
```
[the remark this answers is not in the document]
DamianAntonowicz: I've removed default modifiers

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
DamianAntonowicz: thanks, I've added caching

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
DamianAntonowicz: I've refactored name of this method

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
DamianAntonowicz: I can see that UniformGrid is missing ColumnDefinitions and RowDefinitions - Calendar logic is based on those properties to show/hide days

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
```diff
@@ -0,0 +1,533 @@
...
+
+		/// <summary>
+		/// Gets days that are currently visible.
+		/// </summary>
+		public IReadOnlyCollection<CalendarDay> Days => days.Where(x => x.IsVisible).ToList();
+
+		/// <summary>
+		/// Event that is triggered when the <see cref="CalendarDay" /> is tapped.
+		/// </summary>
+		public event EventHandler<CalendarDayTappedEventArgs>? DayTapped;
```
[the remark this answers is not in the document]
DamianAntonowicz: I've migrated events in Calendar to use WeakEventManager

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/CalendarWeekDayHeader.shared.cs
DamianAntonowicz: Without this line week day headers at the top of calendar won't show:
Screenshot 2021-05-30 at 17 04 27

This was not happening in the original project.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
DamianAntonowicz: value is now being cached at the top of the method

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
DamianAntonowicz: I've changed DateTime to DateTimeOffset

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
```diff
@@ -0,0 +1,549 @@
...
+				days.Add(calendarDay);
+				gridDays.Children.Add(calendarDay);
+			}
+			else
+			{
+				calendarDay.Date = date;
+				calendarDay.ControlTemplate = DayControlTemplate;
+			}
+
+			calendarDay.Opacity = isVisible ? 1 : 0;
```
DamianAntonowicz, line 452: @brminnick this line was originally wrapped into `Device.BeginInvokeOnMainThread` because I've noticed some visible delay between when day is updated with new value. Below is video that shows this problem:

https://user-images.githubusercontent.com/12022141/120119433-f0da6880-c197-11eb-9527-a9e207742b7f.mov

When switching to next month, days in the first row have old value for a slight of second before correct date is visible. Wrapping this line in `Device.BeginInvokeOnMainThread` solves this issue.

Let me know what you think.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/CalendarWeekDayHeader.shared.cs
brminnick: > Without this line week day headers at the top of calendar won't show
Do you know why?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
brminnick: @DamianAntonowicz - Can you open a new enhancement Issue for adding `ColumnDefinitions` and `RowDefinitions` to `UniformGrid`? 

After `UniformGrid` has that functionality, we can implement here in `Calendar` in a later PR.

After that enhancement Issue is created, we can `Resolve` this conversation 👍

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
brminnick: Looks like that behavior only happens when `Show days from other months` is not enabled (video attached).

I'll investigate why. It's likely just an order-of-operations bug; i.e. We're changing the days before changing the month.

https://user-images.githubusercontent.com/13558917/120235760-9aa90c00-c20f-11eb-87c6-506945f157ec.mp4

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/CalendarWeekDayHeader.shared.cs
DamianAntonowicz: no idea :(

brminnick: Could you investigate and resolve it?

I’d prefer not to merge this workaround into the Toolkit.

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
DamianAntonowicz: ok, I will open enhancement

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/CalendarWeekDayHeader.shared.cs
DamianAntonowicz: me neither :) I will investigate this

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
DamianAntonowicz: I will take a look at this

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/CalendarWeekDayHeader.shared.cs
DamianAntonowicz: @brminnick I've investigated this and have some conclusions. If I delete line 23 from CalendarWeekDayHeader and set "Link behavior" to "Do not link" for "Debug | iPhoneSimulator " in Xamarin.CommunityToolkit.Sample.iOS, then week day headers are properly displayed at the top of calendar. It seems that linker is somehow responsible for this issue. 

What would you suggest?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
DamianAntonowicz: @brminnick  writing code like this:

```
        else
	{
		calendarDay.Date = date;
	}

	if (isVisible)
	{
		OnDayUpdated(calendarDay);
	}

	if (Device.RuntimePlatform.Equals(Device.iOS))
	{
		await Task.Delay(1);
	}

	calendarDay.Opacity = isVisible ? 1 : 0;
}
```

Solves the issue on iOS. Let me know what you think

brminnick: What's the root cause, though?

`await Task.Delay` is the same workaround as `Device.BeginInvokeOnMainThread`; both generate a thread/context switch which negatively impacts performance, and I'd like to avoid that.

DamianAntonowicz: I couldn't figure out the root cause :/ Even if we set `calendarDay.Date = date` before setting `calendarDay.Opacity = 1`, it's still being updated visually AFTER CalendarDay becomes visible, which causes that for a slight of second old value is visible. It seems like iOS is not updating visual presence of UI component until it is visible to user.

DamianAntonowicz: @brminnick I've created simple project to reproduce this issue: https://github.com/DamianAntonowicz/XamarinFormsIosOpacityTest

Here is a video with issue reproduced:
https://user-images.githubusercontent.com/12022141/120925179-aeed7d00-c6d7-11eb-87d2-ff6d8bf5784a.mov

It is not happening every time, but it happens and is visible.

The code for this is very simple:
```
private void ButtonOpacityToZero_OnClicked(object sender, EventArgs e)
        {
            CalendarDay1.Opacity = 0;
 
            ...
        }
        
        private void ButtonOpacityToOne_OnClicked(object sender, EventArgs e)
        {
            var random = new Random();
            
            CalendarDay1.Date = DateTimeOffset.Now.AddDays(random.Next(0, 10));
            CalendarDay1.Opacity = 1;

            ...
        }```

Do you have any suggestions? :)

**Comment by davidbuckleyni** [comment]
Please don't forget to rebase your branch before commiting

File: src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs (hunk 3 above)
Comment (DamianAntonowicz) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs:243]: @brminnick I've opened enhancement: https://github.com/xamarin/XamarinCommunityToolkit/issues/1370

File: src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/CalendarWeekDayHeader.shared.cs (hunk 8 above)
Comment (pictos) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/CalendarWeekDayHeader.shared.cs:23]: on CalendarView ctor, add this line
`_ = new CalendarWeekDayHeader (dumbDayOfWeek);`

and let me know if it works.

Comment (DamianAntonowicz) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/CalendarWeekDayHeader.shared.cs:23]: @pictos I've deleted line 23 from CalendarWeekDayHeader and added suggested line by you to Calendar ctor - this didn't solve the issue. Week day headers at the top of calendar are still not visible

Comment (pictos) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/CalendarWeekDayHeader.shared.cs:23]: @DamianAntonowicz I'll give some love on this issue during this week and let you know if I find any fix.

Comment (DamianAntonowicz) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/CalendarWeekDayHeader.shared.cs:23]: @pictos did you had time to take a look at this?

File: src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs (hunk 13 above)
Comment (DamianAntonowicz) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs:454]: @brminnick did you had time to take a look at this issue?

File: src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/CalendarWeekDayHeader.shared.cs (hunk 8 above)
Comment (pictos) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/CalendarWeekDayHeader.shared.cs]: Fixed

**Comment by pictos** [comment]
@DamianAntonowicz I'm so sorry for the late reply here, life just happened. I just fixed the Linker issue, now the Week header template is working without that `var a = DayOfWeek`. Let me know if you need anything else from me to move this forward. And I'm sorry again for the huge delay

**Comment by DamianAntonowicz** [comment]
@pictos thanks for fixing linker issue! 

There is one more open comment with which I will need help - I will ping you there

File: src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs (hunk 13 above)
Comment (DamianAntonowicz) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs:454]: @pictos could you take a look at this conversation? Maybe you will have some idea :)

Comment (pictos) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs:454]: can this be a bug on XF?
> I only see this behavior happening on iOS; it is not happening on my Android simulator.

From Brandon's comment, this looks like a platform-specific bug, in other words, how does iOS handle the rendering. If I'm not wrong we can disable animations on iOS, maybe this can solve the issue. We disable the animation to change values and re-enabled it after setting `Opacity` to 1.

@DamianAntonowicz can you test it? I don't have iOS environment configured

Comment (DamianAntonowicz) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs]: @pictos thanks for the tip! I've fixed the issue with following code:
```
calendarDay.FadeTo(isVisible ? 1 : 0, length: 0);
```

**Comment by DamianAntonowicz** [comment]
@pictos @brminnick I've resolved all the comments. Let me know if this PR is ready to merge :)

File: src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs (hunk 13 above)
Comment (brminnick) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs]: @DamianAntonowicz This does not resolve the issue.

`FadeTo` returns a `Task`, which we have to `await`. 

This then requires us to update the following methods to make them asynchronous:
- `async Task UpdateDay`
- `async Task UpdateDaysFromPreviousOrNextMonths`
- `async Task UpdateDays`
- `async Task UpdateCalendarDays`
- `async Task UpdateCalendarDays`
- `static async void OnDayControlTemplateChanged`
- `static async void OnFirstDayOfWeekChanged`
- `static async void OnShowWeekendsChanged`
- `static async void OnShowDaysFromOtherMonthsPropertyChanged`

After updating those methods, the Calendar UI on both iOS + Android then has a ripple effect where the days appear & disappear in sequence instead of in parallel:
Simulator Screen Recording - iPhone 12 - 2021-07-28 at 16 28 29

Comment (brminnick) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs]: I see two paths for fixing this bug:
- Keep `await FadeTo`, and update your logic to batch the day changes so that the days update in parallel
- Revert [this commit](https://github.com/xamarin/XamarinCommunityToolkit/pull/1331/commits/ee9788fdde518a01e63b9ffe195d6d43a122ef1a) and fix the original problem on iOS. 
  - it may require a custom renderer on iOS

Comment (DamianAntonowicz) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs]: @brminnick I've rushed with this change, sorry for that. Could you point me in proper direction with custom renderer on iOS? I'm not sure how this renderer should be created.

Comment (DamianAntonowicz) [reply src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs]: @brminnick @pictos I've updated Calendar logic to update days in parallel. I think it fixed update issue on iOS - at least I cannot reproduce issue anymore. 

Let me know what you think about code changes.

File: samples/XCT.Sample/Pages/Views/Calendar/DateTimeOffsetToDayConverter.cs [hunk 14]
```diff
@@ -0,0 +1,22 @@
...
+using System.Globalization;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Sample.Pages.Views
+{
+	public class DateTimeOffsetToDayConverter : IValueConverter
+	{
+		public object Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture)
+		{
+			if (value == null)
```
pictos, line 11: The `null` check will not prevent the possible `InvalidCastException` at line 14
```suggestion
			if (value is not DateTimeOffset)
```

samples/XCT.Sample/Pages/Views/Calendar/DayOfWeekToAbbreviatedNameConverter.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Globalization;
+using Xamarin.Forms;
+
+namespace Xamarin.CommunityToolkit.Sample.Pages.Views
+{
+	public class DayOfWeekToAbbreviatedNameConverter : IValueConverter
+	{
+		public object Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture)
+		{
+			if (value == null)
```
pictos, line 11:
```suggestion
			if (value is not DayOfWeek)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
```diff
@@ -0,0 +1,653 @@
...
+			layoutRoot.Children.Add(gridDays);
+			this.gridDays = gridDays;
+
+			Content = layoutRoot;
+		}
+
+		/// <summary>
+		/// Initializes <see cref="Calendar"/>.
+		/// </summary>
+		public void Initialize()
```
pictos, line 319: Can we do it in another way? I mean, we would like to not have the users call `Init` methods at all. Our goal is to have more of a plug-and-play lib. I imagine that's because of the change to `Task` in other methods, right? If yes, did you try to disable the iOS animation while performing the month change?

pictos (commented):
Hey @DamianAntonowicz I have some questions about this new commit

pictos:
https://user-images.githubusercontent.com/20712372/128099233-413f85c8-4811-4bca-9015-38d1333f2538.mp4

@DamianAntonowicz @brminnick here's my workaround for this issue. Let me know what you think about it, if it's good enough I'll commit it. 

To solve it I disabled the iOS animation and used the `sync` version of the code and the issue doesn't happen anymore. The thing is the mark (red dot) is rendering too soon. The good part is that the code keeps sync and simple but we need to figure out how to present the mark at the right moment.

pictos:
@DamianAntonowicz @brminnick I posted I video before where I fixed the bug, but it was wrong because I forgot to uncheck the `Show days from other months` option... And doing some spikes on my end, adding this extension method to the `FadeTo` line `calendarDay.FadeTo(isVisible ? 1 : 0, length: 0).SafeFireAndForget();` solves the issue. How do you feel about it?

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Calendar/Calendar.shared.cs
DamianAntonowicz: There was following issue: properties ```ShowWeekends``` and ```ShowDaysFromOtherMonths``` both have default value set to ```true```. When I set those values to ```false``` in XAML then methods ```OnShowDaysFromOtherMonthsPropertyChanged``` and ```OnShowWeekendsChanged``` are invoked. Both methods invoke ```UpdateCalendarDaysAsync```, which updates calendar days. Method ```UpdateCalendarDaysAsync``` was called twice when user was entering page where Calendar was used - this was hitting performance.

```Initialize``` method is a way to invoke ```UpdateCalendarDaysAsync``` only once when all other bindable properties are already set.

Let me know what you think about this.

**Comment by DamianAntonowicz** [comment]
> @DamianAntonowicz @brminnick I posted I video before where I fixed the bug, but it was wrong because I forgot to uncheck the `Show days from other months` option... And doing some spikes on my end, adding this extension method to the `FadeTo` line `calendarDay.FadeTo(isVisible ? 1 : 0, length: 0).SafeFireAndForget();` solves the issue. How do you feel about it?

**Comment by DamianAntonowicz** [comment]
> @DamianAntonowicz @brminnick I posted I video before where I fixed the bug, but it was wrong because I forgot to uncheck the `Show days from other months` option... And doing some spikes on my end, adding this extension method to the `FadeTo` line `calendarDay.FadeTo(isVisible ? 1 : 0, length: 0).SafeFireAndForget();` solves the issue. How do you feel about it?

**Comment by DamianAntonowicz** [comment]
> @DamianAntonowicz @brminnick I posted I video before where I fixed the bug, but it was wrong because I forgot to uncheck the `Show days from other months` option... And doing some spikes on my end, adding this extension method to the `FadeTo` line `calendarDay.FadeTo(isVisible ? 1 : 0, length: 0).SafeFireAndForget();` solves the issue. How do you feel about it?

@pictos In my original code (https://github.com/DamianAntonowicz/XamarinForms.CalendarControl/blob/c1f4b646e4b111a685d436f68a17c34c65bf2838/XamarinForms.CalendarComponent/Components/Calendar.xaml.cs#L429) I have following line:

```
 Device.BeginInvokeOnMainThread(() =>
                {
                    if (isVisible)
                    {
                        calendarDay.Opacity = 1;
                    }
                    else
                    {
                        calendarDay.Opacity = 0;
                    }
                });
```

How using `calendarDay.FadeTo(isVisible ? 1 : 0, length: 0).SafeFireAndForget();` is better than using `Device.BeginInvokeOnMainThread()`, which also solves the issue on iOS?

**Comment by DamianAntonowicz** [comment]
@brminnick @pictos let me know if there is anything else that you would like to change in this PR. I think we are very close to merging this PR.

File: samples/XCT.Sample/Pages/Views/Calendar/DateTimeOffsetToDayConverter.cs [hunk 17]
```diff
@@ -0,0 +1,22 @@
+using System;
```
GUOLDEV, line 1: @DamianAntonowicz What about adding the converters as part of XCT converters? Not just in the Sample. It seems to be useful to XCT users.

DamianAntonowicz: @GUOLDEV I think these converters are just too simple to be worth adding them to XCT

brminnick:
Thanks Damien! Let's defer this to the .NET MAUI Community Toolkit.

PR #1368, opened 2021-06-05
Bring in 1.3.2.2 .core on Android

### LINKED ISSUE #1358: [Bug] Typo in Android SemanticEffect implementation
### Description
Typo in Android SemanticEffect implementation 
I'm not sure if this builds correctly but even if I implement this same effect in my app , the IDE can not find this namespace

https://github.com/xamarin/XamarinCommunityToolkit/blob/main/src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Semantic/SemanticEffectRouter.android.cs#L4

jamesmontemagno:
Fixes #1358

Accessibility namespace was fixed here. This will have better compat and minimal APIs differences.

Xamarin.Forms does not use this at all so should not be an issue with compiling at all.

### Bugs Fixed ###

- Fixes #

### API Changes ###

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
``` -->

### Behavioral Changes ###

### PR Checklist ###


brminnick:
Added reference to `Xamarin.AndroidX.RecyclerView` to fix linker error:
https://stackoverflow.com/a/66689618/5953643

brminnick:
Is #1358 a bug that is causing errors? To me, #1358 reads as a typo in the namespace.

I'd prefer not to take a dependency on `Xamarin.AndroidX.Core v1.5.0` and `Xamarin.AndroidX.RecyclerView v1.1.0.8`, because Xamarin.Forms has a dependency on an older version of the libraries. I.e. I could see the Xamarin.Community causing breaking changes by requiring newer versions of these libraries.

| SDK | Xamarin.Forms | PR #1368 |
| ---- | --------------- | --------- |
| Xamarin.AndroidX.Core | v1.3.2.1 | v1.5.0 |
| Xamarin.AndroidX.RecyclerView | v1.1.0.6 | v1.1.0.8 |

jamesmontemagno:
Agreed, I would prefer v 1.3.2.2 instead... let me see

jamesmontemagno:
@brminnick as you can see fails :(

pictos:
Can't we use the versions that Forms uses? Also, we should keep our eyes on [this PR](https://github.com/xamarin/Xamarin.Forms/pull/14101/files#diff-a10b47d2a1fb6f9a0f879b70a6b268191496328089b0bd5776fcbd68759d59dbR25) looks like the Forms is bumping your AndroidX versions.

jamesmontemagno:
Yes, we should align with them and ship at the same time. It may just go away if we build and compile against the new xamarin.forms.

The issue @pictos is that the current XF one we compile against is older and people are building other packages using newer packages of core that have fixes so compilation fails when things are miss matched.

brminnick:
Sounds good! 

We'll test the toolkit against the new version of Xamarin.Forms once it's released, and (assuming all the tests pass) we'll push out a release that targets the updated version of Xamarin.Forms 👍 

Since the Xamarin.Forms PR should resolve #1358, I've put the `Do Not Merge` label on the PR for now.

brminnick:
Closing in favor of https://github.com/xamarin/XamarinCommunityToolkit/pull/1502

Issue #906, opened 2021-02-15
[Enhancement] Better Background Image Source For Content Pages


Axemasta:
## Summary

I think the ability to set a background image on your `ContentPage` is a core feature. Xamarin Forms does support this with `BackgroundImageSource` however there is very little you can do in the way of customising how that image is displayed.

The current implementation is considered a bug by forms users ([XF #8863](https://github.com/xamarin/Xamarin.Forms/issues/8863)). I would call it less of a bug and more of a gap in the tooling. I propose an effect that allows the enum `Xamarin.Forms.Aspect` to control the display of the `BackgroundImageSource` on a `ContentPage`.

A popular solution to this problem is to use add a `Xamarin.Forms.Image` to the back of the`ContentPage` using a layout such as `RelativeLayout`. I do not like or recommend this approach because it causes a break down & seperation of concerns in the UI. A `ContentPage` should be responsible for setting and displaying its background, the `ContentPage.Content` should be responsible for the UI required for that page.

I wrote this effect for my own app and would like to offer it to the Community Toolkit as it will definitely be useful for other people.

## API Changes

Create a new effect named `BackgroundAspectEffect` that has the following parameters:

- Aspect (`Xamarin.Forms.Aspect`)
Through the use of platform specific renderers the background image can be manipulated to display the provided appearance.

Api Changes:

- New Effect
- Helper class to pull effect parameters off forms element at the platform level (see snippet below). This would add an extra piece of tooling for people making their own effects with parameters, the way you're currently expected to get your shared effect is not the nicest...
```csharp
public static class EffectHelpers
{
    public static TEffect GetFormsEffect<TEffect>(this Element element) where TEffect : Effect
    {
        if (element == null)
            throw new ArgumentNullException("Forms element must not be null", nameof(element));

        var effect = element.Effects.FirstOrDefault(e => e is TEffect);

        return (TEffect)effect;
    }
}
```
- iOS Platform Renderer (I have provided a proof of concept below with the iOS code implemented)
- Android Platform Renderer (This would need to be added).

## Intended Use Case
Giving people more control over the look and feel of their app without having to dive into writing their own custom renderer

## Who Will Do The Work?

- [X] I am willing to take this on myself (might need help with android..😄)
- [ ] Just putting this out there for someone to pick up

## Proof Of Concept

[...]

PR #1373, opened 2021-06-07
Added BackgroundAspectEffect

Axemasta:
### Description of Change ###

Created a new Effect named `BackgroundAspectEffect`. This effect allows you to control the display properties of a `ContentPage.BackgroundImageSource`, much like you would control backgrounds on the web using css. Currently when you set the `BackgroundImageSource` of a content page you get 2 very different results on iOS and Android. On iOS the image is converted to a `UIColor` and repeated on the screen according to its original size. On Android the image will stretch to fit the screen.

This behaviour is not ideal and is ideally a feature implemented at the platform level. However an issue has been present in the main Forms repo ([#8863](https://github.com/xamarin/Xamarin.Forms/issues/8863)) and with the transition to MAUI unlikely to make it into the box before LTS. There is a popular and widely regarded workaround to this issue, which is to nest your `ContentPage.Content` within a layout such as `AbsoluteLayout` and then place an `Image` behind the content of your view. This in my opinion is valid but also a little bit hacky and would prefer a cleaner way of setting, much like when working with `background-image` in css.

I have added an attached effect which will override the background image applied to the view and contain it within a native view that allows control of aspect ratio.

This PR relates to #906

I have also updated the sample to demonstrate the effect, seen in the gif below

backgroundaspect-ios

### API Changes ###

Added: 
 
- Attached Effect`BackgroundAspectEffect`

### Behavioral Changes ###

None

### PR Checklist ###


Axemasta:
I have a few questions to cover for this PR:

- I have updated the samples and I am currently using an image from a url. Are there any restrictions of images I can use? should I instead use a local image? if so what image should I use? Ideally the image is rectangular to best show off the difference between `AspectFit` & `AspectFill`
- I'm wouldn't say i'm well versed in `IImageSourceHandler `, in the iOS handler I am newing up specific handlers, would this affect compatibility with external libraries such as [`Xamarin.Forms.Nuke`](https://github.com/roubachof/Xamarin.Forms.Nuke)? 
- Do you know a good way of testing the detaching of effects, the effect does not get cleared when the page is popped (i imagine this is due to the attached effect + being static). What handling would you recommend I add to ensure it gets removed.
- Finally, I am using a `Task` and `CancellationTokenSource` to apply the effect. This is necessitated by the face `IImageSourceHandler` loads asynchronously. Each time the effect property is changed a new event occurs which is cancellable. I have't really put much thought into handlign the token or disposing of the source when completed. I would definitely appreciate any tips to best implement this (1 token source per property changed & then dispose a `List<IDisposable>` during cleanup. I say this because cancellation tokens bring with them easy to miss memory leaks 😅

pictos:
@Axemasta 
> Do you know a good way of testing the detaching of effects, the effect does not get cleared when the page is popped (i imagine this is due to the attached effect + being static). What handling would you recommend I add to ensure it gets removed.

You can try to create a button when clicked, you do :

```csharp
            foreach (var effect in this.Effects)
            {
                this.Effects.Remove(effect);
            }
```
but be aware that you may be facing [this bug](https://github.com/xamarin/Xamarin.Forms/issues/2520)

For others, I need to look further.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/BackgroundAspectEffect/BackgroundAspectEffectRouter.ios.cs
```diff
@@ -0,0 +1,152 @@
...
+			=> ClearBackgroundImage();
+
+		protected override void OnElementPropertyChanged(PropertyChangedEventArgs args)
+		{
+			base.OnElementPropertyChanged(args);
+
+			if (!args.PropertyName.Equals(BackgroundAspectEffect.AspectProperty.PropertyName))
+				return;
+
+			cancellationTokenSource.Cancel();
```
pictos, line 40: after this line, shouldn't we create a new instance?

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/BackgroundAspectEffect/BackgroundAspectEffectRouter.ios.cs
```diff
@@ -0,0 +1,152 @@
...
+
+		CancellationTokenSource cancellationTokenSource;
+
+		public BackgroundAspectEffectRouter()
+		{
+			cancellationTokenSource = new CancellationTokenSource();
+		}
+
+		protected override void OnAttached()
+			=> Task.Run(ApplyBackgroundImageAsync, cancellationTokenSource.Token);
```
pictos, line 28: Does this need to be a `Task.Run`? Why not call the Async method directly?

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/BackgroundAspectEffect/BackgroundAspectEffectRouter.ios.cs
```diff
@@ -0,0 +1,152 @@
...
+			if (imageView != null)
+			{
+				imageView.RemoveFromSuperview();
+
+				imageView.Dispose();
+			}
+
+			// TODO: Is this enough to cleanup?
+			cancellationTokenSource.Cancel();
+			cancellationTokenSource.Dispose();
```
pictos, line 86: Setting it to null, just to be safe.
```suggestion
			cancellationTokenSource.Dispose();
			cancellationTokenSource = null;
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/BackgroundAspectEffect/BackgroundAspectEffectRouter.ios.cs
```diff
@@ -0,0 +1,152 @@
...
+			else if (imageSource is UriImageSource)
+			{
+				handler = new ImageLoaderSourceHandler();
+			}
+			else
+			{
+				throw new NotSupportedException($"ImageSource is not supported by this effect: {imageSource.GetType().Name}");
+			}
+
+			return await handler.LoadImageAsync(imageSource);
```
pictos, line 110:
```suggestion
			return await handler.LoadImageAsync(imageSource).ConfigureAwait(false);
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/BackgroundAspectEffect/BackgroundAspectEffectRouter.ios.cs
```diff
@@ -0,0 +1,152 @@
...
+				child.TrailingAnchor.ConstraintEqualTo(parent.TrailingAnchor),
+				child.BottomAnchor.ConstraintEqualTo(parent.BottomAnchor)
+			};
+
+			NSLayoutConstraint.ActivateConstraints(constraints);
+		}
+
+		UIViewContentMode GetContentModeForAspect(Aspect aspect)
+		{
+			switch (aspect)
```
pictos, line 138: if you can use the new switch expression will be great. I believe that IDE can do it for you (;

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/BackgroundAspectEffect/BackgroundAspectEffectRouter.ios.cs
```diff
@@ -0,0 +1,152 @@
...
+			cancellationTokenSource.Cancel();
+
+			Task.Run(ApplyBackgroundImageAsync, cancellationTokenSource.Token);
+		}
+
+		async Task ApplyBackgroundImageAsync()
+		{
+			var contentPage = Element as ContentPage;
+
+			if (Element is null || contentPage is null || contentPage.BackgroundImageSource is null)
```
pictos, line 49: we don't check for `null` using pattern matching :(
```suggestion
			if (Element == null || contentPage == null || contentPage.BackgroundImageSource == null)
```

Axemasta: No worries, I've recently fallen in love with the pattern matching syntax: `if (something is null)` & `if (something is not null)` makes the code look so nice imo 😁

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/BackgroundAspectEffect/BackgroundAspectEffectRouter.ios.cs
Axemasta: I'll give it a go!

Axemasta (commented):
Apply code review suggestions

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/BackgroundAspectEffect/BackgroundAspectEffectRouter.ios.cs
Axemasta: I've pushed a new commit. Interestingly when I cancel the token, dispose of it and new up a fresh instance it causes a wierd scenario where `Element` becomes null after going past the `Element == null` check and throwing a `NullReferenceException`. I thought once you cancelled a token you should dispose of it and get a new one, but it seems to work better without doing this... 🤨

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/BackgroundAspectEffect/BackgroundAspectEffectRouter.ios.cs
Axemasta: So that it can be cancelled if the `OnElementProperty` fires in quick succession

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/BackgroundAspectEffect/BackgroundAspectEffect.shared.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace Xamarin.CommunityToolkit.Effects
+{
+	public class BackgroundAspectEffect
+	{
+		public static readonly BindableProperty AspectProperty =
+			BindableProperty.CreateAttached("Aspect", typeof(Aspect?), typeof(BackgroundAspectEffect), null,
+				propertyChanged: OnAspectChanged);
+
+		public static Aspect GetAspect(BindableObject view)
+			=> (Aspect)view.GetValue(AspectProperty);
```
pictos, line 15: Shouldn't this be `Aspect?`

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/BackgroundAspectEffect/BackgroundAspectEffect.shared.cs
```diff
@@ -0,0 +1,36 @@
...
+
+		public static void SetAspect(BindableObject view, Aspect value)
+			=> view.SetValue(AspectProperty, value);
+
+		static void OnAspectChanged(BindableObject bindable, object oldValue, object newValue)
+		{
+			if (!(bindable is ContentPage contentPage))
```
pictos, line 22:
```suggestion
			if (bindable is not ContentPage contentPage)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/BackgroundAspectEffect/BackgroundAspectEffect.shared.cs
```diff
@@ -0,0 +1,36 @@
...
+				return;
+
+			var oldEffect = contentPage.Effects.FirstOrDefault(e => e is BackgroundAspectEffectRouter);
+
+			if (oldEffect != null)
+				contentPage.Effects.Remove(oldEffect);
+
+			if (!((Aspect?)newValue).HasValue)
+				return;
```
pictos, line 31: what about?
```suggestion
			if (newValue == null)
				return;
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/BackgroundAspectEffect/BackgroundAspectEffectRouter.ios.cs
```diff
@@ -0,0 +1,159 @@
...
+		async Task ApplyBackgroundImageAsync()
+		{
+			var contentPage = Element as ContentPage;
+
+			if (Element == null || contentPage == null || contentPage.BackgroundImageSource == null)
+				return;
+
+			var image = await GetBackgroundImageFromSource(contentPage.BackgroundImageSource);
+
+			if (image is null)
```
pictos, line 54: we don't use pattern matching for null check
```suggestion
			if (image == null)
```

pictos:
@Axemasta are you able to move forward with this PR? It's fine you can't.

Axemasta:
I need to implement the same for Android, I made an attempt before submitting this PR but didn't get particularly far. I'll have another go this weekend and try and get it update! :)

Axemasta:
I've had a go at getting it to work on Android, I was able to get it working using a `PageRenderer` in a sample project but that required me to use `Xamarin.Forms` views to do any manipulation of the layout. I was unable to use the native API to insert anything that would actually display. `this.AddView(...);` does not appear to add anything, not does `this.AddViewInLayout(_imageView, 0, layout);` appear to have any effect.

I have tried setting a background drawable which does work, and is what the `Xamarin.Forms` renderer will do for `BackgroundImageSource`, however this prevents me from being able to apply any content layout. That is why I thought to add an `Android.Widget.ImageView` to the view and use the `ScaleType` property to control the content formatting. The current options I see are the following:

- Manipulate the background image `Bitmap` to create the scaling manually
- Render the `ImageView` as a `Drawable` and set that as the background
- Have someone with a better understanding of the Android platform explain how I should insert a sub view
- Abandon this effect for Android (not ideal)

There is also another limitation being that I cannot resolve the current `Context` from the toolkit library. This makes any approach with an `ImageView` unattainable since to new up an `ImageView` you need the current content.

Feel free to look at the code, its work in progress and quite frankly very ugly but so it battling with the Android platform 😂😭

Any thoughts @pictos ?

-

pictos:
Hey @Axemasta
> There is also another limitation being that I cannot resolve the current Context from the toolkit library. 

To resolve the context you can use the global Context provided by Toolkit, you can see [it here](https://github.com/xamarin/XamarinCommunityToolkit/blob/main/src/CommunityToolkit/Xamarin.CommunityToolkit/Startup/ToolkitPlatform.android.cs)

For other topics not sure what will be the best approach... Did you try any of those or are they theories?

Axemasta:
Ah global context thats perfect! I'm used to either having it automatically resolved or using `Xamarin.Essentials` to grab it 😃

The 2 theories about rendering the `ImageView` as a Drawable I haven't tried. If I set the background to `ImageView.Drawable` the image gets set on the page, but without the `ScaleType` applied since I've accessed the underlying image, not the view containing it.

For Android i'm not particularly familiar with drawing layouts, I've only done native iOS. Are there any methods I need to call to make Android update the view? When I insert the `ImageView` as a subview into the `ViewGroup` I can see the count of objects increase, could it be i need to invalidate the UI and trigger a redraw?

pictos:
@Axemasta I don't know the answer, I would need to test. Let me know if need me to try those, I can find sometime next week

brminnick:

PR #1437, opened 2021-06-27
Snackbar corner radius

### LINKED ISSUE #1399: Toast and SnackBar CornerRadius
Hi,

Please add CornerRadius to the Toast and SnackBar to set the radius for the corners

CornerRadius="0,25,0,25"

Thanks

VladislavAntonyuk:
### Description of Change ###

Added corner radius property

image

### Bugs Fixed ###

- Fixes #1399 

### API Changes ###


jfversluis:
Thanks for this @VladislavAntonyuk ! I see you only updated 1 sample to use a custom corner radius, yet _all_ the buttons seem to show some different behavior now. Is that expected? I would say that is not desirable and a breaking change. People expect it to look a certain way now, we should try and maintain that while adding new possibilities.


https://user-images.githubusercontent.com/939291/123633568-a8a37a00-d819-11eb-81d8-71b9a6aad7b0.mp4

VladislavAntonyuk:
@jfversluis found the issue. I incorrectly initialized CGRect. It should work now:
image

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/Snackbar/SnackBar.android.cs
```diff
@@ -27,11 +28,21 @@ internal async ValueTask Show(VisualElement sender, SnackBarOptions arguments)
...
 			}
 
+			var shape = snackBar.View.Background as GradientDrawable;
 			if (arguments.BackgroundColor != Forms.Color.Default)
 			{
-				snackBarView.SetBackgroundColor(arguments.BackgroundColor.ToAndroid());
+				shape?.SetColor(arguments.BackgroundColor.ToAndroid().ToArgb());
 			}
 
+			shape?.SetCornerRadii(new float[]
```
maxkoshevoi, line 37: Wouldn't this break default SnackBar style on Android?

https://material.io/components/snackbars#specs

VladislavAntonyuk: It wouldn't break if the default corner radius is set to 4.
```xml
 <shape android:shape="rectangle"
    xmlns:android="http://schemas.android.com/apk/res/android">
    <solid android:color="#323232"/>
    <corners android:radius="4dp"/>
</shape>
```

I've reverted the changes and set it to 4 by default

maxkoshevoi: Compared behavior before and after this PR: 

image

Everything seems fine, but corner radius too small (it's actually set to 4, I checked). Default value feels more like `10` to me.
Here's code from this PR, but default is 10:

image

It's possible that default value is dependent on pixel density or something.

VladislavAntonyuk: Set it to 10, I don't think it is a big issue, because customers can update it to any value. The main idea that it is now consistent between all platforms.

maxkoshevoi: Thank you for setting default to 10. Yes, but to set it to 10 otherwise would require passing an `options` argument to every snackbar, which is no very convenient. 

Could you also not apply corner radius at all if it's a default one (same thing you do with color)?
Some customers may already have their configuration in `xml`, or maybe this default wouldn't work correctly in some cases. In any way, this would be less of a breaking change if it wouldn't be applied unless explicitly requested:

image

jfversluis: I kind of agree. I think we should not set a default at all tbh. We should preserve the corner radius as it was before this change, but just add the ability for the user to now change it. That way, if a user upgrades there is no change unless they explicitly start using this option. Else, there will be a visual change, users will be frustrated and they have to implement code to set it back to how they wanted it to be.

maxkoshevoi: @jfversluis The thing is, Android has good default value (4dp), but all other platforms didn't have one at all (since they don't have native SnackBar control). And the issue was that SnackBar is looking bad on iOS.

Maybe we can make default exactly match Android's one, and then apply it to all platforms for consistency.
If you think, this would be too much of a breaking change, this can be released with next major version.

jfversluis: That is the problem with library development. Once it's in, every change in default behavior is a breaking change. In this case it's a rather small one, but still. People now might have custom code in place to overcome this, if we suddenly start changing default values that might have unexpected behavior that people will get just by upgrading a library and they will have a hard time, spending valuable time, figuring out why that is.

Again, this is a pretty small one so I can let this one slide I guess, but then let's at least preserve the default behavior on Android. Although I don't see why we can't just set no default besides the fact that it will look nicer out of the box.

VladislavAntonyuk: To satisfy all opinions: Set Corner radius to the Android default value (4dp). Calculate density and convert the value to pixel so the radius displayed correctly.

jfversluis: But that's only for Android now, right?

VladislavAntonyuk: Other platforms didn't have the "default" corner radius. So to be consistent between all platforms (Android as etalon) all other platforms will have a corner radius by default. To be honest, who doesn't like the rounded corners. Even Windows 11 has it now))

jfversluis: We're just gonna roll with it, let's do it!

jfversluis (approved):

Cfun1:
@VladislavAntonyuk  @jfversluis  Has the new property introduced in this pr has been documented? 
Explicitly set corner radius values (overriding  the default ones) to show the usage would be a nice addition to the sample too.

VladislavAntonyuk:
> @VladislavAntonyuk  @jfversluis  Has the new property introduced in this pr been documented? 
> Explicitly set corner radius values (overriding  the default ones) to show the usage would be a nice addition to the sample too.

The sample is updated, I will update docs repo a bit later

Issue #1447, opened 2021-06-30
[Enhancement] More complete set of animations


bijington:
## Summary

Provide a larger set of animations to the toolkit.

Please advise on how it would be best to receive these changes. I currently count 97 animations across 16 groupings from the linked page: https://animate.style. Perhaps supplying an initial grouping (e.g. `Attention seekers`) would serve as a good starting point? I appreciate that supplying all those changes in a single hit could make reviewing more complex. Some of the much more complex ones I may in fact struggle with in general anyway (`lightSpeedInRight`).

## API Changes

I currently plan to make a few additions to hopefully keep the potential for this feature open to the consumer of the toolkit. For the purpose of showing my planned changes I will simply refer to a single animation addition for now `tada`.

### VisualElementExtension

Add new extension method(s) to enable invoking the animation directly on the `VisualElement`:

```csharp
// Calls to the below method but offers the simplicity of invoking without `IEnumerable`.
public static void AnimateTada(this VisualElement visualElements)

// This allows us to commit a single Animation for any number of elements.
public static void AnimateTada(this IEnumerable<VisualElement> visualElements)
```

There will also be the potential to add further parameters in order to be able to fine tune the animation for things like `Easing`, `duration`, or maximum/minimum values for scales or rotation angles.

### Behaviors / Animations / AnimationTypes

Additions of new `AnimationBase` implementations to allow for the animations to be defined in XAML and invoked in the usual manner. 

```csharp
public class TadaAnimation : AnimationBase
```

Then these rather light implementations would simply invoke the extension methods defined above.

Currently a number of my animations utilise the `Animation` class in order to achieve some of the more complex sequences, I assume this is fine to fit within the existing `AnimationBase` approach?

## Intended Use Case

To make all our apps look as pretty and fun as possible 😄 

## Who Will Do The Work?

- [x] I am willing to take this on myself
- [ ] Just putting this out there for someone to pick up

## Discussed in https://github.com/xamarin/XamarinCommunityToolkit/discussions/1441

<div type='discussions-op-text'>

<sup>Originally posted by **bijington** June 28, 2021</sup>
I have recently implemented some animations based on this CSS site:
https://animate.style

Tada being a big favourite now (see attached)

https://user-images.githubusercontent.com/17139988/123698685-5896d880-d856-11eb-976c-b4f208e5b73b.mov

I wanted to investigate whether you would like a more complete set of animations?

[...]

PR #1468, opened 2021-07-04
Add tada animation

bijington:
### Description of Change ###

Initial attempt at creating a fluent based animation as discussed in #1447

This is still very much a draft. @pictos please let me know your thoughts on this approach. I do have a couple of points that I am not sure I am happy with/sure on so it would be good to get another opinion:
1. The general use of the factory, we could now quite easily just `new TadaAnimation`.
2. The `AnimationExtensions`, these have been added to keep the fluent syntax less noisy/possible. I am wondering whether those shared options should just be constructor/factory arguments (e.g. `length`).

The items I still need to address once the approach has been finalised are:

- [ ] XML documentation
- [x] `AnimationType` behavior implementation
- [ ] Unit testing
- [ ] Samples page

Here is a sneak peek of the tada in action:

https://user-images.githubusercontent.com/17139988/124399593-339ddc00-dd14-11eb-9a86-22b88c214ad4.mov

### Bugs Fixed ###

- Fixes #

### API Changes ###

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
``` -->

### Behavioral Changes ###

### PR Checklist ###

### CODE REVIEW

**Comment by pictos** [comment]
@bijington I'll take a look at this ASAP and let you know my thoughts. But this almost what I have in mind

**Comment by pictos** [comment]
@bijington I was playing here with other ways to do it, I tried to use the Builder pattern, but when I needed to set the specific values (like Rotation) that becomes harder to implement, that would be easy if we can use the `switch pattern` (or any feature like that) to infer a type to a local variable, something like:
```csharp
public AnimationBuilder WithRotationAngle(double angle)
		{
			var animation = animationType.Name switch // this doesn't compiles
			{
				nameof(TadaAnimation) => (TadaAnimation)instance,
				nameof(TestAnimation) => (TestAnimation)instance,
				_ => null
			};

			animation?.WithRotationAngle(angle);
			return this;
		}
```
Said that and agree with your suggestion, let's keep it as simple as possible

> 1. The general use of the factory, we could now quite easily just new TadaAnimation.

Now makes sense to have those Animations as object and make them `public`, and we don't need to care about the item 2, that you mentioned. With that I have a new question for you. 

Should those Animations be immutable?

bijington:
@pictos I think you have just drastically simplified this whole change with that last question! Yes I believe they should be immutable. Excellent question!

Given this it could mean we don't need the fluent pattern at all and everything could just go in through the constructor. It could now look like:

```csharp
new TadaAnimation(rotationAngle: 30, views: label).Commit();
```

### Additional thought on immutability

While I do think there is great value in making these immutable I would like to ask some questions in return.
1. Should we consider only some parts are immutable (e.g. `rotationAngle`) whereas things like `length` should be mutable?
2. The above makes me question whether the encapsulation of the `Animation` implementation inside this new class is completely wise. It currently removes the ability to set `length` or similar values when calling `Commit`. Perhaps we should consider opening it up more or at least providing the ability to set the mutable parts via the `Commit` method as the `Animation` class allows?
e.g.
```csharp
new TadaAnimation(rotationAngle: 30, views: label).Commit(length: 500, easing: Easing.SpringIn);
```

### Just a response to the builder pattern

With the builder pattern I think we would have needed to expose specific builders for the specific animations then we would be able to simply supply the methods (e.g. `WithRotationAngle`) to only the animations that allow it. This drastically complicates things for very little if any gain.

pictos:
@bijington Ok, so let's move with the ctor approach.

> Should we consider only some parts are immutable (e.g. rotationAngle) whereas things like length should be mutable?

If we go with `immutable` objects. I would say that all props should be immutable as well, that way we can avoid confusion and questions like "Why this prop is mutable and this one no?". 

I forgot about XAML, btw do you think to provide a XAML implementation for this or will this be just C#? If yes, how immutable will look like in XAML? Because AFAIK is very hard to play with ctor that has parameters in XAML, so should we move to the `init` approach `public uint Length {get; init;}`?

I believe that we are in a better place than when we started this conversation about the API. After we define how the XAML part will look, we will have this ready to implement. I'm very cautious with this because I can see in a near feature the community adding more pre-build animations into XCT, so I want to make it easy to implement and maintain.

bijington:
@pictos I completely agree that you want this to end up easy to implement and maintain. I had originally thought we could have the animation implementation in C# and then an additional `AnimationTypes` Behavior that could be used within XAML. This would allow for the attaching for the view that wish to animate. This would then result in some duplication though so I am happy to explore whether there is an option to avoid this. I don't currently have any initial ideas so will have to go away and think unless you have any thoughts?

Another quick question, have I messed up with the destination of this PR? Should it be going in to `develop`?

pictos:
Thanks for moving this forward, if you want you to join the discord server, when we can have a `real-time` chat, and even if you want we try this in a streaming (:

> Another quick question, have I messed up with the destination of this PR? Should it be going in to develop?

@bijington, now that you mentioned... Yes, new features are against `develop` branch

bijington:
@pictos no problem! I'm just happy to be contributing to such a great library!

> Thanks for moving this forward, if you want you to join the discord server, when we can have a `real-time` chat, and even if you want we try this in a streaming (:
> 

Is this the dotnet Discord? I used to think I wasn't going to be up for streaming but I'll give it a go if it helps 😃. There's only one way to find out if it suits me right?

pictos:
@bijington no pressure about the streaming thing haha, yes it's the dotnet Discord, here [is the link](https://discord.gg/eFprASXV)

pictos:
Merging this PR in a feature branch, to make it easy to work on this feature, when this is in a good shape I'll open a PR against `develop`

PR #1527, opened 2021-07-21
Feature/animation samples

### LINKED ISSUE #1447: [Enhancement] More complete set of animations

bijington:
### Description of Change ###

This is very much a work in progress but it is here to show some progress on the pre-built animations along with a samples section.

https://user-images.githubusercontent.com/17139988/126566776-b8aa136a-2543-4b87-b154-6776bba54790.mov

### Bugs Fixed ###

- Fixes #1447

### API Changes ###

Added: 
 
- `void Class.Method();`

Changed:

```csharp
}
``` -->

### Behavioral Changes ###

### PR Checklist ###

### CODE REVIEW

File: src/CommunityToolkit/Xamarin.CommunityToolkit/Animations/TadaAnimation.shared.cs [hunk 1]
```diff
@@ -5,7 +5,7 @@ namespace Xamarin.CommunityToolkit.Animations
 {
     public class TadaAnimation : AnimationBase
```
bijington, line 6: If happy with the approach in `TadaAnimationType` this entire class and folder can disappear

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/Animations/AnimationTypes/AnimationBase.shared.cs
```diff
@@ -32,6 +34,27 @@ static object GetDefaultDurationProperty(BindableObject bindable)
...
+				rate,
+				Duration,
+				Easing,
+				onFinished,
+				shouldRepeat);
+
+		protected virtual Animation CreateAnimation(params View[] views)
+		{
+			return new Animation();
+		}
```
pictos, line 57: shouldn't this be `abstract`?

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/Animations/AnimationTypes/TadaAnimationType.shared.cs
```diff
@@ -34,33 +34,64 @@ public double MinimumScale
...
+
+					taskCompletionSource.SetResult(c);
+				},
+				shouldRepeat: () => IsRepeated,
+				view).Commit();
+
+			return taskCompletionSource.Task;
+		}
+
+		protected override Animation CreateAnimation(params View[] views) => Create(RotationAngle, MinimumScale, MaximumScale, views);
```
pictos, line 71: Here we have a situation. If you do something like

```csharp
		public Test()
		{
			var tada = new TadaAnimationType();
			tada.IsRepeated = true;
			tada.MaximumScale = 1;
			var lbl = new Label();
			var entry = new Entry();

			// This works with more than one view
			var animation = tada.CreateAnimation(16, null, null, lbl, entry);
			animation.Commit();
			animation.Abort();

			// This just works with one view
			tada.Animate(lbl); // Task
		}
```

This can lead to confusion about which method should I call? Just `Animate` or `CreateAnimation`?
Also, if I call `Animate` I can't `Abort` it.

pictos (commented):
@bijington sorry for the delay

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/Animations/AnimationTypes/TadaAnimationType.shared.cs
pictos: So, if we want to use the BehaviorAnimation API, we need to make those animations simpler and drop the `CreateAnimation`, `Commit`, and `Abort` methods, which works for me.

If you see that those methods, at least  `Commit`, and `Abort`, we can :

1. Try to add those methods in BehaviorAnimation API;
2. Create another API, to hold those animations.

Just to point out, that these animations are more complex than the animations that we have, so makes sense to be another implementation if the existing one doesn't fit well.

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/Animations/AnimationTypes/AnimationBase.shared.cs
bijington: Yes good spot!

src/CommunityToolkit/Xamarin.CommunityToolkit/Behaviors/Animations/AnimationTypes/TadaAnimationType.shared.cs
bijington: I completely agree on the confusion this creates. What do you think to keeping the BehaviorAnimation support but creating a second underlying type that BehaviorAnimation can use?

Currently we have AnimationBase, could this stay but then we create 2 abstract classes inheriting from this; SimpleAnimation and ComplexAnimation (or possibly AsyncAnimation and Animation). Then the BehaviorAnimation class could call them based on which is supplied.

Or an alternative could be to bend the complex to work in the existing approach by allowing a CancellationToken to be passed in for the aborting. This all feels a bit wrong considering non of the animations are actually async.

bijington: I guess there are good use cases for both async and non async still. I was going to suggest we could replace the async parts with implementing these directly in an on async manner: https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/ViewExtensions.cs but I am not sure that is a good idea now

bijington:
@pictos i have noticed that the other CI builds aren't running against this PR. I assume it is because it isn't going in to develop or main. Is this ok? I guess it will be run when we eventually merge to develop

pictos:
@bijington you're correct. The CI will just run when we do a merge to `main` or `develop`

bijington:
@pictos what do you think to adding a `CancelCommand` to the new `PreBuildAnimation` base class to allow for a developer to cancel a repeating animation? I can't decide if it covers any extra scenarios or is cleaner than just setting `IsRepeating = false`.

pictos:
> @pictos what do you think to adding a `CancelCommand` to the new `PreBuildAnimation` base class to allow for a developer to cancel a repeating animation? I can't decide if it covers any extra scenarios or is cleaner than just setting `IsRepeating = false`.

I would say that use the `IsRepeating` will be cleaner. Because that BP can be used in triggers, Bindings, converters, etc

bijington:
@pictos is there anything you need from me in order to get this PR in? I need to write up some documentation but I assumed I could do this when we eventually merge in to `develop`?

pictos:
@bijington thanks for you hard work, I'll update this branch and open the official PR ASAP.

bijington:
> @bijington thanks for you hard work, I'll update this branch and open the official PR ASAP.

Awesome. Thank you for this

PR #1604, opened 2021-08-29
Added a sample app

pictos:
This isn't perfect yet, I found a couple of issues during the migration, most of the bugs from the .netMaui, I'll report those bugs ASAP.  This will help us to debug and fix lib errors quickly.

Right now you need to set the MainPage to the page that you want to test, the navigation doesn't work very well.
I didn't do the WinUI integration

List of know issues:
- ReturnResultPopup error source generator;
- App.xaml error breaks all `InitializeComponent` adding too much noise in the build output and errors list
- Resx files cause issues on net6-ios output
- ControlTemplate rendering issues


samples/XCT.MauiCompat.Sample/Directory.Build.props
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
pictos, line 2: This is a hack to avoid the project to import the SDK Build.props. This was causing a lot of issues

PR #1714, opened 2021-11-11
Adding Source Generator

### LINKED ISSUE #1311: [Bug]Error:  : XamlC error XFC0000 : Cannot resolve type "Shield"
### Steps to Reproduce

1.   Installed Xamarin Community Toolkit (v 1.1.0) in All projects (iOS, Android, Main)
2.   Insert xmlns:xct="http://xamarin.com/schemas/2020/toolkit" in xaml page
3.  Insert <xct:Shield x:Name="Shield1"
                Subject="Software Version Number"
                Status="on discord"
                StatusBackgroundColor="SeaGreen"
                StatusTextColor="White"
                Margin="2,0"
               HeightRequest="30"
               HorizontalOptions="Start"/>
in xaml

### Expected Behavior
Expect Shield to Appear without error and intelisense to work in xaml.cs page

### Actual Behavior
intelisense does not work in xaml.cs page -- intelisense indicates Shield not found.  Further when compiling, xaml produces error above -- Error:  : XamlC error XFC0000 : Cannot resolve type "Shield".  
Interestingly, the shield does appear in app when run.
### Basic Information

Would appreciate any suggestions to resolve. None found in searching so far.

pictos:
### Description of Bug ###

This PR adds a Source Generator in our project and nuget, with this we can call an `Init` method and hopefully avoid that fake XAML error.

I didn't test on Mac, so please try it on Mac before merging.

### Issues Fixed ###

- Fixes #1311

### Behavioral Changes ###

This shouldn't change anything in the lib.

### PR Checklist ###


brminnick (commented):
Cool! I know very little about Source Generators, so I have two questions:
- How come it's in its own `csproj`, `Xamarin.CommunityToolkit.SourceGenerator.csproj`?
- Do we need to commit `Xamarin.CommunityToolkit.SourceGenerator.dll` to the repo? Or is that auto-generated at build time?

pictos:
> How come it's in its own csproj, Xamarin.CommunityToolkit.SourceGenerator.csproj?

Not sure if I follow this question. But for Source Generators (SG) we need to have a separate project that targets `netstandard2.0`. Nuget handles SGs the same way as an `analyzer`.

> Do we need to commit Xamarin.CommunityToolkit.SourceGenerator.dll to the repo? Or is that auto-generated at build time?

We can do both since we will not touch in SG too much (I believe that we will never change anything in there) I added the `.dll` to our repo and reference it. Now if we don't want to do that, we need to build the `Xamarin.CommunityToolkit.SourceGenerator.csproj` and then reference the result to `Xamarin.CommunityToolkit.csproj`.
[Here's a sample](https://github.com/reactiveui/refit/blob/main/Refit/Refit.csproj#L30) of how that would be

pictos:
@brminnick 

> How come it's in its own csproj, Xamarin.CommunityToolkit.SourceGenerator.csproj?

Not sure if I follow this question. But for Source Generators (SG) we need to have a separate project that targets `netstandard2.0`. Nuget handles SGs the same way as an `analyzer`.

> Do we need to commit Xamarin.CommunityToolkit.SourceGenerator.dll to the repo? Or is that auto-generated at build time?

We can do both since we will not touch in SG too much (I believe that we will never change anything in there) I added the `.dll` to our repo and reference it. Now if we don't want to do that, we need to build the `Xamarin.CommunityToolkit.SourceGenerator.csproj` and then reference the result to `Xamarin.CommunityToolkit.csproj`.
[Here's a sample](https://github.com/reactiveui/refit/blob/main/Refit/Refit.csproj#L30) of how that would be

brminnick:
Sounds good to me! 

Is there anyway to make a Unit Test for the [error fixed by this](https://github.com/xamarin/XamarinCommunityToolkit/issues/1311)?

pictos:
I'll investigate that. 
In advance, I'm not sure if it's possible because we will need to have access to the VS errors and see if the ` XamlC error XFC0000` is on the list

brminnick:
Agreed. The only thing I can think of is to make a second sample app that doesn't reference any toolkit stuff in C# (only using XAML). Then we build that sample in our CI/CD pipeline to confirm there's no build errors.

It's not exactly a "Unit Test", but it'd accomplish the sample thing.

src/SourceGenerator/Xamarin.CommunityToolkit.SourceGenerator/XCTGenerator.cs
```diff
@@ -0,0 +1,36 @@
...
+	public class XCTGenerator : ISourceGenerator
+	{
+		const string code = @"
+namespace Xamarin.CommunityToolkit.Initializer
+{
+	sealed class XCTInitCaller
+	{
+		public void CallInit()
+		{
+			Xamarin.CommunityToolkit.Helpers.XCT.Init();
```
VladislavAntonyuk, line 22: what do you think if we add DrawingView.wpf after this PR and call its renderer here?

pictos: Can you refresh my memory about it?

VladislavAntonyuk: There was an issue with Renderer initialization, and as workaround we need to call DrawingViewWpfRenderer.Init()

pictos: Thanks. Yeah I remember that now. I can take a look right after we merge this one.

VladislavAntonyuk (approved):
Can we try not include dll as Max suggested?

pictos:
@VladislavAntonyuk IMHO is better to use the dll instead of binding this project every time. This SG is just generating that code all the time, so we don't need to include it on our build process. Referencing the dll in release mode will be the best approach

brminnick (approved):

tranb3r:
I'm having a small issue with this change. The type Xamarin.CommunityToolkit.Initializer.XCTInitCaller is now present in each assembly of my app. This results in a 'duplicate type' warning when I merge these assemblies (obfuscation step during the release process of my app).
I think it would have been better to generate this class in a namespace that depends on the assembly (for example, XCT.[assembly name]). This would avoid potential conflicts when merging assemblies. 
Do you think it's possible to modify this SG in the next version of XCT?

brminnick:
@pictos FYI 👆

Thanks @tranb3r! Could you report this bug in a new Issue and add a reproduction sample?

pictos:
@tranb3r please open an issue and mark me. IF possible provide a small sample that I can use as reference

Issue #184, opened 2020-08-02
[Spec]  Chart view


krdmllr:
# [The feature] 

[Microcharts](https://github.com/dotnet-ad/Microcharts) is a simple and well known chart library for xamarin and xamarin.forms. While it is simple, its dependency to SkiaSharp bring a big size overhead. By using the new Xamarin.Forms Shapes API, we could create a part of Microcharts to add simple Chart views to this library which would not have the SkiaSharp overhead.

The issue of using the Shapes API is that you have to nest Shapes to create multi-colored Charts and nest Labels to show descriptions since the Shapes API does not support layering.

# API

These are the charts we could port:

* Bar Chart
* Donut Chart
* Line Chart
* Pie Chart
* Point Chart
* Radar Chart
* Radial Gauge Chart
 
# Scenarios

Simple charts to show data.

# Platform Compatibility 

All platforms that are supported by Xamarin.Forms shapeshttps://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/shapes/

# Difficulty : medium

jfversluis:
Yes! Awesome! As already pointed out in the chat, @jsuarezruiz was planning on something like this as well. So, please decide whoever wants to do this or even better; collaborate! :D

PR #1733, opened 2021-11-25
Bugfix Issue #1705 (MAUI: 184): Allow null handling for BaseConverterOneWay<TFrom, TTo>

### LINKED ISSUE #1705: [Bug] Allow handling of null for Convert of BaseConverterOneWay<TFrom, TTo>
In order to match a bug fixing coming to the MAUI Toolkit we should aim to solve it here also. The issue in question is:
https://github.com/CommunityToolkit/Maui/issues/184

### Description

Currently it is not possible to pass in a `null` value to the `BaseConverterOneWay<TFrom, TTo>` converter or any implementations of it due to checks within the base class.

Most important details from the discussion/issue:

- The TargetNullValue of XAML's Converter for bindings is applied to the result of the converter and not to the value before it goes in to the converter.
- Let the BaseConverterOneWay<TFrom, TTo> check for null and fallback to the default value for TFrom instead of throwing an Exception. (Suggested by @pictos )
- > If we went with a blanket always replace nulls with default then that is a behavioral change but is still probably unlikely to have a big impact. If we added a way to opt-in as @Countryen suggested then it would have even less of an impact.
- This would then mean that the actual custom converter's ConvertFrom(TFrom value) is called with default(TFrom) when value is null. We could then add a null check to return something the view can use.

Countryen:
### Description of Bug ###

See this (MAUI) discussion: https://github.com/CommunityToolkit/Maui/discussions/164

Fixes #1705 by adding a new property `AllowsNullOrDefault` to the `BaseConverterOneWay<TFrom, TTo>`.
By overriding this property in a custom Converter with `true`, the converter will not throw an `ArgumentException` when trying to convert a `null` TFrom value, instead it will pass `default` to the `ConvertFrom` method.

Opt-in: Property "AllowsNullOrDefault" returns 'false' by default.

This allows custom converters to handle null cases if required.
Tested by adding new unit tests and running all current unit tests green.

Note: Also contains some small fixes for typos and a missing `</summary>`.

@bijington There was an issue with what we discussed:
The recommended approach did not work because we were talking about `null` and a `Forms.Color` can not be `null` (struct). There was no way for me to test it with the ColorToRgbStringConverter (or any color conversion class). I've created a simple derived class of the `BaseConverterOneWay<TFrom, TTo>` inside the test class and created tests based on them. 

### Issues Fixed ###

- Fixes #1705 
- See also (MAUI): [#184](https://github.com/CommunityToolkit/Maui/issues/184)

### Behavioral Changes ###

No change as the feature/fix is opt-in only.

### PR Checklist ###

- [] Has samples (if omitted, state reason in description)
- [] Rebased on top of main at time of PR


bijington:
@Countryen thank you very much for this effort and very good point about `Color` being a `struct`!

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/BaseConverterOneWay_Tests.cs
```diff
@@ -0,0 +1,97 @@
...
+			public override Color ConvertFrom(string? value) =>
+				value switch
+				{
+					// By setting AllowsNullOrDefault to true, it is possible to handle this case by the derived class.
+					null => Color.Black,
+					"Red" => Color.Red,
+					"Blue" => Color.Blue,
+					_ => throw new ArgumentException($"{value} unknown.")
+				};
+		}
```
bijington, line 31: It would be nice to combine this with the other `ColorNameToColorConverterDisallowingNull` class and simply pass in a value for `AllowsNullOrDefault` to the constructor. Perhaps we could move it to the mocks and name it so it is clearly a mock?

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/BaseConverterOneWay_Tests.cs
```diff
@@ -0,0 +1,97 @@
...
+
+		static IEnumerable<object?[]> GetTestDataWithNull()
```
bijington, line 33: Could this return a value tuple so we can have a fixed number of values and more concrete typing (e.g. `IEnumerable<(string, Color)>`

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/BaseConverterOneWay.shared.cs
```diff
@@ -20,14 +25,17 @@ public abstract class BaseConverterOneWay<TFrom, TTo> : ValueConverterExtension,
...
+		public object? Convert(object? value, Type targetType, object parameter, CultureInfo culture)
 		{
-			if (value is not TFrom valueFrom)
-				throw new ArgumentException($"value needs to be of type {typeof(TFrom)}");
-
 			if (targetType != typeof(TTo) && !(typeof(TFrom) != typeof(string)))
 				throw new ArgumentException($"targetType needs to be typeof {typeof(TTo)}");
 
+			if (AllowsNullOrDefault && value is null)
+				return ConvertFrom(default);
```
bijington, line 34: Perhaps if `AllowsNullOrDefault == false && value is null` we could throw an `ArgumentNullException` and mention `AllowsNullOrDefault` in the message to aid developers in understanding how they could fix the issue.

I think this will affect your unit tests but I believe that is a good thing, `ColorNameToColorConverterDisallowingNullConvertInvalidValuesThrowArgumentException` would throw 2 different exceptions based on the `""` and `null` inputs

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/BaseConverterOneWay_Tests.cs
```diff
@@ -0,0 +1,97 @@
...
+			var colorNameToColorConverterDisallowingNull = new ColorNameToColorConverterDisallowingNull();
+			var result = colorNameToColorConverterDisallowingNull.Convert(value, typeof(Color), null!, CultureInfo.CurrentCulture);
+
+			Assert.AreEqual(result, expectedResult);
+		}
+
+		[TestCase("")]
+		[TestCase("Green")]
+		[TestCase("red")]
+		[TestCase(null)]
```
bijington, line 89: I think this could use your `GetTestDataWithNull` data source?

I am slightly confused by this test and I must be missing something, I read it that `"red"` and `"Green"` should throw an `ArgumentException` but these are valid values. Am I having a meltdown and reading this incorrectly 😃 ?

bijington (changes requested):
I hope these suggestions shouldn't cause much pain/effort but do let me know your thoughts.

Thanks so much for fixing all of those typos too

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/BaseConverterOneWay_Tests.cs
Countryen: "Green" is not defined in the converter.
"red" is lower case, which is only defined as "Red" (Pascal Case) in the converter (the tests are meaningless because we don't need to test the actual converter's logic but wanted them to look like a real test would look like).

If you want I can make it more logical and test for any `Forms.Color` (OrdinalIgnoreCase), but that's a bit more complex, I think.

This can't use `GetTestDataWithNull`, because these are "invalid argument" tests. I can't use Color.* (as expectedResult) in `TestCase`s otherwise I would not have used `TestCaseSource` for the valid tests and just used `TestCase`s.

I could make it more general and return `(string value, Color expectedValue, bool expectsError)` and just use one test (testing both valid and invalid arguments), I've copied most of this from the other converter tests to stay consistent.

bijington: Thanks for explaining it, I think this just shows I probably need a break from my screen haha.

I am fine with keeping this mock converter as simple as possible so ignore my point about the data source.

Countryen:
@bijington Thanks for all recommendations. After all changes it was easier to update/simplify the tests than to keep them as before. Should I update the description of the PR now because the change from ArgumentException to ArgumentNullException (new behavior?)?

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/BaseConverterOneWay.shared.cs
```diff
@@ -20,14 +25,20 @@ public abstract class BaseConverterOneWay<TFrom, TTo> : ValueConverterExtension,
...
-			if (value is not TFrom valueFrom)
-				throw new ArgumentException($"value needs to be of type {typeof(TFrom)}");
-
 			if (targetType != typeof(TTo) && !(typeof(TFrom) != typeof(string)))
 				throw new ArgumentException($"targetType needs to be typeof {typeof(TTo)}");
 
+			if (AllowsNullOrDefault is false && value is null)
+				throw new ArgumentNullException($"value needs to be of type {typeof(TFrom)} but is null. Consider setting {nameof(AllowsNullOrDefault)} to true.");
+
+			if (AllowsNullOrDefault is true && value is null)
```
pictos, line 36: AllowsNullOrDefault is a Boolean, so we don't need to check it
```suggestion
			if (AllowsNullOrDefault && value is null)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/BaseConverterOneWay.shared.cs
```diff
@@ -20,14 +25,20 @@ public abstract class BaseConverterOneWay<TFrom, TTo> : ValueConverterExtension,
...
-		public object? Convert(object value, Type targetType, object parameter, CultureInfo culture)
+		public object? Convert(object? value, Type targetType, object parameter, CultureInfo culture)
 		{
-			if (value is not TFrom valueFrom)
-				throw new ArgumentException($"value needs to be of type {typeof(TFrom)}");
-
 			if (targetType != typeof(TTo) && !(typeof(TFrom) != typeof(string)))
 				throw new ArgumentException($"targetType needs to be typeof {typeof(TTo)}");
 
+			if (AllowsNullOrDefault is false && value is null)
```
pictos, line 33:
```suggestion
			if (!AllowsNullOrDefault && value is null)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/BaseConverterOneWay.shared.cs
bijington: Perhaps something like this might be easier to read?

```csharp
if (value is null)
{
    if (AllowsNullOrDefault)
        throw new ArgumentNull...

    return Convert(default);
}
```

pictos: love it /\

bijington:
> @bijington Thanks for all recommendations. After all changes it was easier to update/simplify the tests than to keep them as before. Should I update the description of the PR now because the change from ArgumentException to ArgumentNullException (new behavior?)?

No I wouldn't worry about updating the description of the PR but thanks for checking

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/BaseConverterOneWay.shared.cs
Countryen: I've changed it but swapped them out (inverted).

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/BaseConverterOneWay_Tests.cs
```diff
@@ -0,0 +1,57 @@
...
+			yield return (true, "red",  typeof(ArgumentException));
+			yield return (true, "Green",  typeof(ArgumentException));
+			yield return (true, "red",  typeof(ArgumentException));
+			yield return (true, "Green",  typeof(ArgumentException));
+
+			yield return (false, null,  typeof(ArgumentNullException));
+		}
+
+		[TestCaseSource(nameof(GetValidTestData))]
+		public void MockConverterOneWayConvert((bool ShouldAllowNull, string? Value, Color? ExpectedResult) testCase)
```
bijington, line 37: I don't think `ExpectedResult` needs to be nullable here does it?

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/BaseConverterOneWay_Tests.cs
```diff
@@ -0,0 +1,57 @@
...
+			var mockConverterOneWay =
+				testCase.ShouldAllowNull ? new MockConverterOneWay(true) : new MockConverterOneWay(false);
+
+			var result = mockConverterOneWay.Convert(testCase.Value, typeof(Color), null!, CultureInfo.CurrentCulture);
+
+			Assert.AreEqual(result, testCase.ExpectedResult);
+		}
+
+		[TestCaseSource(nameof(GetInvalidTestData))]
+		public void MockConverterOneWayConvertInvalidValuesThrowException((bool ShouldAllowNull, string? Value, Type? ExpectedExceptionType) testCase)
```
bijington, line 48: I don't think `ExpectedExceptionType` needs to be nullable here does it?

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/BaseConverterOneWay_Tests.cs
```diff
@@ -0,0 +1,57 @@
...
+			yield return (true, "Green",  typeof(ArgumentException));
+
+			yield return (false, null,  typeof(ArgumentNullException));
+		}
+
+		[TestCaseSource(nameof(GetValidTestData))]
+		public void MockConverterOneWayConvert((bool ShouldAllowNull, string? Value, Color? ExpectedResult) testCase)
+		{
+			var mockConverterOneWay =
+				testCase.ShouldAllowNull ? new MockConverterOneWay(true) : new MockConverterOneWay(false);
```
bijington, line 40: I think this could be simplified down to:

```csharp
var mockConverterOneWay = new MockConverterOneWay(testCase.ShouldAllowNull);
```

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/BaseConverterOneWay_Tests.cs
```diff
@@ -0,0 +1,57 @@
...
+
+			var result = mockConverterOneWay.Convert(testCase.Value, typeof(Color), null!, CultureInfo.CurrentCulture);
+
+			Assert.AreEqual(result, testCase.ExpectedResult);
+		}
+
+		[TestCaseSource(nameof(GetInvalidTestData))]
+		public void MockConverterOneWayConvertInvalidValuesThrowException((bool ShouldAllowNull, string? Value, Type? ExpectedExceptionType) testCase)
+		{
+			var mockConverterOneWay = testCase.ShouldAllowNull ? new MockConverterOneWay(true) : new MockConverterOneWay(false);
```
bijington, line 50: I think this could be simplified down to:

```csharp
var mockConverterOneWay = new MockConverterOneWay(testCase.ShouldAllowNull);
```

Countryen:
Maybe the name `AllowsNullOrDefault` could also be misleading? My thought was that it is similar to `FirstOrDefault()` (LINQ) methods but people could assume that `default(TFrom)` is not possible without?

First time using this review/rework feature of GitHub question: Is there a status symbol or something that allows me to push again? My guess is, that I should wait until review is done before sending a new push? I don't see any symbol or indicator here (just the "1 change requested" since the beginning).

bijington:
> Maybe the name `AllowsNullOrDefault` could also be misleading? My thought was that it is similar to `FirstOrDefault()` (LINQ) methods but people could assume that `default(TFrom)` is not possible without?

This is a very good point! Perhaps we could just drop the OrDefault suffix and go with `AllowsNull`?

> First time using this review/rework feature of GitHub question: Is there a status symbol or something that allows me to push again? My guess is, that I should wait until review is done before sending a new push? I don't see any symbol or indicator here (just the "1 change requested" since the beginning).

You can push up as many commits as you like. I guess technically you can then request a re-review to indicate you are done but we usually get notifications to let us know things have been pushed.

bijington (approved):
This looks great! Thank you for all of this effort!

bijington:
@Countryen I think this is complete now would you agree? If so then we can hit the merge button 😄

Countryen:
@bijington Looks good to me but should at least once be tested with a real app I think but maybe you guys do this before a release anyways so it's in your hand now . I say 🆗 merge.

bijington:
> @bijington Looks good to me but should at least once be tested with a real app I think but maybe you guys do this before a release anyways so it's in your hand now . I say 🆗 merge. 

Perhaps we should create something in the sample app to prove this then.

Countryen:
> Perhaps we should create something in the sample app to prove this then.

Agreed. I think we could add a simple "LanguageToNameConverter" for the `Language` model and give back `$"{l.Name} [{l.CI}]"` or `"No language"` and display that somewhere with XAML on an existing page?

Or would we rather add a separate page that demonstrates the use (null allowed vs. null disallowed)?

I am fine both ways, but I don't know exactly where this would be placed best, could you please show me locations/suggestions?

I am also fine if someone else want's to create a sample-addition for it, I (want to) trust in unit tests 😃

bijington:
I think we can potentially do this with minimal effort... I think you pointed our earlier on that there is only one class that inherits from `BaseOneWayConverter` and I am now suspecting that this change could allow other classes to. Looking at `IsNullOrEmptyConverter` I think this could and now should use the base class and set `AllowsNull` to true. If so then you don't need to add any samples as they already exist.

What do you think?

https://github.com/xamarin/XamarinCommunityToolkit/blob/main/src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/IsNullOrEmptyConverter.shared.cs

Countryen:
Good point. Probably most of the converters should inherit from the BaseConverterOneWay<TFrom, TTo> or BaseConverter<TFrom, TTo> If no special handling of the other arguments is needed, of course.

I think only updating this one converter could seem off, as the other one (IsNotNullOrEmptyConverter) would look different then.
(Same for the ListIsNullOrEmptyConverter and ListIsNotNullOrEmptyConverter)

For now, you mean it'd be enough to simply update the `IsNullOrEmptyConverter` to the new possibility and call it a day, can do that 👍🏻.
I recommend to update all others (where possible) later on to have them consistent again (in MAUI).

bijington:
I'd love to get them all converted 😃 perhaps if you can prove that it works for the first one then I can happily help our and convert (pun not intended) the rest. If you are ok with that? Otherwise I can happily look in to this at some point soon

Countryen:
Sure let us do this 😄 I will start with `IsNullOrEmptyConverter` and then we'll see.

Countryen:
Ah, @bijington could you please mark this as in progress or something? Merge souldn't be possible now again.

bijington:
@Countryen done! I added do not merge just to be safe too 😃

Countryen:
Just working on it, noticed that users should also be aware of this: When using the (updated) `BaseConverterOneWay<TFrom, TTo>` the ConvertFrom(TFrom) must now allow nullabillity in form of `TFrom?`, see the warning/error (depending on nullable reference type settings).

image

Countryen:
@bijington First one is done, tested with all unit tests running green still/again and by using the samples app on UWP.
If you agree on converting the others now too, feel free to tell me which one I can simply "convert". Maybe we can split up the work for converting the others ✂️.

bijington:
> Just working on it, noticed that users should also be aware of this: When using the (updated) `BaseConverterOneWay<TFrom, TTo>` the ConvertFrom(TFrom) must now allow nullabillity in form of `TFrom?`, see the warning/error (depending on nullable reference type settings).
> 
> 
> 
> image

@pictos @brminnick I don't think there is much we can do about this but I wanted to get your thoughts on whether there is anything else we could do with the shift to allowing nulls. The only other thought I now have is whether we have a separate base class that allows nulls but I'm not sure it gives us a big benefit.

bijington:
> @bijington First one is done, tested with all unit tests running green still/again and by using the samples app on UWP.
> 
> If you agree on converting the others now too, feel free to tell me which one I can simply "convert". Maybe we can split up the work for converting the others ✂️.

Great thank you. Sorry I disappeared for a bit. I will try and go over the list of existing converters over the weekend and see what we think should change.

pictos:
> > Just working on it, noticed that users should also be aware of this: When using the (updated) `BaseConverterOneWay<TFrom, TTo>` the ConvertFrom(TFrom) must now allow nullabillity in form of `TFrom?`, see the warning/error (depending on nullable reference type settings).
> > image
> 
> @pictos @brminnick I don't think there is much we can do about this but I wanted to get your thoughts on whether there is anything else we could do with the shift to allowing nulls. The only other thought I now have is whether we have a separate base class that allows nulls but I'm not sure it gives us a big benefit.

To solve this we can do `TFrom?` instead of `TFrom`, does it work?

bijington:
> > > Just working on it, noticed that users should also be aware of this: When using the (updated) `BaseConverterOneWay<TFrom, TTo>` the ConvertFrom(TFrom) must now allow nullabillity in form of `TFrom?`, see the warning/error (depending on nullable reference type settings).
> 
> > > image
> 
> > 
> 
> > @pictos @brminnick I don't think there is much we can do about this but I wanted to get your thoughts on whether there is anything else we could do with the shift to allowing nulls. The only other thought I now have is whether we have a separate base class that allows nulls but I'm not sure it gives us a big benefit.
> 
> 
> 
> To solve this we can do `TFrom?` instead of `TFrom`, does it work?

Yes that would be the change. I think the point @Countryen was making relates to the impact this change would have on anyone using this class as the base for any of their converters. It should be relatively straightforward to fix though

brminnick:
I'm fine with using `TFrom?`. 

It's not necessarily a breaking change. 

It just requires parent classes to perform a null check (which they probably should be doing anyways).

bijington:
OK great. We can make more of the XCT converters use this base class now and then this should be good to merge

Countryen:
> It just requires parent classes to perform a null check (which they probably should be doing anyways).

@bijington Sorry for reiterating so late (weekend).
Just to make sure I explained it correctly, here an example: 

If users previously defined converters like those:
````
public class SomeConverter1 : BaseConverterOneWay<string, Color>
{
	public override Color ConvertFrom(string value) => ...
}

public class SomeConverter2 : BaseConverter<string, Color>
{
	public override Color ConvertFrom(string value) => ...
	public override string ConvertBackTo(Color value) => ...
}
````

After update with setting "treat nullability warnings as errors", **this will not compile** anymore because the `value` doesn't allow null. **Users must change their code after update** to `ConvertFrom(string? value)`. The null check is not necessary as long as they don't return `true` for `AllowsNull`.

We could instead play it safe and remove the nullability and go back to `ConvertFrom(TFrom)` and instead invoke `ConvertFrom(default!)`. 
This allows child classes to define either `ConvertFrom(string value)` or `ConvertFrom(string? value)` . For a "bugfix" this should be preferred, don't you think? Maybe you want to test that when you convert one of the converters and see for yourself 😃 

I think for MAUI it'd be better to define it as `ConvertFrom(TFrom?)`, with nullability by default.

**Comment by bijington** [comment]
@Countryen no problem on the delay, I ended up avoiding the computer after some internet related issues which was nice in the end.

I think I am ok with forcing `TFrom?` as it is closer to what most people will expect in terms of actual implementations.

I am hoping to get on to looking at the converters tomorrow now, hopefully we can fly through them and get it ready for the upcoming release

**Comment by bijington** [comment]
@Countryen I have just pushed a load of converters to now use the base class. I think this covers the set of what we can achieve for now. Would you be willing to reverse roles and review my changes? I also switched from using the protected getter to a constructor parameter as it fits more with the MAUI based approach that we will be using and I don't believe the value should ever need to change.

 Interestingly I think we could convert more if the `ConvertFrom` virtual method passed across more parameters like `targetType` and `parameter`. This may well be out of scope for now though.

I do wonder whether we should look at making the `ConvertBack` in `BaseConverter` symmetrical with these changes we have now made.

**Comment by Countryen** [comment]
> Would you be willing to reverse roles and review my changes?

I've reviewed the changes. some Unit Tests are red, I can fix that or do you want to fix it yourself? Probably because of the new mismatch between `ArgumentNullException` and `ArgumentException`.

I like the change to a constructor argument if that fits better with other MAUI code.
I like the fact that the current `ConvertFrom` only has 1 argument (the value) as I care about this value in most situations.
More arguments makes it more complex, and if I really need a complex one I don't use the/this base class.
(Maybe add more arguments later in MAUI.)

`ConvertBack` should probably also allow `TTo? value` instead of `TTo value`, as you said, seen already in `TimeSpanToDoubleConverter` (`object? value`) , but that should be another issue.

**Comment by Countryen** [comment]
@bijington How do we use constructors with arguments in XAML for the Converter? Could you make an example? I think constructors in XAML are a bit annoying to use, especially when used in Markup like
`<Label Text="{Binding Data.Name, Converter={x:Static xct:SomeConverter}"/>`

**Comment by bijington** [comment]
> @bijington How do we use constructors with arguments in XAML for the Converter? Could you make an example? I think constructors in XAML are a bit annoying to use, especially when used in Markup like
> 
> `<Label Text="{Binding Data.Name, Converter={x:Static xct:SomeConverter}"/>`

My expectation is that this usage shouldn't change. We will only specify the argument in the implementing converter like this:

```csharp
public MyConverter() : base(true)
```

Therefore the converter constructor remains parameter less for use within XAML

Countryen:
@bijington I've added some comments / reviews (if I did this correctly), too. Thanks for the answer about XAML, that was my misunderstanding 😄 

Coming with 2.0.0 sounds awesome if then all converters can share the same base logic.

Countryen (commented):
See my comments (Sorry,  I didn't know that I need to "finish" this with a button first).

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/ListIsNullOrEmptyConverter_Tests.cs
```diff
@@ -22,7 +22,7 @@ public void ListIsNullOrEmptyConverter(object value, bool expectedResult)
 		{
 			var listIstNullOrEmptyConverter = new ListIsNullOrEmptyConverter();
 
-			var result = listIstNullOrEmptyConverter.Convert(value, typeof(ListIsNullOrEmptyConverter), null, CultureInfo.CurrentCulture);
+			var result = listIstNullOrEmptyConverter.Convert(value, typeof(ListIsNullOrEmptyConverter), null!, CultureInfo.CurrentCulture);
```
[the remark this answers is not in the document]
brminnick: Please never use the [null-forgiving operator, `!`](https://docs.microsoft.com/dotnet/csharp/language-reference/operators/null-forgiving)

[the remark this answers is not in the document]
bijington: Yeah sorry this was my bad investigating why it wasn't happy and I missed this class to tidy up :)

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/CompareConverter.shared.cs
```diff
@@ -60,11 +57,8 @@ enum Modes
 		/// Converts an object that implements IComparable to a specified object or a boolean based on a comparaison result.
 		/// </summary>
 		/// <param name="value">The value to convert.</param>
-		/// <param name="targetType">The type of the binding target property. This is not implemented.</param>
-		/// <param name="parameter">Additional parameter for the converter to handle. This is not implemented.</param>
-		/// <param name="culture">The culture to use in the converter.  This is not implemented.</param>
 		/// <returns>The object assigned to <see cref="TrueObject"/> if (value <see cref="ComparisonOperator"/> <see cref="ComparingValue"/>) equals True and <see cref="TrueObject"/> is not null, if <see cref="TrueObject"/> is null it returns true, otherwise the value assigned to <see cref="FalseObject"/>, if no value is assigned then it returns false.</returns>
-		public virtual object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+		public override object ConvertFrom(TObject? value)
```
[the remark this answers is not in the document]
bijington: I am actually going to revert this change as it has a fair amount of further implication

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/IsInRangeConverter.shared.cs
```diff
@@ -42,33 +40,16 @@ public object MaxValue
...
-				throw new ArgumentException("is expected to implement IComparable interface.", nameof(value));
-
 			if (MinValue is not IComparable)
 				throw new ArgumentException("is expected to implement IComparable interface.", nameof(MinValue));
 
 			if (MaxValue is not IComparable)
 				throw new ArgumentException("is expected to implement IComparable interface.", nameof(MaxValue));
 
-			return comparable.CompareTo(MinValue) >= 0 && comparable.CompareTo(MaxValue) <= 0;
+			return value!.CompareTo(MinValue) >= 0 && value.CompareTo(MaxValue) <= 0;
```
brminnick, line 52: Let's remove the null-forgiving operator
```suggestion
			return value.CompareTo(MinValue) >= 0 && value.CompareTo(MaxValue) <= 0;
```

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Converters/IsNullOrEmptyConverter_Tests.cs
```diff
@@ -14,7 +14,7 @@ public void IsNullOrEmptyConverter(object value, bool expectedResult)
 		{
 			var isNullOrEmptyConverter = new IsNullOrEmptyConverter();
 
-			var result = isNullOrEmptyConverter.Convert(value, typeof(IsNotNullOrEmptyConverter_Tests), null, CultureInfo.CurrentCulture);
+			var result = isNullOrEmptyConverter.Convert(value, typeof(IsNotNullOrEmptyConverter_Tests), null!, CultureInfo.CurrentCulture);
```
brminnick, line 17: Let's remove the null-forgiving operator
```suggestion
			var result = isNullOrEmptyConverter.Convert(value, typeof(IsNotNullOrEmptyConverter_Tests), null, CultureInfo.CurrentCulture);
```

brminnick (changes requested):
This isn't a full review. Rather, I found a couple more null-forgiving operators that need to be removed.

Just commenting here so that we don't forget to remove these before merging 👍

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/IsInRangeConverter.shared.cs
bijington: @brminnick this is a slightly annoying introduction of the code changes by allowing nulls in. We know that null is not allowed here before `allowNulls` is `false` in the base class. Do you know of a way to inform the compiler this is the case? Something like `[NotNullWhen(allowsNull == false)]`

brminnick: XCT can't use [Nullable Static Analysis](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis), because XCT targets .NET Standard 1.0 and Nullable Static Analysis is only available in .NET Standard 2.1+.

I'll take a look at the code to see if there's a different way we can architect it to avoid this issue.

brminnick: Here's my recommendation:

Make two abstract base classes, `BaseConverterOneWay` and `NullableBaseConverterOneWay`.

Not related to Nullable, but I also recommend implementing `object? IValueConverter.Convert` implicitly, e.g. making it a private method, since classes inheriting from `BaseConverterOneWay` and `NullableBaseConverterOneWay` shouldn't directly call `object? IValueConverter.Convert`.

```cs
using System;
using System.Globalization;
using Xamarin.CommunityToolkit.Extensions.Internals;
using Xamarin.Forms;

namespace Xamarin.CommunityToolkit.Converters
{
	/// <summary>
	/// Abstract class used to implement converters that implements the ConvertBack logic.
	/// </summary>
	/// <typeparam name="TFrom">Type of the input value</typeparam>
	/// <typeparam name="TTo">Type of the output value</typeparam>
	public abstract class NullableBaseConverterOneWay<TFrom, TTo> : ValueConverterExtension, IValueConverter
	{
		/// <summary>
		/// Method that will be called by <see cref="Convert(object, Type, object, CultureInfo)"/>.
		/// </summary>
		/// <param name="value">Value to be converted from <see cref="TFrom"/> to <see cref="TTo"/>.</param>
		/// <returns>An object of type <see cref="TTo"/>.</returns>
		public abstract TTo ConvertFrom(TFrom? value);

		/// <summary>
		/// Not implemented, use <see cref="BaseConverter{TFrom, TTo}"/>
		/// </summary>
		public virtual object? ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
			=> throw new NotImplementedException("Impossible to revert to original value. Consider setting BindingMode to OneWay.");

		/// <summary>
		/// Converts the incoming value from <see cref="TFrom"/> and returns the object of a type <see cref="TTo"/>.
		/// </summary>
		/// <param name="value">The value to convert.</param>
		/// <param name="targetType">The type of the binding target property. This is not implemented.</param>
		/// <param name="parameter">Additional parameter for the converter to handle. This is not implemented.</param>
		/// <returns>An object of type <see cref="TTo"/>.</returns>
		object? IValueConverter.Convert(object? value, Type targetType, object? parameter, CultureInfo culture)
		{
			if (value is null)
				return ConvertFrom(default);

			if (value is not TFrom valueFrom)
				throw new ArgumentException($"value needs to be of type {typeof(TFrom)}");

			if (targetType != typeof(TTo) && !(typeof(TFrom) != typeof(string)))
				throw new ArgumentException($"targetType needs to be typeof {typeof(TTo)}");

			return ConvertFrom(valueFrom);
		}
	}

	public abstract class BaseConverterOneWay<TFrom, TTo> : ValueConverterExtension, IValueConverter  where TFrom : notnull where TTo : notnull
	{
		/// <summary>
		/// Method that will be called by <see cref="Convert(object, Type, object, CultureInfo)"/>.
		/// </summary>
		/// <param name="value">Value to be converted from <see cref="TFrom"/> to <see cref="TTo"/>.</param>
		/// <returns>An object of type <see cref="TTo"/>.</returns>
		public abstract TTo ConvertFrom(TFrom value);

		/// <summary>
		/// Not implemented, use <see cref="BaseConverter{TFrom, TTo}"/>
		/// </summary>
		public virtual object? ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
			=> throw new NotImplementedException("Impossible to revert to original value. Consider setting BindingMode to OneWay.");

		/// <summary>
		/// Converts the incoming value from <see cref="TFrom"/> and returns the object of a type <see cref="TTo"/>.
		/// </summary>
		/// <param name="value">The value to convert.</param>
		/// <param name="targetType">The type of the binding target property. This is not implemented.</param>
		/// <param name="parameter">Additional parameter for the converter to handle. This is not implemented.</param>
		/// <returns>An object of type <see cref="TTo"/>.</returns>
		object? IValueConverter.Convert(object? value, Type targetType, object? parameter, CultureInfo culture)
		{
			if (value is null)
				throw new ArgumentNullException(nameof(value), $"value needs to be of type {typeof(TFrom)} but is null. If type {typeof(TFrom)} should be nullable, inherit from {nameof(NullableBaseConverterOneWay<TFrom, TTo>)} instead");

			if (value is not TFrom valueFrom)
				throw new ArgumentException($"value needs to be of type {typeof(TFrom)}");

			if (targetType != typeof(TTo) && !(typeof(TFrom) != typeof(string)))
				throw new ArgumentException($"targetType needs to be typeof {typeof(TTo)}");

			return ConvertFrom(valueFrom);
		}
	}
}
```

brminnick: Then update the Converters accordingly. 

For example, `ListIsNullOrEmptyConverter` will become this:

```cs
using System.Collections;

namespace Xamarin.CommunityToolkit.Converters
{
	/// <summary>
	/// Converts the incoming value to a <see cref="bool"/> indicating whether or not the value is null or empty.
	/// </summary>
	public class ListIsNullOrEmptyConverter : NullableBaseConverterOneWay<IEnumerable, bool>
	{
		/// <summary>
		/// Converts the incoming value to a <see cref="bool"/> indicating whether or not the value is null or empty.
		/// </summary>
		/// <param name="value">The value to convert.</param>
		/// <returns>A <see cref="bool"/> indicating if the incoming value is null or empty.</returns>
		public override bool ConvertFrom(IEnumerable? value) => ConvertInternal(value);

		internal static bool ConvertInternal(IEnumerable? value)
		{
			if (value == null)
				return true;

			return !value.GetEnumerator().MoveNext();
		}
	}
}
```

bijington: Thanks I was literally writing to say the only alternative I could think of was an alternative base class or just allowing nulls everywhere which is obviously not ideal. I will have a play with this approach but I think the separate base class makes the most amount of sense.

I really like the idea of the explicit interface implementation!

brminnick: And for the .NET MAUI Community Toolkit, we can likely architect it a bit differently, taking advantage of Nullable Static Analysis and additional goodies in .NET 6 👍

Countryen: 2 small notes for this:
- I think abstract classes should start with "Base" for easy searching so I'd prefer to read "**Base**NullableConverter" over "Nullable**Base**Converter".
- I think abstract classes should end with the actual class name (because you'd speak "one way converter") so I'd prefer to read "NullableBaseOneWay**Converter**" over "BaseNullable**Converter**OneWay"

Also, does that also mean we get a "NullableBaseConverter" separate from the "BaseConverter" (two way)?

Countryen:
@bijington I lost track a bit so please notify me whenever there is a chance to further contribute or my input is required. Looking forward to the results, looks very promising (with or without new base classes).

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/IsInRangeConverter.shared.cs
bijington: Yes very good point there!

bijington:
@Countryen sorry this chunk of work has been a little bigger and more involved than I think we both initially expected.

I have gone ahead and modified some bits further to make use of 2 base classes to avoid too much punctuation in the code `? !`.

This means we won't have to release any breaking changes as part of this and we can ship it sooner than 2.0.0.

src/CommunityToolkit/Xamarin.CommunityToolkit/Converters/DateTimeOffsetConverter.shared.cs
```diff
@@ -1,43 +1,30 @@
...
-		/// <returns>The <see cref="DateTime"/> value.</returns>
-		public object Convert(object? value, Type targetType, object? parameter, System.Globalization.CultureInfo culture)
-			=> value is DateTimeOffset dateTimeOffset
-				? dateTimeOffset.DateTime
-				: throw new ArgumentException("Value is not a valid DateTimeOffset", nameof(value));
+		/// <returns>The <see cref="DateTimeOffset"/> value.</returns>
+		public override DateTimeOffset ConvertBackTo(DateTime value) =>
+			value.Kind switch
+			{
+				DateTimeKind.Local => new DateTimeOffset(DateTime.SpecifyKind(value, DateTimeKind.Unspecified), DateTimeOffset.Now.Offset),
```
bijington, line 18: I am open to anyone and everyone reviewing this change. I believe this solves our failing unit test when running locally based on this documentation: https://stackoverflow.com/a/36255983/32348

I suspect it passed on the CI as the agents are likely running in UTC and therefore the local test is no different to the UTC test.

Potentially also solves https://github.com/xamarin/XamarinCommunityToolkit/issues/1744

@Countryen if you grab the latest changes does the DateTimeOffsetConverter_Tests all pass now?

Countryen: @bijington  Yes, all (including all DateTimeOffsetConverter_Tests) tests now pass green for me ✔️

bijington:
@pictos I think this could now go in to the next release as we are no longer shipping any breaking changes as part of this fix. I will write up the breaking changes for another PR. Would you be willing to take a look over the changes and see if we have done anything crazy 😄

src/CommunityToolkit/Xamarin.CommunityToolkit.UnitTests/Mocks/MockConverterOneWay.cs
```diff
@@ -0,0 +1,35 @@
...
+				_ => throw new ArgumentException($"{value} unknown.")
+			};
+	}
+
+	/// <summary>
+	/// Mock to test the <see cref="BaseConverterOneWay{TFrom, TTo}"/> internals.
+	/// </summary>
+	public class MockConverterOneWay : BaseConverterOneWay<string, Color>
+	{
+		public override Color ConvertFrom(string? value) =>
```
Countryen, line 27: @bijington Shouldn't it be `string value` here (without nullability)?

bijington: Very good spot!

bijington:
@brminnick I requested a re-review as GitHub is claiming you have a change requested but I can't find the detail anywhere

brminnick (approved):
Thanks @Countryen + @bijington!! 🙌

Issue #1231, opened 2021-04-24
[Bug] TabView SelectedIndex


Saleh-Software:
hello im coding this:

` <xct:TabView x:Name="tbView"  Style="{StaticResource CustomTabStyle}" SelectedIndex="1" >

          <xct:TabViewItem /> 
           <xct:TabViewItem /> 
           <xct:TabViewItem /> 
        
</xct:TabView>
`
i didnt show tabitem content its just example.

and when i will run getting this error!:

System.ArgumentOutOfRangeException: 'Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index'

pictos:
@Saleh-Software 
Could you attach a small repro?

Saleh-Software:
Sure !

this is small check code behind also

idk how zip my project like you its low size but for me its 50mb 
so i uploaded in file upload 

link for download: https://ufile.io/fjnhm6ms

pictos:
@Saleh-Software I'll take a look during the week. To reduce the size of the sample, delete the obj and bin folders before zip

Issue #979, opened 2021-02-25
[Bug][iOS] Binding to TabView.SelectedIndex cause that TabView doesn't work


SeRgI1982:
### Description

I would like to have a chance preselect different TabItem than the first on when navigate to page with TabView.
Unfortunately, it doesn't work when you bind SelectedIndex to your ViewModel and set a proper index inside ViewModel. 
The control stops working - there is no chance to navigate to different TabItems when this Binding exists.

### Steps to Reproduce

1. Create Page with TabView and a few TabViewItem's
2. Bind TabView.SelectedIndex to your ViewModel.SelectedTabIndex
3. Try to set ViewModel.SelectedTabIndex to correct value

### Expected Behavior

TabView will navigate automatically to an appropriate TabViewItem 

### Actual Behavior

On iOS TabView stuck - there is no chance to navigate to other tabs - even by tapping other TabViewItem icons. 
It worked in version 1.0.0

### Basic Information

- Version with issue: 1.0.3 like also 1.1.0-pre1
- Last known good version: 1.0.0
- IDE: VS for Mac
- Platform Target Frameworks: 
  - iOS:  Xcode Version 12.4 (12D4e)
  - Android: 10.0 

Sample app below:

[TabViewTest.zip](https://github.com/xamarin/XamarinCommunityToolkit/files/6042558/TabViewTest.zip)

Cfun1:
Seems the same issue as https://github.com/xamarin/XamarinCommunityToolkit/issues/870 isn't supposed to be fixed @jfversluis ?

jfversluis:
@jsuarezruiz

Kapusch:
Actually, I can provide the same comment as for #1012 

I also have noticed this bug many times, but I cannot say why exactly.
Anytime I sign out / sign in again to my app, then I can see my tab well positioned on the value I have initialized in the ViewModel.

In my XAML file, the property binding is set as followed :
```
        SelectedIndex="{Binding SelectedTopTabsIndex, Mode=TwoWay}"
``` 
And the value is configured in the ViewModel as followed :
```
        private int _selectedTopTabsIndex = 1;
        public int SelectedTopTabsIndex
        {
            get => _selectedTopTabsIndex;
            set
            {
                _selectedTopTabsIndex = value;
                RaisePropertyChanged(() => SelectedTopTabsIndex);
            }
        }
```

I guess it could be something related to events ordering, BindingContext initializations...
I'll comment more in case I get it.

[...]

PR #1748, opened 2021-11-30
Fix issue with the SelectedIndex on TabView thats appears when you try to set the bindable property SelectedItem from the view itself.

OlivCl:
### Description of Bug ###

We can't specified a selected page without entering in an infinite loop and many glitches of the screen.

`<xct:TabView x:Name="tbView"  Style="{StaticResource CustomTabStyle}" SelectedIndex="1" >
          <xct:TabViewItem /> 
           <xct:TabViewItem /> 
           <xct:TabViewItem />         
</xct:TabView>`

### Issues Fixed ###

https://github.com/xamarin/XamarinCommunityToolkit/issues/1231
https://github.com/xamarin/XamarinCommunityToolkit/issues/979

- Fixes #

After some investigations it appears that is due to the update on BindableProperty SelectedIndex inside Device.BeginInvokeOnMainThread

If we set the selected index inside the contructor of the page this exceptions is throws :

System.ArgumentOutOfRangeException: 'Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index'

If we set the selected index inside the OnAppearing event, no exception is throws but some glitches of the bottom tab and the selected page is not shown.


pictos (changes requested):
@OlivCl thanks for the PR, but this didn't fix the related issues, can you take a look on it?

PR #1760, opened 2021-12-07
Allow user interaction during animation to enable faster tap repetiti…

### LINKED ISSUE #1681: [Bug] On iOS - Animation stops double tap from being registered.
### Description
On iOS the animation will block other touch effects from occuring, despite UserInteractionEnabled being true.
In our case this occured in combination with a Grid, but it happens with other elements as well.
On Android both taps are registered like one would expect. 

### Steps to Reproduce
1. Add a command to the main label
`<Label
Margin="20,32,20,0"
Style="{StaticResource label_welcome_header}"
Text="Introducing the Xamarin Community Toolkit" 
xct:TouchEffect.NativeAnimation="True"
xct:TouchEffect.Command="{Binding ItemTappedCommand, Source={x:Reference Page}}"
/>`

2. A possible code behind
`public WelcomePage() {
    ItemTappedCommand = CommandFactory.Create(() =>
    {
        var bDoubleTap = (DateTime.Now - TapTime).TotalMilliseconds < 600;
        Console.WriteLine($"Tapped twice? - {bDoubleTap}");
        TapTime = DateTime.Now;
    });
    InitializeComponent();
}`

3. Now if you tap at the speed one would for a double tap there will only be one execution of the command. 
    With enough time between the first and second tap (in this example my estimation would be around 300-400ms)
    the second execution of the command will occur.

### Expected Behavior
Two taps at the speed of a double tap should result in two execution of the command. Like it already does on Android.
### Actual Behavior
The second tap will be blocked by the animation.
### Basic Information
- Version with issue: the current version of the main branch
- only on iOS

### Workaround
Although I did not manage to track the exact root cause which results in this behaviour the following solution fixes the issue and is working for us.
In "_PlatformTouchEffect.ios.cs_" change the function `await UIView.AnimateAsync(() => {...})` to the newer one in the description of the function itself. For example:
`UIViewPropertyAnimator.CreateRunningPropertyAnimator(.2, 0, UIViewAnimationOptions.AllowUserInteraction, () =>{...}, null);`

Andreas0xffff:
### Description of Bug ###
Quick tapping not possible with TouchEffect an animations.

Before:
AnimationSlow

After:
AnimationFast

### Issues Fixed ###
- Fixes #1681

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/PlatformTouchEffect.ios.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/PlatformTouchEffect.ios.cs
@@ -247,21 +247,24 @@ namespace Xamarin.CommunityToolkit.iOS.Effects
 			defaultShadowRadius = (float?)(defaultShadowRadius ?? renderer.Layer.ShadowRadius);
 			defaultShadowOpacity ??= renderer.Layer.ShadowOpacity;
 
-			await UIView.AnimateAsync(.2, () =>
-			{
-				if (color == Color.Default)
-					renderer.Layer.Opacity = isStarted ? 0.5f : (float)control.Opacity;
-				else
-					renderer.Layer.BackgroundColor = (isStarted ? color : control.BackgroundColor).ToCGColor();
-
-				renderer.Layer.CornerRadius = isStarted ? radius : defaultRadius.GetValueOrDefault();
-
-				if (shadowRadius >= 0)
+			var tcs = new TaskCompletionSource<UIViewAnimatingPosition>();
+			UIViewPropertyAnimator.CreateRunningPropertyAnimator(.2, 0, UIViewAnimationOptions.AllowUserInteraction,
+				() =>
 				{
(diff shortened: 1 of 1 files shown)


VladislavAntonyuk (approved):

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/Touch/PlatformTouchEffect.ios.cs
```diff
@@ -247,21 +247,24 @@ public async Task HandleTouch(TouchStatus status, TouchInteractionStatus? intera
...
+						renderer.Layer.BackgroundColor = (isStarted ? color : control.BackgroundColor).ToCGColor();
+
+					renderer.Layer.CornerRadius = isStarted ? radius : defaultRadius.GetValueOrDefault();
+
+					if (shadowRadius >= 0)
+					{
+						renderer.Layer.ShadowRadius = isStarted ? shadowRadius : defaultShadowRadius.GetValueOrDefault();
+						renderer.Layer.ShadowOpacity = isStarted ? 0.7f : defaultShadowOpacity.GetValueOrDefault();
+					}
+				}, endPos => tcs.SetResult(endPos));
```
VladislavAntonyuk, line 266: we need to check if task has already completed to prevent
image

PR #1780, opened 2022-01-02
Fixed camera preview for Xiaomi devices

### LINKED ISSUE #1259: [Bug] Error: Failed to create captire sesstion
I get the error: "Failed to create capture session" when trying to use the front camera on the Xiaomi Redmi note 7 Android 10 device.

Axemasta:
### Description of Bug ###
`CameraView` preview does not work correctly for front facing camera on Xiaomi Note 7/8 devices.

This appears to be a niche issue on Xiaomi devices of a suitable age (the newer ones don't appear to show the issue) where the front camera cannot be used to preview / capture media. Instead of displaying a preview, the media capture failed event is raised & the camera cannot be used.

Below is the XCT sample running on a Xiaomi Redmi Note 7 & a Samsung A21S. The Samsung has no issues displaying the front camera and the Xiaomi cannot display it.

Xiaomi with issue

https://user-images.githubusercontent.com/33064621/147880984-c3600894-d4d6-4bd5-8d15-879f7f8f5ef7.mp4

Samsung without issue

https://user-images.githubusercontent.com/33064621/147881025-d2706bc1-edfa-4cef-9481-b747013192d1.mp4

Upon debugging the source code with the affected device, I found that the `ChooseOptimalSize` method was responsible for returning a resolution that could not be displayed.

The following [line](https://github.com/xamarin/XamarinCommunityToolkit/blob/ccf782c98d28bcdc61ec3ec7e5102a3c6508ac39/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs#L1061) on the affected device returns a size of `4224x3136` and when the camera is initialized the following errors are logged to the debug console:
```
[CameraDevice-JV-1] Stream configuration failed due to: endConfigure:520: Camera 1: Unsupported set of inputs/outputs provided
[CameraCaptureSession] Session 0: Failed to create capture session; configuration failed
```

After some reading up on the `Camera2` api it looks like this lib is following the example set by [Google](https://github.com/googlearchive/android-Camera2Basic/blob/4cc1c3e219d8d168b7893f7a6b2b348740679e5a/kotlinApp/Application/src/main/java/com/example/android/camera2basic/Camera2BasicFragment.kt#L777). Stepping through the issue comes with the aspect ratio check which always seems to fail on this particular device. Maybe it is an Android flavour conflict?

To sort of the issue I've added a check to see if the fallback value used is a valid size (smaller than the max preview), which will be used instead of the invalid value. This code change shouldn't affect working devices since it only alters the fallback logic and my Xiaomi is now happily giving me both cameras!

https://user-images.githubusercontent.com/33064621/147881330-65b81d77-4187-42bf-a137-4e7096b52026.mp4

### Issues Fixed ###

- Fixes #1259

### Behavioral Changes ###

`CameraView` should work on Xiaomi Redmi Note 7/8 devices 🥳

I removed the else clause on this [line](https://github.com/xamarin/XamarinCommunityToolkit/blob/ccf782c98d28bcdc61ec3ec7e5102a3c6508ac39/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs#L210) & made the failed to retrieve camera state return early, sorry guys I love to reduce nesting in methods 😉

### PR Checklist ###

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs
@@ -206,117 +206,118 @@ namespace Xamarin.CommunityToolkit.UI.Views
 
 				// _texture.ClearCanvas(Element.BackgroundColor.ToAndroid()); // HANG after select valid camera...
 				Element?.RaiseMediaCaptureFailed($"No {Element.CameraOptions} camera found");
+				return;
 			}
-			else
+
+			try
 			{
-				try
+				var characteristics = Manager.GetCameraCharacteristics(cameraId);
+				var map = (StreamConfigurationMap)(characteristics?.Get(CameraCharacteristics.ScalerStreamConfigurationMap) ?? throw new NullReferenceException());
+
+				flashSupported = characteristics.Get(CameraCharacteristics.FlashInfoAvailable) == Java.Lang.Boolean.True;
+				stabilizationSupported = false;
+				var stabilizationModes = characteristics.Get(CameraCharacteristics.ControlAvailableVideoStabilizationModes);
+
+				if (stabilizationModes is IEnumerable<int> modes)
 				{
-					var characteristics = Manager.GetCameraCharacteristics(cameraId);
-					var map = (StreamConfigurationMap)(characteristics?.Get(CameraCharacteristics.ScalerStreamConfigurationMap) ?? throw new NullReferenceException());
-
-					flashSupported = characteristics.Get(CameraCharacteristics.FlashInfoAvailable) == Java.Lang.Boolean.True;
-					stabilizationSupported = false;
-					var stabilizationModes = characteristics.Get(CameraCharacteristics.ControlAvailableVideoStabilizationModes);
-
-					if (stabilizationModes is IEnumerable<int> modes)
+					foreach (var mode in modes)
 					{
-						foreach (var mode in modes)
+						if (mode == (int)ControlVideoStabilizationMode.On)
+							stabilizationSupported = true;
+					}
+				}
+
+				if (Element != null)
+					Element.MaxZoom = maxDigitalZoom = (float)(characteristics.Get(CameraCharacteristics.ScalerAvailableMaxDigitalZoom) ?? throw new NullReferenceException());
+
+				activeRect = (Rect)(characteristics.Get(CameraCharacteristics.SensorInfoActiveArraySize) ?? throw new NullReferenceException());
+				sensorOrientation = (int)(characteristics.Get(CameraCharacteristics.SensorOrientation) ?? throw new NullReferenceException());
+
+				var displaySize = new APoint();
+				Activity.WindowManager?.DefaultDisplay?.GetSize(displaySize);
+
+				_ = texture ?? throw new NullReferenceException();
+				var rotatedViewWidth = texture.Width;
+				var rotatedViewHeight = texture.Height;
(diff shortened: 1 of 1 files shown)


bijington:
@Axemasta thank you for this submission! The changes look good to me especially given they don't effect the existing checks for a matching size (I do wonder if we could improve upon that when this shifts to the Maui Community Toolkit but that is outside of the scope of this change).

It will be nice to see if the 2 people that you tagged on the issue are able to confirm the fix, I will aim to test on some other devices I have here tomorrow to double check all is fine, I don't possess a Xiaomi device though.

Axemasta:
@bijington

To address your comments... in reverse order 😊:

> It will be nice to see if the 2 people that you tagged on the issue are able to confirm the fix, I will aim to test on some other devices I have here tomorrow to double check all is fine, I don't possess a Xiaomi device though.

I'm fortunate to have access to an affected device, mine is a Xiaomi Note 7 running Android 9. I potentially have access to more affected devices through a fork which I will add the change & observe if devices can access the camera correctly.

I've tested on 3 devices:

- Xiaomi Note 7
- Samsung A21
- Samsung A7 Tablet

Both the samsungs return valid surface mappings:
`map.GetOutputSizes(Class.FromType(typeof(SurfaceTexture)))`

The Xiaomi does return mappings but the aspect ratios are off, causing the checking code to completely fallthrough. I can't confirm whether its a Xiaomi or an Android flavour issue but it seems the fallback at least works for affected devices.

---

> @Axemasta thank you for this submission! The changes look good to me especially given they don't effect the existing checks for a matching size (I do wonder if we could improve upon that when this shifts to the Maui Community Toolkit but that is outside of the scope of this change).

When looking into this issue I did some research into how the Camera2 api should be implemented and it looks like the lastest google sample implements it in a different way. The current implementation matches the following:

- [Google: android-Camera2Basic](https://github.com/googlearchive/android-Camera2Basic)
- [Blog post](https://www.dynamsoft.com/codepool/android-camera-preview-app-camera2.html) | [Backing repo with references used](https://github.com/yushulx/android-camera-preview)

Both of these use the `ChooseOptimalSize` method which appears to be a method google cooked up to select the smallest preview size possible.

The [current google samples repo](https://github.com/android/camera-samples) does not contain this method and whilst I've not completely wrapped my head around the differences between old & new, it appears to be using a method called `getPreviewOutputSize` which you can find implemented [here](https://developer.android.com/training/camera2/multiple-camera-streams-simultaneously#output-size).

Since changing the current implementation would affect more devices, perhaps it would be something that could be revisited in another pr to update the android renderers to use the latest & greatest from the google sample?

bijington:
@Axemasta thanks for the guidance on the differences between this implementation and googles docs. I will add a note on the MAUI Community Toolkit proposal so we certainly don't miss this in the new implementation (https://github.com/CommunityToolkit/Maui/issues/106)

Given the rest appears to work and testing my devices here (Goole Pixel 3a, One Plus Nord N100 and some cheap Umidigi device) has been fine I suspect we might not worry about changing things in this repo.

bijington (approved):

bijington:
@Axemasta thanks again for this effort. It really is appreciated!

Axemasta:
Here is a Camera2 preview control that uses the preview builder to set the preview window. This is possibly the implementation we'd want to use for the MAUI version of the control [ref](https://github.com/JimmyPun610/BarcodeScanner.XF/blob/dd8c0d099b4c155a9aefd2accb92afe28affcf51/GoogleVisionBarCodeScanner/Android/Renderer/CameraViewRenderer.cs#L88).

PR #1817, opened 2022-02-21
Nullable fixes in DrawingView

### LINKED ISSUE #1747: [Bug] DrawingView In Expander (Android :☠️/ iOS : 🙈 )
### Description

Use a DrawingView in Expander is not stable.

Crash on Android 
NotVisible at first openning on iOS

### Link to Reproduction Sample

Works alone 👍
```xml
 <extensions:DrawingView x:Name="drawing"
  IsVisible="false"
  HeightRequest="200"
  BackgroundColor="LightGray"
  ClearOnFinish="False"
  LineColor="Red" />
```
Crash/NotShow on "openning" Expander 
```xml
   <extensions:Expander>
      <extensions:Expander.Header>
          <Label Text="test" />
      </extensions:Expander.Header>
      <extensions:Expander.Content>
          <extensions:DrawingView x:Name="drawing"
                                  IsVisible="false"
                                  HeightRequest="200"
                                  BackgroundColor="LightGray"
                                  ClearOnFinish="False"
                                  LineColor="Red" />
      </extensions:Expander.Content>
  </extensions:Expander>

```
### Steps to Reproduce

1. Tap on Expander
2. App Crash :) // DrawingView not showing

### Expected Behavior

Android : No Crash
iOS : Show on openning

### Android
image

### iOS
image

### Basic Information

- Version with issue: from 1.20 to 1.3.1
- Last known good version: 🤷
- IDE: VS2022
- Platform Target Frameworks: 
  - iOS:  iOS 15.0+
  - Android: 10+

### Workaround

Binging Expander State to IsVisible property of DrawingView

```xml
        <extensions:Expander State="{Binding State}">
            <extensions:Expander.Header>
                <Label Text="test" />
            </extensions:Expander.Header>
            <extensions:Expander.Content>
                <extensions:DrawingView
                    x:Name="drawing"
                      IsVisible="{Binding IsOpenned}"
                    BackgroundColor="LightGray"
                    ClearOnFinish="False"
                    DefaultLineColor="Red"
                    HeightRequest="200"
                    />
            </extensions:Expander.Content>
        </extensions:Expander>
```
 ```cs
  public bool IsOpenned { get; set; } = false;

        private ExpandState _state;

        public ExpandState State
        {
            get { return _state; }
            set
            {
                if (value == ExpandState.Expanded)
                    IsOpenned = true;
                else
                    IsOpenned = false;

                _state = value;
            }
        }
```
https://user-images.githubusercontent.com/39386863/144059846-b2ca1e4f-e6c4-47d7-932b-3d2892b723dc.mp4

### PR DESCRIPTION
**Pull request by VladislavAntonyuk**
### Description of Bug ###

Remove null-forgiving operator in drawing view.

I am not able to deploy the app to android device with VS 17.2, I was not able to test it on Android.
It works fine on UWP, WPF

### Issues Fixed ###

- Fixes #1747 

### PR Checklist ###

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Renderer/DrawingViewRenderer.gtk.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Renderer/DrawingViewRenderer.gtk.cs
@@ -50,17 +50,17 @@ namespace Xamarin.CommunityToolkit.UI.Views
 				SetNativeControl(vBox);
 			}
 
-			if (e.OldElement != null)
+			if (e.OldElement != null && area is not null)
 			{
-				area!.ExposeEvent -= OnDrawingAreaExposed;
+				area.ExposeEvent -= OnDrawingAreaExposed;
 				area.ButtonPressEvent -= OnMousePress;
 				area.ButtonReleaseEvent -= OnMouseRelease;
 				area.MotionNotifyEvent -= OnMouseMotion;
 			}
 
-			if (e.NewElement != null)
+			if (e.NewElement != null && area is not null)
 			{
-				area!.ExposeEvent += OnDrawingAreaExposed;
+				area.ExposeEvent += OnDrawingAreaExposed;
 				area.ButtonPressEvent += OnMousePress;
 				area.ButtonReleaseEvent += OnMouseRelease;
 				area.MotionNotifyEvent += OnMouseMotion;
@@ -82,8 +82,13 @@ namespace Xamarin.CommunityToolkit.UI.Views
 
 		void OnDrawingAreaExposed(object source, ExposeEventArgs args)
 		{
+			if (area is null)
+			{
+				return;
+			}
+
 			Context ctx;
-			using (ctx = CairoHelper.Create(area!.GdkWindow))
+			using (ctx = CairoHelper.Create(area.GdkWindow))
 			{
 				ctx.SetSource(new SurfacePattern(surface));
 				ctx.Paint();
@@ -111,7 +116,7 @@ namespace Xamarin.CommunityToolkit.UI.Views
 			};
 			previousPoint = point;
 			isDrawing = true;
-			area!.QueueDraw();
+			area?.QueueDraw();
 			if (!Element.MultiLineMode)
 				Element.Lines.Clear();
 		}
@@ -125,7 +130,7 @@ namespace Xamarin.CommunityToolkit.UI.Views
 			using var ctx = new Context(surface);
 			DrawPoint(ctx, point);
 
-			area!.QueueDraw();
+			area?.QueueDraw();
 			if (currentLine != null)
 			{
 				Element.Lines.Add(currentLine);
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Renderer/DrawingViewRenderer.gtk.cs [hunk 1]
```diff
@@ -146,19 +151,23 @@ void OnMouseMotion(object source, MotionNotifyEventArgs args)
...
+				ctx.SetSourceRGBA(currentLine.LineColor.R, currentLine.LineColor.G, currentLine.LineColor.B, currentLine.LineColor.A);
+				ctx.LineWidth = currentLine.LineWidth;
+			}
+
 			ctx.MoveTo(previousPoint);
 			previousPoint = pointD;
 			ctx.LineTo(pointD);
 			ctx.Stroke();
-			currentLine.Points.Add(new Point(pointD.X, pointD.Y));
+			currentLine?.Points.Add(new Point(pointD.X, pointD.Y));
```
bijington, line 170: Could this go inside the `if (currentLine is not null)` and avoid the need for the `?`

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Renderer/DrawingViewRenderer.tizen.cs
```diff
@@ -36,17 +36,17 @@ protected override void OnElementChanged(ElementChangedEventArgs<DrawingView> e)
 				SetNativeControl(canvasView);
 			}
 
-			if (e.OldElement != null)
+			if (e.OldElement != null && canvasView is not null)
```
bijington, line 39: I'm definitely a fan of removing the `!` operator! One question though, is `canvasView` being null a valid scenario? If not should we still throw an exception so the consumer can detect something is wrong?

This potentially ripples through most of the changes but I didn't want to add loads of points if it is valid.

bijington:
Thanks @VladislavAntonyuk my comments may well be my not understanding how the `DrawingView` works but they stood out to me

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Renderer/DrawingViewRenderer.tizen.cs
VladislavAntonyuk: In this case I don’t think it’s a good idea to throw the exception. Here we are cleaning all resources and unsubscribe from events (most likely user closed the page with control). Throwing exception here means user cannot navigate to another page.
Furthermore it is unlikely that canvasView will be null at this moment

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Renderer/DrawingViewRenderer.gtk.cs
VladislavAntonyuk: Fix me if I am wrong but it’s a short form of writing it. 
`
instance?.Method()
`

is the same as 
`
if (instance is not null)
    instance.Method()
`

bijington: Yes it is the same, I actually meant inside the if higher up on line 160. Only a very minor point

VladislavAntonyuk: ahh got you, just move it a few lines up. yes. Let's do it!

src/CommunityToolkit/Xamarin.CommunityToolkit/Views/DrawingView/Renderer/DrawingViewRenderer.tizen.cs
bijington: Ah yes that makes sense. I picked the worst example but just wanted to check we accept a `null` `canvasView` is a valid scenario in our renderer which it sounds like it is.

VladislavAntonyuk: actually no. the canvasView should always be initialized because it is actually the place where you draw. But because we initialize it here (not in ctor), we mark it as nullable and need to check that it is not null in a whole class.

bijington (approved):
Great work thanks @VladislavAntonyuk

Issue #934, opened 2021-02-18
[Bug] IconTintColorEffect does not work when the image is in Shared project as EmbeddeResource


azapatac:
### Description
An added image in the shared project as EmbeddeResource does not change TintColor

### Steps to Reproduce

1. Create Xamarin project
2. Create ImageResourceExtension : IMarkupExtension
3. Add image the shared project and set up Build Action = EmbeddeResource
4. Add image in content page and apply IconTintColorEffect

### Expected Behavior
The image color must change the color

### Actual Behavior
The image has the original color

### Basic Information

Visual Studio Community 2019 for Mac
Version 8.8.8 (build 12)
Installation UUID: 4fcbc297-9169-4ef5-b390-ed7872967085
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 6.18.0.23 (d16-6 / 088c73638)

	Package version: 612000113

Mono Framework MDK
Runtime:
	Mono 6.12.0.113 (2020-02/4fdfb5b1fd5) (64-bit)
	Package version: 612000113

Roslyn (Language Service)
3.8.0-5.20519.18+4c195c3ac1974edcefa76774d7a59a2350ec55fa

NuGet
Version: 5.8.0.6860

.NET Core SDK
SDK: /usr/local/share/dotnet/sdk/5.0.103/Sdks
SDK Versions:
	5.0.103
	5.0.101
	3.1.406
	3.1.405
	3.1.404
MSBuild SDKs: /Applications/Visual Studio.app/Contents/Resources/lib/monodevelop/bin/MSBuild/Current/bin/Sdks

.NET Core Runtime
Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	5.0.3
	5.0.1
	3.1.12
	3.1.11
	3.1.10

.NET Core 3.1 SDK
SDK: 3.1.406

Xamarin.Profiler
Version: 1.6.15.68
Location: /Applications/Xamarin Profiler.app/Contents/MacOS/Xamarin Profiler

Updater
Version: 11

Apple Developer Tools
Xcode 12.4 (17801)
Build 12D4e

Xamarin.Mac
Xamarin.Mac not installed. Can't find /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/Version.

Xamarin.iOS
Version: 14.10.0.4 (Visual Studio Community)
Hash: 5a05865f6
Branch: xcode12.4
[...]

PR #1836, opened 2022-03-25
Fix for Bug #934 (IconTintColorEffect does not work when the image is…

mike-rowley:
This is the PR for Bug #934 IconTintColorEffect does not work when the image is in Shared project as EmbeddeResource
https://github.com/xamarin/XamarinCommunityToolkit/issues/934

I have implemented the fix @TheDude1604 suggested in the bug report and it resolves the issue on IOS.  I have tested it on a physical device as well as a simulator.

If there are any issues please let me know and I will try to resolve them.  Thanks for considering this PR.

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/IconTintColor/IconTintColorEffectRouter.ios.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/IconTintColor/IconTintColorEffectRouter.ios.cs
@@ -99,8 +99,8 @@ namespace Xamarin.CommunityToolkit.iOS.Effects
 				{
 					if (e.PropertyName == ImageButton.IsLoadingProperty.PropertyName)
 					{
-						var b = Element as Xamarin.Forms.ImageButton ?? throw new NullReferenceException();
-						if (!b.IsLoading)
+						var element = Element as Xamarin.Forms.ImageButton ?? throw new NullReferenceException();
+						if (!element.IsLoading)
 						{
 							SetUIButtonTintColor(button, color);
 						}
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/IconTintColor/IconTintColorEffectRouter.ios.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/IconTintColor/IconTintColorEffectRouter.ios.cs
@@ -54,11 +54,16 @@ namespace Xamarin.CommunityToolkit.iOS.Effects
 			{
 				case UIImageView imageView:
 					if (imageView.Image != null)
+					{
+						Element.PropertyChanged -= ImageViewTintColorPropertyChanged;
 						imageView.Image = imageView.Image.ImageWithRenderingMode(UIImageRenderingMode.AlwaysOriginal);
+					}
+
 					break;
 				case UIButton button:
 					if (button.CurrentImage != null)
 					{
+						Element.PropertyChanged -= ButtonTintColorPropertyChanged;
 						var originalImage = button.CurrentImage.ImageWithRenderingMode(UIImageRenderingMode.AlwaysOriginal);
 						button.SetImage(originalImage, UIControlState.Normal);
 					}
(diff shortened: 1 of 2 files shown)


pictos:
@mike-rowley can you assign the CLA?

mike-rowley:
I did both CLAs, now when I click the "sign now" link it says "You have signed the CLA for multiple repositories or organizations" and all options are disabled with my name etc in the fields.  Seems like some sort of glitch in the CLA process unless there is something I am missing.

brminnick:
Done ✅

pictos (changes requested):
@mike-rowley thanks for sharing the solution with us❣️
So, this retry logic doesn't look very solid to me, because there's a chance to the image loads after all retries attempts, would be good if we can use something more solid. So my suggestion is to use the `IsLoading` property inside the `XF.Image` control, here's a snippet of code that I tried on my end and worked. I put this code inside the `void SetUIImageViewTintColor(UIImageView imageView, Color color)` method.

```csharp
if (imageView.Image == null)
{
    Element.PropertyChanged += (_, e) =>
    {
        if (e.PropertyName == Image.IsLoadingProperty.PropertyName)
        {
            var b = Element as Xamarin.Forms.Image ?? throw new NullReferenceException();
            if (!b.IsLoading)
            {
                SetUIImageViewTintColor(imageView, color);
            }
        }
    };
}
```

I'm not sure if we have something like that for the `ImageButton` control, can you take a look on that? If you need any help, just let me know (:

mike-rowley:
Thanks @pictos I took your code and tested it on my side and it works great, a much more elegant solution!  I also converted that code to the Button and it works there as well so I pushed another PR with the updated code.

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/IconTintColor/IconTintColorEffectRouter.ios.cs
```diff
@@ -69,43 +67,45 @@ void ClearTintColor()
...
 			if (imageView.Image == null)
 			{
-				if (_retryCount < 4)
-				{
-					_retryCount++;
-					await Task.Delay(100);
-					SetUIImageViewTintColor(imageView, color);
-				}
-				else
+				Element.PropertyChanged += (_, e) =>
```
pictos, line 74: We are almost there; I suggest creating a method for this because we need to unsubscribe this event when the effect is Detached (same for ImageButton)

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/IconTintColor/IconTintColorEffectRouter.ios.cs
```diff
@@ -69,43 +67,45 @@ void ClearTintColor()
...
 				{
-					return;
-				}
+					if (e.PropertyName == Image.IsLoadingProperty.PropertyName)
+					{
+						var b = Element as Xamarin.Forms.Image ?? throw new NullReferenceException();
+
+						if (!b.IsLoading)
```
pictos, line 80:
```suggestion
						var element = Element as Xamarin.Forms.Image ?? throw new NullReferenceException();

						if (!element.IsLoading)
```

src/CommunityToolkit/Xamarin.CommunityToolkit/Effects/IconTintColor/IconTintColorEffectRouter.ios.cs
```diff
@@ -69,43 +67,45 @@ void ClearTintColor()
...
-				}
-				else
+				Element.PropertyChanged += (_, e) =>
 				{
-					return;
-				}
+					if (e.PropertyName == ImageButton.IsLoadingProperty.PropertyName)
+					{
+						var b = Element as Xamarin.Forms.ImageButton ?? throw new NullReferenceException();
+						if (!b.IsLoading)
```
pictos, line 103:
```suggestion
						var element = Element as Xamarin.Forms.ImageButton ?? throw new NullReferenceException();
						if (!element.IsLoading)
```

pictos:
@mike-rowley glad that you liked the solution! I have just a feel comments and after that we should be good to merge it

mike-rowley:
@pictos Good call on unsubscribing, I think this version should do it.  

Somehow I accidentally referenced a bug from 2017 in the summary when I pushed it and I am not sure how to change that "xamarin#2" in the summary.

pictos (approved):
@mike-rowley thanks for this!

PR #1857, opened 2022-05-15
Fix for ClearTintColor NullReferenceException on ImageButton related to PR #1836

mike-rowley:
I originally sent over a PR for Issue #934 and have found a small issue not addressed in PR #1836 which I also have this PR to resolve.

This PR is to fix a NullReferenceException in ClearTintColor when setting Application.MainPage to a different page when the current page uses the Tint Effect.  In ClearTintColor the button.ImageView is null and not checked.  I have added the conditional access qualifier and it resolves this issue.
I have also moved the detach PropertyChanged outside of the Image Null Checks in ClearTintColor as I noticed that when the ImageButton is being detached the ImageView of the button is null (I am unsure why) however I believe the PropertyChanged Event added in SetUITintColor functions should be removed regardless of the null state of the Image.


bijington (approved):
@mike-rowley thanks for this submission

PR #1864, opened 2022-05-19
CameraView Flash fix for Android

### LINKED ISSUE #1863: [Bug] [CameraView] [Android] Flash works just for the first capture
### Description

The CameraView does not trigger the flashlight after the first capture even keeping the FlashMode as On.

### Link to Reproduction Sample

This issue can be reproduced on XCT CameraView Sample.
### Steps to Reproduce

1. Open CameraView
2. Turn on Flash
3. Tap SNAP PICTURE, the flash works
4. Tap SNAP PICTURE again, the flash doesn't work

### Expected Behavior

The flash works for all creatures if on.

### Actual Behavior

The flash works just for the first capture

### Basic Information

- Version with issue: main branch
- Last known good version:
- IDE:
- Platform Target Frameworks: 
  - iOS:  
  - Android: 11  
  - UWP:   
- Nuget Packages:
- Affected Devices:

### Workaround

I couldn't find a good workaround, I have even tried forcing Turn Flash Off and then Turn Flash On again.

GUOLDEV:
### Description of Bug ###

The CameraView does not trigger the flashlight after the first capture even keeping the FlashMode as On.

### Issues Fixed ###

- Fixes #1863 

### Behavioral Changes ###

```
if (flashMode == FlashMode.Single)
{
    sessionBuilder.Set(CaptureRequest.FlashMode ?? throw new NullReferenceException(), (int)FlashMode.Off);
    session.Capture(sessionBuilder.Build(), null, null);
}
```

### PR Checklist ###

### DIFF
--- a/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs
+++ b/src/CommunityToolkit/Xamarin.CommunityToolkit/Views/CameraView/Android/CameraFragment.android.cs
@@ -341,6 +341,14 @@ namespace Xamarin.CommunityToolkit.UI.Views
 				{
 					session.StopRepeating();
 					repeatingIsRunning = false;
+
+					// Reset FlashMode if Single
+					if (flashMode == FlashMode.Single)
+					{
+						sessionBuilder.Set(CaptureRequest.FlashMode ?? throw new NullReferenceException(), (int)FlashMode.Off);
+						session.Capture(sessionBuilder.Build(), null, null);
+					}
+
 					sessionBuilder.AddTarget(photoReader.Surface);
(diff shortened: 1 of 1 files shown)


GUOLDEV:
Thanks @pictos 👍  Just to give you some context about the fix, the reason that I need to reset the session is that with the current structure we keep using the same session during CameraView's life cycle, and having the FlashMode as Single, it is not going to trigger the Flash until you turn off and turn on again. To do a proper fix, it will require a large refactoring and I believe it is not what you want as it would require more testing, etc... Please let me know what you think and I am happy to follow your guidance. Thanks in advance 🙏

pictos (approved):
@GUOLDEV is really to see you here after a while. I tried it on my end works fine. Thanks for the fix

